Issue #6, opened 2020-02-20
epic: `shallow: true` mounting option (previously `shallowMount`)


lmiller1990:
Will probably rely on #3. I have some ideas on how to handle this. 

Instead of `shallowMount`, I think we should have a `shallow` mounting option (discussed internally).

```js
const Comp = mount({
  shallow: true
})
```

### How this should work

`shallowMount` was pretty buggy in VTU beta. It should completely stub out and child components. So:

```js
<template>
  <div>Content</div>
  <Foo />
  <Bar>
    <div>BAR</div>
  </Bar>
</template>
```

Should render as follows when using `shallow`:

```js
<div>Content</div>
<foo-stub />
<bar-stub />
```

Whether it's `<FooStub />` or `<foo-stub> `/> doesn't matter, the point is that the components are no longer rendered, nor is any of their code executed.

Things to consider:

- should a user be able to `find` a stubbed out component? Feels a bit weird to try and find a component that is purposefully NOT rendered. But it might be a common use case
- current `find` does not support `find(Component)` syntax. I'm not sure it should - it was impossible `find`. One such issue: https://github.com/vuejs/vue-test-utils/issues/1272 and 
- consider this use case https://github.com/vuejs/vue-test-utils/issues/1410 people often stub out components then want to emit an event from them. Can we handle this better?

PR #21, opened 2020-03-26
wip: shallow

lmiller1990:
It works! Tracking in #6.

Evan made a feature to let us have shallowMount without hacking Vue to pieces like we do in beta:

[Source](https://github.com/vuejs/vue-next/blob/925ba6bb7bb21c9bc7d3bf2b246026c7f3b41caa/packages/runtime-core/src/vnode.ts#L227)
[Tests](https://github.com/vuejs/vue-next/blob/master/packages/runtime-core/__tests__/vnode.spec.ts#L342)

Basically, you call `transformVNodeArgs` and you can override the default behavior - this is so clean.

TODO:

1. no way to access the instance of stubbed component (maybe we don't need it anymore? only use case I can think of is making a stubbed child emit and event, we will need to figure this out)
2. this stubs out EVERYTHING; implementing `stubs` should be easy, though (I prefer people use a test runner for this, eg `jest.mock`? Although this is a popular feature in beta)
3. I really like the idea of having a `shallow` mounting option instead of a `shallowMount` method. I see `shallow` as just one feature of VTU, not a main API. Instead of "you can mount or shallowMount" I want people to say "you mount your component. There is a `shallow` mounting option, among other mounting options, you can use to test your components).

I wonder instead of using a basic stub we can use a more specialized on that can emit events or something 🤔


lmiller1990:
closed in favor of #56, I will use the work done there for shallow mount

PR #52, opened 2020-04-11
feat: setData

lmiller1990:
Still not a big fan of this feature, but at least it was easy to implement 👍

### DIFF
--- a/src/vue-wrapper.ts
+++ b/src/vue-wrapper.ts
@@ -1,25 +1,18 @@
 import { ComponentPublicInstance, nextTick, App } from 'vue'
-import { ShapeFlags } from '@vue/shared'
-import merge from 'lodash/merge'
 
+import { ShapeFlags } from './utils/vueShared'
 import { config } from './config'
 import { DOMWrapper } from './dom-wrapper'
-import {
-  FindAllComponentsSelector,
-  FindComponentSelector,
-  WrapperAPI
-} from './types'
-import { ErrorWrapper } from './error-wrapper'
+import { FindAllComponentsSelector, FindComponentSelector } from './types'
+import { createWrapperError } from './error-wrapper'
 import { TriggerOptions } from './create-dom-event'
 import { find } from './utils/find'
 
-// @ts-ignore
-export class VueWrapper<T extends ComponentPublicInstance>
-  implements WrapperAPI {
+export class VueWrapper<T extends ComponentPublicInstance> {
   private componentVM: T
   private rootVM: ComponentPublicInstance
   private __app: App | null
-  private __setProps: (props: Record<string, any>) => void
+  private __setProps: ((props: Record<string, any>) => void) | undefined
 
   constructor(
     app: App | null,
@@ -27,7 +20,7 @@ export class VueWrapper<T extends ComponentPublicInstance>
     setProps?: (props: Record<string, any>) => void
   ) {
     this.__app = app
-    this.rootVM = vm.$root
+    this.rootVM = vm.$root!
     this.componentVM = vm as T
     this.__setProps = setProps
     // plugins hook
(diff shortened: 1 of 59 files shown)


src/vue-wrapper.ts
```diff
@@ -78,6 +79,13 @@ export class VueWrapper implements WrapperAPI {
     return Array.from(results).map((x) => new DOMWrapper(x))
   }
 
+  setData(data: Record<string, any>) {
+    // lodash.merge merges by *reference* so this will update
+    // any existing data with the newly passed data.
+    merge(this.componentVM.$data, data)
```
dobromir-hristov, line 85: Should we guard against setting data that is not in `$data` already? Not sure

lmiller1990: Let's check how Vue 3 behaves. I wonder if the new reactivity API allows you to update with new properties? I think we should match whatever Vue 3 does.

lmiller1990:
Should I re-base and merge this? I forgot why we left this open.

afontcu:
> Should I re-base and merge this? I forgot why we left this open.

I think we said to wait until later alphas, and see if people asked about this particular feature. I won't be using it personally, but some concerns were raised in the RFC so we might just add it and keep it as a low priority feature.

lmiller1990:
Dropped lodash dep, will need to write our own `merge` function before merging this.

lmiller1990:
I will close this for now, we can implement if/when we need it.

jackmu95:
We at [BootstrapVue](https://github.com/bootstrap-vue/bootstrap-vue) are using `wrapper.setData()` quite heavily.
We are currently in the process of creating a Vue 3 compatible version and it would be good to have that feature in VTU v2.

I'am also willing to test this with our current tests and give feedback.

JessicaSachs:
Would you like help in refactoring to move away from setData? I see 11 code
usages in bootstrap Vue (
https://github.com/bootstrap-vue/bootstrap-vue/search?q=setData&type=
) and I feel comfortable submitting a PR to migrate off of it.

Implementing setData allows users to test the implementation of the
component instead of the contract and is generally a bad practice (which is
why Adria, Lachlan, and I don’t use it or promote it ever)

jackmu95:
You are right, we don't use it that much.
A PR would be really appreciated :)

lmiller1990:
I think we *do* need to implement this - backwards compat is pretty important, and I think this feature is very easy to implement. I also expect a lot of push-back if we were to deprecate this. I would recommend implementing it and referencing it in the API reference part of the docs, but not including it in any of our guides (I think we settled on this been unideal in most cases - just use the `data` mounting option).

What does everyone think?

jackmu95:
@lmiller1990 I would prefer to have this in v2 too.

A guide in the docs on why you should avoid using it and what better alternatives are seems to be a good idea.

lmiller1990:
The reason it's not ideal is because arbitrarily setting data on a component is not something a user can do - ideally, test like a user does. Eg, if the data is set via a button, click the button with `trigger` etc. 

Either way, we should support this. I will make a PR in the near future unless someone else does first - you can reference this PR to see how it can be implemented.

JessicaSachs:
Surprised to hear your stance on this -- I remember this being an unknown
amount of complexity. I'm happy to give users a non-breaking path. +1

lmiller1990:
In terms of implementation, this was working https://github.com/vuejs/vue-test-utils-next/pull/52/files (this PR). Unless I am missing some specific use case?

Don't get me wrong, I still think `setData` sucks an is an anti-pattern... but this conflicts with "unopinionated utilities for testing". We can be opinionated in the docs 👍

afontcu:
I'm also OK with adding `setData` for the sake of avoiding breaking changes, and document it as one of the "Advanced topics" (re https://github.com/vuejs/vue-test-utils-next-docs/issues/17#issue-611372985) VTU-next offers 👍

PR #53, opened 2020-04-11
Add global configuration settings

### LINKED ISSUE #35: Add global configuration settings
## Description

We should allow users to register default global components, directives and plugins, instead of passing to each test all the time.

## Detailed 

In VTU beta we could add a few global configs. Stubs was the main one with a few more later added.

Components, filters and directives were easy to add globally to all components, you just had to do this in your setup file:

```js
Vue.use(Plugin)
Vue.component(GlobalComponent)
// etc
```

With Vue 3, you now have a Vue App instance, which means you no longer can add global configuration like that. This would lead to allot of repetition in tests, and large codebases could suffer greatly. 

We could provide the same api as we have now in `MountOptions.global`, just at a global level.

dobromir-hristov:
This PR adds a global config object, to allow users to add global Mixins, mocks or others for all Test files.

Closes #35


tests/mountingOptions/globalProperties.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+describe('globalProperties', () => {
+  it('adds globally available properties to the component', () => {
+    const Component = {
+      render() {
+        return h('div', this.foo)
+      }
+    }
+    const wrapper = mount(Component, {
+      global: {
+        globalProperties: {
```
lmiller1990, line 13: Is there any use case for to override a global property one just one component?

dobromir-hristov: Idk... Data there would most probably be static, or just functions.

lmiller1990: Sorry I forgot about this. I still don't see the use case for `globalProperties` inside of `global`. Can we remove it (at least until someone has a real use case).

Everything else looks good.

dobromir-hristov: You want to remove it entirely, or remove it from `global` but keep it in the global `config` file?
I think it makes sense to have this somewhere, as it is an API, that is available to the App, and people would use it to provide fallbacks for old projects, like filters, constants or attached properties to the `prototype`.
We should be able to use the `mocks` for that, I am just not sure if they function the same, reactivity and all.

afontcu: could we put the feature on hold until someone asks for a legitimate use case for it? I understand what Dobromir suggests, and I'd like people to ask for it – providing fallbacks for old projects might be a slippery slope. Maybe `mocks`  is enough after all.

lmiller1990: The first one @dobromir-hristov  - I think we should keep the global config - that's something beta supported. I was thinking we remove the `globalProperties` mounting option, since `mocks` can already accomplish this (and it's a standard API from beta).

For the global config, I am happy to wait for someone to ask for it.  I can think of at least two common use cases for global config:

- transition stubs
- `$t` for apps using i18n

src/mount.ts
```diff
@@ -28,16 +27,7 @@ interface MountingOptions {
...
-    mixins?: ComponentOptions[]
-    mocks?: Record<string, any>
-    stubs?: Record<any, any>
-    provide?: Record<any, any>
-    // TODO how to type `defineComponent`? Using `any` for now.
-    components?: Record<string, Component | object>
-    directives?: Record<string, Directive>
-  }
+  global?: GlobalMountOptions
   stubs?: Record<string, any>
```
dobromir-hristov, line 31: @lmiller1990 should `stubs` stay here or be moved in `global` ?

lmiller1990: hmmm good point, probably global yeah? since they are "globally" stubbed.

dobromir-hristov: Fixed

src/utils.ts
```diff
@@ -2,7 +2,28 @@ import camelCase from 'lodash/camelCase'
...
 
 const pascalCase = flow(camelCase, upperFirst)
 
 export { kebabCase, pascalCase }
+
+export function mergeGlobalProperties(
+  configGlobal = {},
+  mountGlobal = {}
+): GlobalMountOptions {
+  return mergeWith({}, configGlobal, mountGlobal, (objValue, srcValue, key) => {
```
afontcu, line 16: Nice 🙌

PR #55, opened 2020-04-12
POC extending without plugin

dobromir-hristov:
This is just an example how one can extend VTU with either normal prototype attachment, or via a dedicated API.

Please give me your opinions on which one you think is better?

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,0 +1,6 @@
+export const config = {
+  plugins: {
+    VueWrapper: {} as any,
+    DOMWrapper: {}
+  }
+}
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,4 +2,6 @@ import { mount } from './mount'
 import { RouterLinkStub } from './components/RouterLinkStub'
 import { VueWrapper } from './vue-wrapper'
 
-export { mount, RouterLinkStub, VueWrapper }
+import { config } from './config'
+
+export { mount, RouterLinkStub, VueWrapper, config }
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,5 @@
 import { mount } from './mount'
 import { RouterLinkStub } from './components/RouterLinkStub'
+import { VueWrapper } from './vue-wrapper'
 
-export { mount, RouterLinkStub }
+export { mount, RouterLinkStub, VueWrapper }
--- a/src/vue-wrapper.ts
+++ b/src/vue-wrapper.ts
@@ -1,5 +1,6 @@
 import { ComponentPublicInstance, nextTick } from 'vue'
 import { ShapeFlags } from '@vue/shared'
+import { config } from './config'
 
 import { DOMWrapper } from './dom-wrapper'
 import { WrapperAPI } from './types'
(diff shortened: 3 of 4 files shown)


dobromir-hristov:
@afontcu @JessicaSachs  What is your opinion? 

Should we promote `prototype` extension or create a dedicated api, like the one I have proposed here? I think i am aiming at the prototype one.

afontcu:
> Should we promote `prototype` extension or create a dedicated api, like the one I have proposed here? I think i am aiming at the prototype one.

hey! 👋 Not 100% sure, actually 😇  I'm a bit concerned with the actual implementation of both examples:

`return this.componentVM.$.type.name`

vs

`return this.name().toLocaleLowerCase()`

the former looks way more complicated to grasp than the latter. Are those two equivalent? Is there a way to simplify the first one?

Thanks for this!

dobromir-hristov:
Haha no no, that was just a way to test whether they work. One uses the method added prior.

 How you access the internals is the same in both cases.

afontcu:
> Haha no no, that was just a way to test whether they work. One uses the method added prior.
> How you access the internals is the same in both cases.

Yeah okay makes sense hahaha. 

I do think that relying on prototypes "feels" hacky, even though I'm OK with this given that it's an advanced feature and that it should be documented 👍

lmiller1990:
So if we are going with the `VueWrapper.prototype` for now, do we just need to  export VueWrapper and that's it?

What is the status of this PR @dobromir-hristov

dobromir-hristov:
We just need to export DOMWrapper and VueWrapper and that should be it. We should also test this after it is transpiled, so that means when it is published as a package.

lmiller1990:
Ok, want to update this to just export those two? 

I made a ticket for some e2e tests: https://github.com/vuejs/vue-test-utils-next/issues/68

lmiller1990:
@dobromir-hristov should we export DOMWrapper too?

afontcu:
I guess this would become the "right" approach to extend VTU as Vue Testing Library does? I mean, the way to add "custom" matchers and other stuff. I'll try to give this a go during the weekend and come back to you folks with the result 👍

JessicaSachs:
So I just finished toying around. I found the API a bit awkward at first. I'll see if there's a pattern that feels more natural. I mostly found that I just wanted the `this` context in a closure while I was extending VueWrapper.

Before:
```javascript
config.plugins.VueWrapper.name = function name() {
  return this.componentVM.$.type.name
}

config.plugins.VueWrapper.lowerCaseName = function lowerCaseName() {  
  return this.name().toLocaleLowerCase()
}
```

Possible solution to enclose the wrapper instance?
```javascript
config.plugins.VueWrapper.extend((wrapper) => ({
  lowerCaseName: () => wrapper.name,
  name: () => wrapper.componentVM.$.type.name
})
```

The other thing I'd like to solve is the ability to use reactive properties so that not everything has to be a method. I'll continue toying with that.

dobromir-hristov:
So no prototype extension?

Private field would not be accessible if passed as a param.

JessicaSachs:
You would still apply it to the prototype when you install the plugin in the constructor. I'm mostly just concerned with getting a reference to `wrapper`.

lmiller1990:
Are we leaning towards @JessicaSachs implementation for a plug-in or this one? Let's pick one and focus on that. I am excited to write some plug-ins (mainly `findByTestId`).

PR #56, opened 2020-04-13
feat: stubs mounting option

### LINKED ISSUE #3: feat: stubs mounting option
We should support 

```js
mount: {
  stubs: {
    ComponentToStub: true // default stub
    Another: h => h('div') // custom stub
  }
}
```

lmiller1990:
resolves #3!!!! :partyparrot: I'm really happy with this implementation; it feels super clean, main because of the transforVNodeArgs function @JessicaSachs  and Evan made in core.

## Points of interest:

- `stubs` is now in the `global` mounting option, since when you stub a component, it stubs them all the way down the tree (same as beta)
- the stub is not a DOM node as discussed, but a bare-bones Vue component. Gives us some flexibility for implementing shallowMount.

## Breaking change:

We do not support an array of stubs (at least right now). So instead of 

```
stubs: [Foo, Bar]
```

You need to do:

```
stubs: {
  Foo: true, Bar: True
}
```

If this is a bit problem for people we can implement array notation for beta. For alpha, I think this is a good place to leave this and move on to other features.


src/mount.ts
```diff
@@ -29,6 +31,7 @@ interface MountingOptions {
     plugins?: Plugin[]
     mixins?: ComponentOptions[]
     mocks?: Record<string, any>
```
lmiller1990, line 33: Typing Vue components is very difficult, will need to investigate the Vue codebase to figure out how best to do this

src/mount.ts
```diff
@@ -133,6 +137,46 @@ export function mount(originalComponent: any, options?: MountingOptions) {
...
+
+      if (
+        typeof args[0] === 'object' &&
+        args[0]['name'] in options?.global?.stubs
+      ) {
+        const name = args[0]['name']
+        // default stub
+        if (options?.global?.stubs[name] === true) {
+          return [createStub({ name: args[0]['name'] })]
+        }
```
lmiller1990, line 166: @dobromir-hristov we one of us gets to working on `findComponent` we will probably need to pass `props` as the second arg here, so we can do `findComponent(Foo).props)`. So it will be like 

`return [createStub({ name: args[0]['name'] }), args[1]]` or something. I didn't do it here because I am not exactly sure how this will work yet, but I don't expect it to be a blocker.

src/mount.ts
```diff
@@ -133,6 +137,46 @@ export function mount(originalComponent: any, options?: MountingOptions) {
   const { emitMixin, events } = createEmitMixin()
   vm.mixin(emitMixin)
 
+  // stubs
+  if (options?.global?.stubs) {
+    transformVNodeArgs((args) => {
+      // regular HTML Element. Do not stubs these
+      if (Array.isArray(args) && typeof args[0] === 'string') {
```
afontcu, line 144: Do you think we could extract all this assertions+comments into inline functions? I think it would make each early return cleaner:

```js
      if (isHtmlElement(args) || isComment(args) || isParentComponent(args)) {
````

or similar ^^

same thing with the if statement in lines 164 and 168, too.


Just my two cents 😇 feel free to resolve this if you think this will either be simple enough to understand, or change significantly so this refactor does not make sense.

Comment (lmiller1990) [reply src/mount.ts:144]: yep. sounds good. I was so focused on getting it to work I lost track of how unreadable it was getting. let's keep this codebase clean and understandable!

Comment (lmiller1990) [reply src/mount.ts:144]: (if anyone is reading this who would like to make this refactor before I do feel free to make a PR targeting this one)

File: src/mount.ts (hunk 2 above)
Comment (dobromir-hristov) [reply src/mount.ts:166]: Yep, we will check it out.

File: src/mount.ts [hunk 4]
```diff
@@ -133,6 +137,46 @@ export function mount(originalComponent: any, options?: MountingOptions) {
   const { emitMixin, events } = createEmitMixin()
   vm.mixin(emitMixin)
 
+  // stubs
+  if (options?.global?.stubs) {
+    transformVNodeArgs((args) => {
```
dobromir-hristov, line 142: Maybe we can destructure like this, so its easier to understand context later down?
```suggestion
    transformVNodeArgs(([componentOptions]) => {
```

src/mount.ts
```diff
@@ -133,6 +137,46 @@ export function mount(originalComponent: any, options?: MountingOptions) {
...
+      ) {
+        const name = args[0]['name']
+        // default stub
+        if (options?.global?.stubs[name] === true) {
+          return [createStub({ name: args[0]['name'] })]
+        }
+
+        // custom stub implementation
+        if (typeof options?.global?.stubs[name] === 'object') {
+          return [options?.global?.stubs[name]]
```
dobromir-hristov, line 170:
```suggestion
          return [options.global.stubs[name]]
```

src/mount.ts
```diff
@@ -133,6 +137,46 @@ export function mount(originalComponent: any, options?: MountingOptions) {
...
+        return args
+      }
+
+      if (
+        typeof args[0] === 'object' &&
+        args[0]['name'] in options?.global?.stubs
+      ) {
+        const name = args[0]['name']
+        // default stub
+        if (options?.global?.stubs[name] === true) {
```
dobromir-hristov, line 164: You are already in the conditional check, its there.

```suggestion
        if (options.global.stubs[name] === true) {
```

src/mount.ts
```diff
@@ -133,6 +137,46 @@ export function mount(originalComponent: any, options?: MountingOptions) {
...
+      }
+
+      // do not stub the VTU Parent component
+      if (typeof args[0] === 'object' && args[0]['name'] === 'VTU_COMPONENT') {
+        return args
+      }
+
+      if (
+        typeof args[0] === 'object' &&
+        args[0]['name'] in options?.global?.stubs
```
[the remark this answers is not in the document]
lmiller1990: Yeah this is probably a good idea, make it easier to find the stubs, I will implement this

src/mount.ts
lmiller1990: heh, copy pasted... thanks, I'll clean this up

src/mount.ts
JessicaSachs: I noticed this too. Am interested to hear about what you find.

JessicaSachs:
Did we discuss dropping support for render functions? If we’re dropping support for render functions, how does this affect JSX unit tests?

src/mount.ts
lmiller1990: @afontcu this has been done lmk if you have any feedback.

I have no problem making further changes if it improves the quality of the codebase

src/mount.ts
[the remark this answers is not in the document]
lmiller1990: I matched Vue 3, exact name, kebab case, pacal case (aka capital case)

https://github.com/vuejs/vue-next/blob/e8a866ec9945ec0464035be4c4c58d6212080a50/packages/runtime-core/src/helpers/resolveAssets.ts#L57

src/mount.ts
lmiller1990: Thanks for the suggestion, but I am not really a fan of this refactor. This variable can be a string (eg 'div'), or an object of options, so componentOptions is a bit misleading. I guess the ideal name is `htmlTagOrComponentOptions`. I will just put a comment here for now.

src/mount.ts
lmiller1990: very good

lmiller1990:
@JessicaSachs we support render functions. You can do this:

```
stubs: {
  Foo: { render() { return h('div') } }
```

etc. There is some tests for this.

src/mount.ts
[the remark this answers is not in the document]
dobromir-hristov: I also did that for the findComponent by name Branch. Need to sync these up later, so we don't duplicate :)

src/utils.ts
```diff
@@ -0,0 +1,5 @@
+import camelCase from 'lodash/camelCase'
+import upperFirst from 'lodash/upperFirst'
+import flow from 'lodash/flow'
+
+export const pascalCase = flow(camelCase, upperFirst)
```
lmiller1990, line 5: Lodash told me to do it this way https://github.com/lodash/lodash/issues/3102

src/mount.ts
afontcu: I think this is completely fine for now! 😄 Thanks for this. Once we've established a good foundations in terms of features, we can try and see the whole codebase using a bird's-eye view and see where we can improve readalibity and maintainability.

.github/workflows/ci.yml
```diff
@@ -24,9 +24,9 @@ jobs:
...
       with:
         node-version: ${{ matrix.node-version }}
-    - run: npm install
-    - run: npm run lint
-    - run: npm run build
-    - run: npm test
+    - run: yarn install
+    - run: yarn lint
+    - run: yarn test
+    - run: yarn build
```
afontcu, line 30: oops :D

src/mount.ts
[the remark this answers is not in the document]
lmiller1990: Sure, sounds good!

lmiller1990:
Going to merge this up. I fully expect to revisit this code soon, probably for `shallow`.

PR #62, opened 2020-04-13
feat: get method on wrapper

cexbrayat:
As it looks like the `get` method was not yet implemented, but was planned to, here is a little PR that adds it (I have so many tests that use it, it would make my life easier to test-drive this enw release 🙂 ).

I implemented the same logic it has in VTU v1.0, feedback welcome.

### DIFF
--- a/src/dom-wrapper.ts
+++ b/src/dom-wrapper.ts
@@ -50,6 +50,15 @@ export class DOMWrapper<ElementType extends Element> implements WrapperAPI {
     return new ErrorWrapper({ selector })
   }
 
+  get<T extends Element>(selector: string): DOMWrapper<T> {
+    const result = this.find<T>(selector)
+    if (result instanceof ErrorWrapper) {
+      throw new Error(`Unable to find ${selector} within: ${this.html()}`)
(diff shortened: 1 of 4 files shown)


cexbrayat:
I think the failing unit test is unrelated to this PR (I do have the same error on master).

tests/find.spec.ts
```diff
@@ -45,6 +45,32 @@ describe('find', () => {
   })
 })
 
+describe('get', () => {
+  test('returns the element if it exists', () => {
```
lmiller1990, line 49: nice job adding tests, can we put this in a `get.spec.ts` (each wrapper method has it's own file)

cexbrayat:
@lmiller1990 Thank you for the review! I skipped the faulty test, and isolated the get tests in their own file. PTAL

lmiller1990:
The broken test has an issue. I hope we fix it soon. 

Great job, looks good. I'll try and resolve the faulty test and probably be doing alphas fairly regularly (maybe every couple of days) so `get` will be available by the end of the week.

PR #63, opened 2020-04-13
fix: vm should be properly typed

cexbrayat:
This is on top of #64 

The types have been tweaked to avoid the need of casting like `(wrapper.vm as any).msg`.

The test showcases the correct behavior: you can reproduce the issue by adding the test to the master banch. It will:
- not compile without a `as any` cast
- fail as it returns the vm of the parent which has no `msg`, `isEnabled` fields and `toggle` function.

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -8,10 +8,11 @@ import {
   Plugin,
   Directive,
   Component,
-  reactive
+  reactive,
+  ComponentPublicInstance
 } from 'vue'
 
-import { createWrapper } from './vue-wrapper'
+import { createWrapper, VueWrapper } from './vue-wrapper'
 import { createEmitMixin } from './emitMixin'
 import { createDataMixin } from './dataMixin'
 import { MOUNT_ELEMENT_ID } from './constants'
@@ -37,7 +38,14 @@ interface MountingOptions {
   stubs?: Record<string, any>
 }
 
-export function mount(originalComponent: any, options?: MountingOptions) {
+export function mount<T extends any>(
+  originalComponent: any,
+  options?: MountingOptions
+): VueWrapper<any>
+export function mount<T extends ComponentPublicInstance>(
+  originalComponent: new () => T,
+  options?: MountingOptions
+): VueWrapper<T> {
   const component = { ...originalComponent }
 
   // Reset the document.body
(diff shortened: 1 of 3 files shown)


cexbrayat:
I splitted the two fixes in 2 separate PRs:
- #64 fixes the vm returned
- this PR improves the typings

lmiller1990:
Hey, thanks for helping out! Typing this correctly is something I am struggling with. I don't think this is entirely correct, though. `wrapper` needs to be able to accept objects like :

```ts
const Foo = {
  template: '<div>Foo</div>
}
```

And with this change, that won't work.

image

Full error:

```
const App: {
    render(): VNode<RendererNode, RendererElement>;
}
Argument of type '{ render(): VNode<RendererNode, RendererElement>; }' is not assignable to parameter of type 'new () => ComponentPublicInstance<{}, {}, {}, {}, {}, {}, {}>'.
  Type '{ render(): VNode<RendererNode, RendererElement>; }' provides no match for the signature 'new (): ComponentPublicInstance<{}, {}, {}, {}, {}, {}, {}>'.ts(2345)
```

cexbrayat:
I get your point. Maybe I can simply add an override, so people with `defineComponent` will have a proper typing and others will be at the same point as now. Most TS users are going to use it anyway, otherwise you don't have proper typings in your components, and we can probably improve later.

cexbrayat:
@lmiller1990 PTAL: I added a simple override instead of changing the signature. So everything works the same, but, if you use `defineComponent`, you get proper typings of `vm`

src/mount.ts
```diff
@@ -37,7 +38,14 @@ interface MountingOptions {
   stubs?: Record<string, any>
 }
 
-export function mount(originalComponent: any, options?: MountingOptions) {
+export function mount<T extends any>(
+  originalComponent: any,
+  options?: MountingOptions
+): VueWrapper<any>
```
lmiller1990, line 44: thanks @cexbrayat I will give this a quick test tmr and get it merged.

While we are here, can we get typings in the `props` mounting options? I was able to do that in an experiment I built [prior to this, found here](https://github.com/lmiller1990/vue-testing-framework/blob/bd83c4de5474aceaf95285635fa38908fa682b5d/packages/testing-framework/framework/mount.ts#L17)

Eg then we could do

```js
mount(Foo, {
  props: { 
    bar: 'sadf' // <= can we get typesafety here? that would be cool as
  }
})
```

If this is difficult to accomplish don't worry about it too much for now, but I agree, we should provide as many typings as possible.

lmiller1990: eg: `component: new () => ComponentPublicInstance<P>` (or something)

cexbrayat: Sure, I'll try to look into it soon, and I'll open a dedicated PR 👌

lmiller1990:
Awesome, love it. TS changed my life

src/mount.ts
cexbrayat: @lmiller1990 FYI I added some typings tests using tsd in #72 , improved the typing of vm some more in #73 and started working on properly typings the props options in #74

PR #64, opened 2020-04-13
fix: vm should be the vm of the tested component

cexbrayat:
It currently returns the vm of the parent element.
With this commit, it now returns the vm of the tested element.

The test showcases the correct behavior: you can reproduce the issue by adding the test to the master banch. It will fail as it returns the vm of the parent which has no msg, isEnabled fields and toggle function.

### DIFF
--- a/tests/vm.spec.ts
+++ b/tests/vm.spec.ts
@@ -1,0 +1,21 @@
+import { defineComponent, ref } from 'vue'
+
+import { mount } from '../src'
+
+describe('vm', () => {
+  it('returns the component vm', () => {
+    const Component = defineComponent({
+      template: '<div>{{ msg }}</div>',
+      setup() {
+        const msg = 'hello'
+        const isEnabled = ref(true)
+        return { msg, isEnabled }
+      }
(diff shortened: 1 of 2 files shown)


dobromir-hristov:
Yep I think I fixed this one in the `find` branch. Good catch.

tests/vm.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+import { mount } from '../src'
+
+describe('vm', () => {
+  it('returns the component vm', () => {
+    const Component = defineComponent({
+      template: '<div>{{ msg }}</div>',
+      setup() {
+        const msg = 'hello'
+        const isEnabled = ref(true)
+        const toggle = () => (isEnabled.value = !isEnabled.value)
```
dobromir-hristov, line 12: Toggle is not necessary here.

cexbrayat: 👍 Removed (it was a leftover of a more in-depth test for #64 )

lmiller1990:
Thanks!

PR #65, opened 2020-04-13
Add props accessor method

dobromir-hristov:
Mostly useful for the upcoming `stubs` and `findComponent` implementations.

### DIFF
--- a/tests/props.spec.ts
+++ b/tests/props.spec.ts
@@ -1,0 +1,14 @@
+import { mount } from '../src'
+import WithProps from './components/WithProps.vue'
+
+describe('props', () => {
+  it('returns a single prop applied to a component', () => {
+    const wrapper = mount(WithProps, { props: { msg: 'ABC' } })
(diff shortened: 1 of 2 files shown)


tests/props.spec.ts
```diff
@@ -0,0 +1,14 @@
...
+  it('returns a single prop applied to a component', () => {
+    const wrapper = mount(WithProps, { props: { msg: 'ABC' } })
+    expect(wrapper.props('msg')).toEqual('ABC')
+  })
+
+  it('returns all props applied to a component', () => {
+    const wrapper = mount(WithProps, { props: { msg: 'ABC' } })
+    expect(wrapper.props()).toEqual({ msg: 'ABC' })
+  })
+})
```
afontcu, line 14: Should we add a test to document what happens if the requested prop is not there, or if the target component has no props?

dobromir-hristov: Done.

PR #66, opened 2020-04-14
Find by component

dobromir-hristov:
This PR allows finding a Vue Component and returning a new VueWrapper with it's instance.

## Caveats:

* Returns only Vue components, and works only on VueWrapper. 
* You cant chain it from a DOMWrapper 
* setProps does not work on newly wrapped instances, from FindComponent. (which is fine)

## Reasoning behind this

Many codebases use Components, Ref or Name  to search for Vue instances, when doing tricky assertions, or mounting shallow.

### DIFF
--- a/src/emitMixin.ts
+++ b/src/emitMixin.ts
@@ -1,7 +1,7 @@
-import { getCurrentInstance, App } from 'vue'
+import { getCurrentInstance } from 'vue'
 
-export const attachEventListener = (vm: App) => {
-  const emitMixin = {
+export const attachEmitListener = () => {
+  return {
     beforeCreate() {
       let events: Record<string, unknown[]> = {}
       this.__emitted = events
@@ -15,6 +15,4 @@ export const attachEventListener = (vm: App) => {
       }
     }
   }
-
-  vm.mixin(emitMixin)
 }
--- a/src/emitMixin.ts
+++ b/src/emitMixin.ts
@@ -1,10 +1,11 @@
-import { getCurrentInstance } from 'vue'
+import { getCurrentInstance, App } from 'vue'
 
-export const createEmitMixin = () => {
-  const events: Record<string, unknown[]> = {}
-
+export const attachEventListener = (vm: App) => {
   const emitMixin = {
     beforeCreate() {
+      let events: Record<string, unknown[]> = {}
+      this.__emitted = events
+
       getCurrentInstance().emit = (event: string, ...args: unknown[]) => {
         events[event]
           ? (events[event] = [...events[event], [...args]])
@@ -15,8 +16,5 @@ export const createEmitMixin = () => {
     }
   }
 
-  return {
-    events,
-    emitMixin
-  }
+  vm.mixin(emitMixin)
 }
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -14,7 +14,7 @@ import {
 } from 'vue'
 
 import { createWrapper, VueWrapper } from './vue-wrapper'
-import { createEmitMixin } from './emitMixin'
+import { attachEventListener } from './emitMixin'
 import { createDataMixin } from './dataMixin'
 import { MOUNT_ELEMENT_ID } from './constants'
 import { stubComponents } from './stubs'
@@ -145,8 +145,7 @@ export function mount<T extends ComponentPublicInstance>(
   }
 
   // add tracking for emitted events
-  const { emitMixin, events } = createEmitMixin()
-  vm.mixin(emitMixin)
+  attachEventListener(vm)
 
   // stubs
   if (options?.global?.stubs) {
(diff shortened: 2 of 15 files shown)


lmiller1990:
woah, awesome job @dobromir-hristov !

I will play with this a bit tonight and see if I can find any bugs. Looks solid, very impressive that you managed to port the entire find api.

src/vue-wrapper.ts
```diff
@@ -2,45 +2,41 @@ import { ComponentPublicInstance, nextTick } from 'vue'
...
 
   constructor(
     vm: ComponentPublicInstance,
-    events: Record<string, unknown[]>,
-    setProps: (props: Record<string, any>) => void
+    setProps?: (props: Record<string, any>) => void
   ) {
-    this.__vm = vm
+    // TODO Remove cast after Vue releases the fix
+    this.rootVM = (vm.$root as any) as ComponentPublicInstance
```
dobromir-hristov, line 23: https://github.com/vuejs/vue-next/commit/8ae362400e8c259364ccd91f17b2887bb3b55a0b

src/emitMixin.ts
```diff
@@ -1,10 +1,11 @@
...
-
-export const createEmitMixin = () => {
-  const events: Record<string, unknown[]> = {}
+import { getCurrentInstance, App } from 'vue'
 
+export const attachEventListener = (vm: App) => {
   const emitMixin = {
     beforeCreate() {
+      let events: Record<string, unknown[]> = {}
+      this.__emitted = events
```
lmiller1990, line 7: How come we are doing this by attaching `__emitted` to the component instance instead of just returning the object and passing it to `mount` (like we did before?) Does this change/fix some behavior or just personal preference?

Imo it is better to avoid mutating the component instance - the less we mix VTU implementation details with the actual components, the better.

src/mount.ts
```diff
@@ -134,8 +135,7 @@ export function mount(originalComponent: any, options?: MountingOptions) {
   }
 
   // add tracking for emitted events
-  const { emitMixin, events } = createEmitMixin()
-  vm.mixin(emitMixin)
+  attachEventListener(vm)
```
lmiller1990, line 138: Same here - why the change? At least imo it is much better to use the `vm.mixin` - just seeing a function with an arg and no return values screams side effects/mutation, which is rarely a good thing.

src/utils/find.ts
```diff
@@ -0,0 +1,61 @@
...
+function matches(node: VNode, selector: FindAllComponentsSelector): boolean {
+  // do not return none Vue components
+  if (!node.component) return false
+
+  if (typeof selector === 'string') {
+    return node.el?.matches?.(selector)
+  }
+
+  if (typeof selector === 'object' && typeof node.type === 'object') {
+    if (selector.name && ('name' in node.type || 'displayName' in node.type)) {
```
lmiller1990, line 20: did not know about `displayName`, awesome

src/utils/matchName.ts
```diff
@@ -0,0 +1,14 @@
+import { camelize, capitalize } from '@vue/shared'
+
```
lmiller1990, line 2: ooo did not realize vue exported these guys, nice find

lmiller1990: this is 10x better than my implementation in stubs, I will refactor stubs once we merge this up

src/mount.ts
dobromir-hristov: This one is not necessary yeah. I was just trying to clean up. 

We could return the mixin and apply it here.

src/emitMixin.ts
dobromir-hristov: To be able to track each Vue Instance's events, I had to create the `events` variable for each instance's scope. The way it was made before would not have worked.

I know it is hacky, I also did not love the idea, but I could not figure a different way to track the events for each instance. 

Since `findComponent` can now return arbitrary Component Instance, it made sense to do so.

dobromir-hristov:
1. What part of `find` do you not get? Let me try to explain :)
2. We can do `find(Comp).vm.$emit`.

src/mount.ts
lmiller1990: That would be good! Thanks

src/emitMixin.ts
lmiller1990: Yep, figured that it was something like that, no problem

lmiller1990:
I think I need to use a debugger a bit and play around. Will ask if I still can't understand it.

If we can revert that one mixin (top level) happy to merge this up.

Awesome we can do `vm.$emit` should be very few breaking changes! Your codebase should upgrade easily.

lmiller1990:
Let's resolve this conflict and merge this up, I think a lot of people are waiting for this feature

PR #72, opened 2020-04-17
chore: add tsd tests for typings

cexbrayat:
This adds a new Ci task to check that the typings are as expected. It uses `tsd` (as done on `vue-next`), and adds a basic test in `test-dts`for the wrapper inference.

Note that I had to to @types/node to avoid an issue I fixed in Vue itself https://github.com/vuejs/vue-next/pull/968

This is a base PR for a follow-up that will improve the typings of `mount` and `props` options as @lmiller1990 suggested

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,3 +30,4 @@ jobs:
     - run: yarn build
       env:
         CI: true
+    - run: yarn tsd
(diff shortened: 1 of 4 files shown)


lmiller1990:
Great! Should I merge this or are you going to make more PRs targeting this?

lmiller1990:
Oops, I get it now.

PR #73, opened 2020-04-17
fix: improve vm types

cexbrayat:
This improves the signature of `mount` in order to have a proper type checking of `vm`.

On top of #72

PR #74, opened 2020-04-17
feat: properly type mount with props

cexbrayat:
This offers proper typings for the props given in the mounting options.
For example `mount(defineComponent({ props: { a: string}}), {props: {a: 2}})` fails to compile.

Only issue with this PR is that we need `ExtractPropTypes` to be public to properly type check the case `mount({ props: { a: String }}, { props: {a: 2}})`.

I'll open a discussion on vue-next and link back here (see https://github.com/vuejs/vue-next/pull/983).
Update: Evan merged it, so we'll be able to merge this one when beta.2 is released.

On top of #73

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -10,7 +10,11 @@ import {
   Directive,
   Component,
   reactive,
-  ComponentPublicInstance
+  ComponentPublicInstance,
+  ComponentOptionsWithObjectProps,
+  ComponentOptionsWithArrayProps,
+  ComponentOptionsWithoutProps,
+  ExtractPropTypes
 } from 'vue'
 
 import { createWrapper, VueWrapper } from './vue-wrapper'
@@ -25,9 +29,9 @@ import { stubComponents } from './stubs'
 
 type Slot = VNode | string | { render: Function }
 
-interface MountingOptions {
+interface MountingOptions<Props> {
   data?: () => Record<string, unknown>
-  props?: Record<string, any>
+  props?: Props
   slots?: {
     default?: Slot
     [key: string]: Slot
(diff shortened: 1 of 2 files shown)


lmiller1990:
Cool, will wait for beta 2.

src/mount.ts
```diff
@@ -41,14 +45,42 @@ interface MountingOptions {
...
+export function mount<
+  TestedComponent extends ComponentOptionsWithProps,
+  PublicProps extends ExtractPropTypes<TestedComponent['props']>
+>(
+  originalComponent: TestedComponent,
+  options?: MountingOptions<PublicProps>
+): VueWrapper<any>
+// Component declared with { props: [] }
+export function mount<
+  TestedComponent extends ComponentOptionsWithArrayProps,
```
lmiller1990, line 66: This is so ugly lol. I can't believe there isn't a better way without making 1/2 the file definitions.

cexbrayat: I agree ^^ I spent a long time to make this work, but I think there is no magic possible here: I had to do what Vue itself does... Maybe someone with a better TS knowledge will come with a better solution, but that should still be an improvement in the meantime 🙂 Contributors can play around with it easily: as long as the dts tests are OK, we should be good.

lmiller1990: cool. I wonder if we can at least hide the types in another file?

lmiller1990:
Will leave this open until beta-2, which fixes a few things. Looks good otherwise

cexbrayat:
@lmiller1990 I rebased on top of master, we should be good to merge 👌

Issue #57, opened 2020-04-13
Unable to add modifiers to triggered event


blacksonic:
I was trying to alter the keyup event as with the old version, but the new API doesn't have an options argument yet.

Here is a link to the source code, you can reproduce it by removing `.skip`
https://github.com/blacksonic/todomvc-vue-composition-api/blob/master/src/components/header/header.spec.js#L15

lmiller1990:
Yep, looks like a bug. If you don't destructure the `{ emit }` arg, and do `ctx.emit` instead, it works. 

I will investigate this.

lmiller1990:
This is a very easy (I think) first issue. Just port from VTU beta:

- trigger source in VTU beta is [here](https://github.com/vuejs/vue-test-utils/blob/f102a9a35e280f7afe6c1b62a5df7090ebdd6f98/packages/test-utils/src/wrapper.js#L676)
- you can steal most of this [this code](https://github.com/vuejs/vue-test-utils/blob/dev/packages/test-utils/src/create-dom-event.js) for the event keycodes: 
- [this](https://github.com/vuejs/vue-test-utils/blob/f102a9a35e280f7afe6c1b62a5df7090ebdd6f98/test/specs/wrapper/trigger.spec.js#L60) is the relevant test

SergiCL:
I'll probably have some time off in the next few days, if it's okay with you @blacksonic, I can try to solve this.

blacksonic:
@SergiCL go for it, I'll have time only on the weekend

lmiller1990:
Fixed in #77 by @SergiCL

PR #77, opened 2020-04-17
feat: add modifiers to trigger function

SergiCL:
Sorry for the delay, I've been a little more busy than expected.

This should be enough to solve #57. However, I would like to spend some more time (this weekend) to add more robust tests and to rethink some parts.

TODO:
- [x] Don't allow trigger in disabled elements
- [x] Add all old tests cases (from vue-test-utils)
- [x] Add more robust tests
- [x] Drop IE support?


src/create-dom-event.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  const event = document.createEvent('Event')
+  event.initEvent(eventType, bubbles, cancelable)
+  event['keyCode'] = keyCodesByKeyName[modifier]
+  return event
+}
+
+export default function createDOMEvent(
+  eventString: String,
+  options: Object = {}
```
lmiller1990, line 71: cool, you can type this like so:

```
interface TriggerOptions {
  key: 'enter' | 'keyup' | ...
}
```

if you aren't sure about TS i can look into this

src/create-dom-event.ts
```diff
@@ -0,0 +1,104 @@
...
+    cancelable: true,
+    bubbles: true
+  }
+
+  const eventParams = { eventType, modifier, meta, options }
+
+  const event =
+    typeof window.Event === 'function'
+      ? createEvent(eventParams)
+      : createEventForOldBrowsers(eventParams) // Fallback for IE10,11 - https://stackoverflow.com/questions/26596123
```
lmiller1990, line 85: It's 2020 - I wonder if we can drop IE support? @JessicaSachs do you guys run VTU in IE? I know you support old browsers (or did at your old org) - what do you think?

lmiller1990:
This looks great so far! Are there types for dom-event-types? eg `yarn add --dev @types/dom-event-types`.

Take your time - no rush :)

src/create-dom-event.ts
blacksonic: +1 on this as Vue 3 won't support IE 11 by default, there'll be compatibility build, this is more desired to be put into a separate compatibility file.

src/create-dom-event.ts
```diff
@@ -0,0 +1,104 @@
...
+    options.key ||
+    options.keyCode ||
+    options.code
+
+  return {
+    ...options, // What the user passed in as the second argument to #trigger
+    bubbles: meta.bubbles,
+    meta: meta.cancelable,
+    // Any derived options should go here
+    key: keyCode,
```
blacksonic, line 34: `key`, `keyCode`, `code` are 3 different things with different values, they shouldn't get the same value

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/KeyboardEvent

SergiCL: I have a couple of questions about this, surely someone with more knowledge of events than me can solve them.

If I am not mistaken, currently in 'vue-test-utils' this distinction is not being made. With the current implementation it seems that events are being fired without problem (more tests should be added to confirm that).

It is clear that the more real the function, the fewer unexpected errors, but do the three values always have to be defined? Right now one of them is correct and the rest are usually `undefined`.

Assuming that each key has the correct value, should the expected results of these values be as follows?

`trigger('keydown', { key: 'Enter' })` and `trigger('keydown.enter')`
- `key`: 'Enter', `code`: 'Enter', `keyCode`: 13

`trigger('keydown', { keyCode: '65' })`
- `key`: 'A', `code`: 'KeyA', `keyCode`: 65

lmiller1990: I don't know very much about the intricacies of `KeyboardEvent`.  I'd say write the tests how you would want to write them, and implement whatever we need to get them to pass. Ideally we should match whatever Vue lets you do. Eg, you can do `keyup.enter` in Vue, so we should support `trigger('keyup.enter')`.

I think this was something VTU beta did fairly well, if we can just replicate that (which it looks like you are doing) I think we will be in a good place.

We can definitely make improvements after that, I'd say for now we aim for basic feature parity with beta (we are pretty close) then we can dig deeper if there are any outstanding problems.

SergiCL: > We can definitely make improvements after that, I'd say for now we aim for basic feature parity with beta (we are pretty close) then we can dig deeper if there are any outstanding problems.

Perfect, I'll focus on a first version that works properly and then we'll think about what improvements can be made

src/create-dom-event.ts
dobromir-hristov: I say drop it.

src/create-dom-event.ts
SergiCL: Thanks, I'm not sure I know what you mean. If it's something simple and you want to take care of it yourself that's fine with me, go ahead with it 👍.

src/create-dom-event.ts
SergiCL: I agree with you. I will remove the support for IE, if it's needed in the future for some reason it can always be added again.

src/create-dom-event.ts
SergiCL: I added some types, I not sure if that's what you meant.

lmiller1990:
This is a very solid PR. I played with it and seems to cover everything - excellent job! I merged in the latest master with beta-2 and everything is passing.

SergiCL:
Thank you very much. You're doing a great job with this library. If I can help you with anything else, let me know.

PR #78, opened 2020-04-18
feat: add unmount to wrapper

pikax:
add `unmount` to the wrapper

Named `unmount` to be inline with vue3

Not throwing exception when you `unmount` multi times, because vue3 only warns 

https://github.com/vuejs/vue-next/blob/a51b05267231fb1770d6fc727bc276475c21d27f/packages/runtime-core/src/apiCreateApp.ts#L233-L239

### DIFF
--- a/src/vue-wrapper.ts
+++ b/src/vue-wrapper.ts
@@ -1,4 +1,4 @@
-import { ComponentPublicInstance, nextTick } from 'vue'
+import { ComponentPublicInstance, nextTick, App } from 'vue'
 import { ShapeFlags } from '@vue/shared'
 
 import { DOMWrapper } from './dom-wrapper'
@@ -9,15 +9,18 @@ import { MOUNT_ELEMENT_ID } from './constants'
 export class VueWrapper<T extends ComponentPublicInstance>
   implements WrapperAPI {
   private componentVM: T
+  private __app: App
   private __emitted: Record<string, unknown[]> = {}
   private __vm: ComponentPublicInstance
   private __setProps: (props: Record<string, any>) => void
 
   constructor(
+    app: App,
     vm: ComponentPublicInstance,
     events: Record<string, unknown[]>,
     setProps: (props: Record<string, any>) => void
   ) {
+    this.__app = app
     this.__vm = vm
     this.__setProps = setProps
     this.componentVM = this.__vm.$refs['VTU_COMPONENT'] as T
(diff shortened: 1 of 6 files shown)


lmiller1990:
Hi! Thanks for the PR.

We aren't looking to expand the API much right; but maintain the API from VTU beta. Is there a specific reason why you need this functionality in your app? 

If you are looking to contribute, we had a [`destroy` method](https://vue-test-utils.vuejs.org/api/wrapper/#destroy) on the VueWrapper. You could add that if you like.

pikax:
isn't `unmount` and `destroy` the same thing? 

I thought the `wrapper.destroy` method is called `destroy` to keep the same name as `vm.$destroy` in v2

In v3 the `destroy` was renamed to `unmount` [lifecycle-hooks](https://vue-composition-api-rfc.netlify.app/api.html#lifecycle-hooks)

We can have a `destroy` method that calls `unmount` with an deprecated `warning` (I think that's what will happen on v2.7)

lmiller1990:
Right, I didn't know they had renamed that hook. Makes sense.

In this case it's maintaining an existing API so I think this PR makes sense.

I don't think we need a deprecation warning; we are targeting Vue 3 exclusively, where there is no `destroy`.

Looks good to me. I will look for one more approval and merge this up. If you want, you can mark `destroy` as completed in the README with a note that it was renamed.

src/vue-wrapper.ts
```diff
@@ -109,12 +112,20 @@ export class VueWrapper<T extends ComponentPublicInstance>
     const rootElementWrapper = new DOMWrapper(this.element)
     return rootElementWrapper.trigger(eventString)
   }
+
+  unmount() {
+    if (this.parentElement) {
+      this.parentElement.removeChild(this.element)
```
afontcu, line 118: looks good!

I think we should add a test to ensure this removal is happening?

pikax:
Just merged with master, but I have a question, when we use findComponent it will return a new instance of VueWrapper, my question is do we allow to `unmount` that component instance?

```ts
const c = wrapper.findComponent({ ref: 'refChild' })
c.unmount()
```

If we allow should we unmount the root as well or only the component?

src/vue-wrapper.ts
pikax: done :)

lmiller1990:
I see 0 use case for unmounting a vue wrapper found via `findComponent`. Let's not do that unless a use case arises.

PR #80, opened 2020-04-19
Reractor and improve stub implementation.

dobromir-hristov:
## Summary

1. Cleans up stubbing implementation, passes props and attributes down to default stubs, so people can do `stub.attributes()` and `stub.props()`.
2. Cleans up the matching of stubs, using the one for `findComponent`, aka and the vue exported methods.
3. Make stub tags hyphenated, `FooBar` becomes `<foo-bar-stub/>` instead of `<foobar-stub/>`.
4. Allow stubbing functional components. (niche case)
5. Pass all props and for custom stub implementation.


dobromir-hristov:
Failing test is from stubbing a component without a name. I asked Evan if there is some way to find those, I have an idea, but still waiting.

lmiller1990:
Great job maintaining backwards compat with VTU beta! Not sure on the anonymous stub problem.

lmiller1990:
Are you planning to add more to this, or am I good to review and get this merged?

We can just `skip` the failing test for now. These changes seem good!

src/stubs.ts
```diff
@@ -32,26 +31,24 @@ const resolveComponentStubByName = (
...
 
 const isCommentOrFragment = (args: VNodeArgs) => typeof args[0] === 'symbol'
 
 const isParent = (args: VNodeArgs) =>
-  typeof args[0] === 'object' && args[0]['name'] === 'VTU_COMPONENT'
+  isComponent(args) && args[0]['name'] === 'VTU_COMPONENT'
 
 const isComponent = (args: VNodeArgs) => typeof args[0] === 'object'
 
+const isFunctionalComponent = ([type]: VNodeArgs) =>
```
lmiller1990, line 49: nice job figuring this out

src/stubs.ts
```diff
@@ -62,24 +59,34 @@ export function stubComponents(stubs: Record<any, any>) {
...
+          patchFlag,
+          dynamicProps
+        ]
       }
 
       // case 2: custom implementation
       if (typeof stub === 'object') {
-        return [stubs[name], {}]
+        // pass the props and children, for advanced stubbing
+        return [stubs[name], props, children, patchFlag, dynamicProps]
```
lmiller1990, line 89: nice! I like the fact we can keep these advanced tools without increasing complexity.

src/utils/find.ts
```diff
@@ -17,6 +17,8 @@ function matches(node: VNode, selector: FindAllComponentsSelector): boolean {
   }
 
   if (typeof selector === 'object' && typeof node.type === 'object') {
+    if (selector === node.type) return true
```
lmiller1990, line 20: what does this do? in what cases does this eval to true?

dobromir-hristov: If you pass the component definition and it's identical to the one the component was created with. Like imported SFC

dobromir-hristov:
Yeah Evan said he fixed that case, but I could not verify. We can merge and I will get back to it later. It's an edge case anyway

src/utils/find.ts
lmiller1990: Right, ok. so many ways to make a component in vue haha

src/vue-wrapper.ts
```diff
@@ -114,8 +114,12 @@ export class VueWrapper<T extends ComponentPublicInstance>
 
   findComponent(selector: FindComponentSelector): VueWrapper<T> | ErrorWrapper {
     if (typeof selector === 'object' && 'ref' in selector) {
-      return createWrapper(null, this.vm.$refs[selector.ref] as T)
+      const result = this.vm.$refs[selector.ref]
+      return result
+        ? createWrapper(null, this.vm.$refs[selector.ref] as T)
```
afontcu, line 119: would this be possible?

```suggestion
        ? createWrapper(null, result as T)
```

PR #81, opened 2020-04-20
feat: add options.attachTo

pikax:
Tests from [here](https://github.com/vuejs/vue-test-utils/pull/1492)

SSR test not ported, because I'm not sure how to test it.

### Note

The reset of the document body now only happens if you don't pass a child, maybe this should be passed as an option?


lmiller1990:
Hm, I think we need to reset the body for the majority of cases (except this one).

VTU beta did not mount on the DOM by default. Only with this option. Vue 3 doesn't seem to be able to `mount` without an element and a DOM... I wonder what the implications of this are.

Code seems fine, I will play with it a bit before merging. thanks for porting this!

src/mount.ts
```diff
@@ -26,6 +26,7 @@ import {
   MOUNT_PARENT_NAME
 } from './constants'
 import { stubComponents } from './stubs'
+import { isString } from './utils'
```
lmiller1990, line 29: Maybe we can use `isString` from lodash. You would need to update the `rollup.config.js` too.

lmiller1990:
Seems good! Thanks 👍

PR #82, opened 2020-04-21
POC: Plugin interface with wrapper in closure

JessicaSachs:
This POC aims to add on to the current work surrounding plugins. I used the branch at #55 to build some plugins. I struggled with the existing API. I was limited from using fat-arrow functions due to the wrapper only being available on `this`. I wanted a closure where the wrapper would be passed in.

## Importance of Plugins
Our intension is that plugins will be a means for users to...
1. create short-hand aliases 🚀
2. implement helper functions 🛠
3. customize Vue Test Utils as they need to 👩🏻‍🎨

## The pattern
Certain entities are Pluggable (VueWrapper, DOMWrapper). You can **install** plugins, which may return objects to be extended onto the instance of wrappers being mounted. This pattern looks like `data` or `setup` functions in Vue core and should feel familiar.

## Basic example

This plugin will add the `width` property and `myMethod` once the wrapper is mounted.

```js
const plugin = (wrapper) => {
  return {
    width: 200,
    myMethod() {}
  }
}

config.plugins.VueWrapper.install(plugin)
```

Here is a plugin that will provide a new method to clean up the output of `text()`. (The use case is based off of [jest-dom](https://github.com/testing-library/jest-dom/blob/cdc8a06fb2709367427f02459ca4e3e705e00431/src/to-have-text-content.js#L11-L13))

```js
// A plugin to add optional whitespace stripping to VueWrapper.text()
const normalize = text =>  text.replace(/\s+/g, ' ').trim()

// HTML-friendly text matcher. Based off of jest-dom
const friendlyTextPlugin = (wrapper) => ({
  normalizeText: () => normalize(wrapper.text()),
})

// installing it into VueWrapper

config.plugins.VueWrapper.install(friendlyTextPlugin)

const wrapper = mount({ template: `<div>    Foo</div>` })
wrapper.normalizeText() // 'Foo'
```

### Weak spots:
1. Install functions are executed every mount
  * Perhaps this is ok. If most of the plugins' work is just setting up functions to be called later, this might not be a concern
2. I don't think Pluggable as a class is the right abstraction
  * I'm really new to Typescript 😬

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,6 +1,40 @@
+class Pluggable {
+  installedPlugins: any
+  constructor() {
+    this.installedPlugins = []
+  }
+
+  install(handler, options = {}) {
+    if (typeof handler !== 'function') {
+      console.error('plugin.install must receive a function')
+      handler = () => ({})
+    }
+    this.installedPlugins.push({ handler, options })
+  }
+
+  extend(instance) {
+    const invokeSetup = (plugin) => plugin.handler(instance) // invoke the setup method passed to install
+    const bindProperty = ([property, value]: [string, any]) => {
+      instance[property] =
+        typeof value === 'function' ? value.bind(instance) : value
+    }
+    const addAllPropertiesFromSetup = (setupResult) => {
+      setupResult = typeof setupResult === 'object' ? setupResult : {}
+      Object.entries(setupResult).forEach(bindProperty)
+    }
+
+    this.installedPlugins.map(invokeSetup).forEach(addAllPropertiesFromSetup)
+  }
+
+  /** For testing */
+  reset() {
+    this.installedPlugins = []
+  }
+}
+
 export const config = {
   plugins: {
-    VueWrapper: {} as any,
-    DOMWrapper: {}
+    VueWrapper: new Pluggable(),
+    DOMWrapper: new Pluggable()
   }
 }
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,0 +1,6 @@
+export const config = {
+  plugins: {
+    VueWrapper: {} as any,
+    DOMWrapper: {}
+  }
+}
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,3 +1,13 @@
+import { GlobalMountOptions } from './types'
+
+interface GlobalConfigOptions {
+  global: GlobalMountOptions
+  plugins: {
+    VueWrapper: Pluggable
+    DOMWrapper: Pluggable
+  }
+}
+
 class Pluggable {
   installedPlugins: any
   constructor() {
@@ -32,7 +42,8 @@ class Pluggable {
   }
 }
 
-export const config = {
+export const config: GlobalConfigOptions = {
+  global: {},
   plugins: {
     VueWrapper: new Pluggable(),
     DOMWrapper: new Pluggable()
--- a/tests/features/plugins.spec.ts
+++ b/tests/features/plugins.spec.ts
@@ -4,6 +4,7 @@ declare module '../../src/vue-wrapper' {
   interface VueWrapper {
     width(): number
     $el: Element
+    myMethod(): void
   }
 }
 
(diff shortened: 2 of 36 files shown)


src/config.ts
```diff
@@ -0,0 +1,40 @@
+class Pluggable {
+  installedPlugins: any
+  constructor() {
+    this.installedPlugins = []
+  }
+
+  install(handler, options = {}) {
+    if (typeof handler !== 'function') {
+      console.error('plugin.install must receive a function')
+      handler = () => ({})
```
JessicaSachs, line 10: Should decide if we're going to exit the process here or quietly swallow the misconfigured plugin

tests/features/plugins.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+      const plugin = () => ({ width })
+      config.plugins.VueWrapper.install(plugin)
+      const wrapper = mountComponent()
+      expect(wrapper).toHaveProperty('width', width)
+    })
+
+    it('receives the wrapper inside the plugin setup', () => {
+      const plugin = (wrapper) => {
+        return {
+          $el: wrapper.element // simple aliases
```
dobromir-hristov, line 31: Will we have access to private properties on the instance? Like the rootVM?

JessicaSachs: Yeah, your original example will work. You can get access to the `ComponentPublicInstance` under `componentVM` which is private.

JessicaSachs:
```js
const namePlugin = wrapper => ({ lowerCaseName: wrapper.componentVM.$.type.name })
config.plugins.VueWrapper.install(namePlugin)
const wrapper = mount({ template: `<h1>Hello</h1>`, name: 'My_Component' })
wrapper.lowerCaseName() // 'my_component'
```

src/config.ts
lmiller1990: I think failing loudly is usually a good idea

lmiller1990: The louder the better, ideally with "you screwed up and here's where and how to fix it"

lmiller1990:
I resolved conflicts, and deployed latest master to my npm account and played around and so far this seems to work great 👍  nice job @JessicaSachs 

image

src/config.ts
JessicaSachs: Yeah. I'm fine with that. I wish I had a plugin name to point people at.

lmiller1990: hmmm didn't think of that. 🤔

tests/features/plugins.spec.ts
```diff
@@ -0,0 +1,71 @@
+import { mount, config } from '../../src'
+import { WrapperAPI } from '../../src/types'
+
+declare module '../../src/vue-wrapper' {
+  // @ts-ignore
+  interface VueWrapper {
```
cexbrayat, line 6: `@ts-ignore` should be removed here: you should just need to have the same signature than `VueWrapper`, i.e `VueWrapper<T extends ComponentPublicInstance>` and TS will be happy to merge the declarations.
You should also be able to remove all the other `@ts-ignore` in the file.

JessicaSachs: 🤔 I will give this a try

lmiller1990: To no-one's surprised I changed the types based on @cexbrayat 's recommendation and it worked

afontcu:
> Looks good to me! CI is passing and we have no conflicts. We need some docs then we can merge this (this seems like an important feature to document and document well).

I've been trying to keep docs up to date with vtu-next releases, just to make sure we don't document any POC or temporary feature (I think this is a sensible way of handling docs until we're able to merge repos). What I mean by this is that feel free to merge this up if you feel confident enough with the implementation and release a new alpha, and then we can work out the docs part :)

lmiller1990:
Ok, let's do it. Great job all

JessicaSachs:
Nice! Thanks :-) I'll pick up the docs this weekend.

PR #83, opened 2020-04-21
fix: remove globalProperties

cexbrayat:
Fixes something I spotted after #53 was merged

`globalProperties` is not supported AFAIK and was removed from the original PR except in the `mergeGlobalProperties` utils.
With the various parameters correctly typed, the code was throwing a compilation error as TS can spot that `globalProperties` is not a key of `GlobalMountOptions`.


lmiller1990:
Good catch

PR #84, opened 2020-04-21
fix: emit should propagate the event

cexbrayat:
This commit fixes an issue where emit is replaced but does not does its job, so a component emitting an event never really emits it, and a parent component never receives it.
The unit test added reproduces the issue.


lmiller1990:
Thanks! You have made this codebase a lot better in the last week or two.

PR #86, opened 2020-04-22
Improve build step, support iife build for browser usage

lmiller1990:
Now generates an iife build, that you can use (assumes `Vue` and `_` are globally available).

I tested it like this

```
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <title></title>
  <script src="https://cdnjs.cloudflare.com/ajax/libs/lodash.js/4.17.15/lodash.js"></script>
  <script>
  </script>
    <script src="./vue.global.js"></script>
    <script src="./dist/vue-test-utils.browser.js"></script>
</head>
<body>
 

</body>
</html>
```

And I was able to access `VueTestUtils` globally and do some basic stuff.

I am still learning rollup, so this config can probably be improved as we go. I will add some e2e tests for this soon, too.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,16 +1,9 @@
-import camelCase from 'lodash/camelCase'
-import upperFirst from 'lodash/upperFirst'
-import kebabCase from 'lodash/kebabCase'
-import flow from 'lodash/flow'
 import isString from 'lodash/isString'
 import mergeWith from 'lodash/mergeWith'
+
 import { GlobalMountOptions } from './types'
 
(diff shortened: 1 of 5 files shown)


src/utils.ts
```diff
@@ -1,16 +1,9 @@
-import camelCase from 'lodash/camelCase'
-import upperFirst from 'lodash/upperFirst'
-import kebabCase from 'lodash/kebabCase'
-import flow from 'lodash/flow'
```
lmiller1990, line 4: unused

PR #87, opened 2020-04-23
repro: setProps breaks reactivity

cexbrayat:
This adds a failing unit-test to showcase how `setProps` can break reactivity.
The example works fine in an application, but can't be unit tested with the current `setProps`, so I think this is an issue no?

### DIFF
--- a/tests/setProps.spec.ts
+++ b/tests/setProps.spec.ts
@@ -102,6 +102,30 @@ describe('setProps', () => {
     expect(wrapper.html()).toContain('Foo is: qux. Foobar is: qux-bar')
   })
 
+  it('does not break reactivity', async () => {
+    const useFoobar = (foo) => computed(() => `${foo}-bar`)
+    const Foo = defineComponent({
+      props: {
+        foo: { type: String }
+      },
+      setup(props) {
+        const foobar = useFoobar(props.foo)
+        return () =>
+          h('div', `Foo is: ${props.foo}. Foobar is: ${foobar.value}`)
+      }
+    })
+    const wrapper = mount(Foo, {
+      props: {
+        foo: 'foo'
+      }
+    })
+    expect(wrapper.html()).toContain('Foo is: foo. Foobar is: foo-bar')
+
+    await wrapper.setProps({ foo: 'qux' })
+
+    expect(wrapper.html()).toContain('Foo is: qux. Foobar is: qux-bar')
+  })
+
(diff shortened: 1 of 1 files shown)


lmiller1990:
And so the `setProps` bugs begin... this was a problem in beta as well.

I will investigate this now! Thanks.

lmiller1990:
I just tried something very similar to this in my app and it did not work. Did you get this working in your app?

lmiller1990:
Here is how I tried to reproduce it:

```
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <title></title>
  <script src="./vue.global.js"></script>
</head>
<body>
  <div id="app"></div> 

  <script>
    const { ref, h, computed, defineComponent, createApp } = Vue

    const useFoobar = (foo) => computed(() => `${foo}-bar`)

    const Foo = defineComponent({
      props: {
        foo: { type: String }
      },
      setup(props) {
        const foobar = useFoobar(props.foo)
        return () =>
          h('div', `Foo is: ${props.foo}. Foobar is: ${foobar.value}`)
      }
    })

    const App = {
      components: { Foo },
      setup() {
        const foo = ref('foo')
        setTimeout(() => {
          foo.value = 'qux'
        }, 1000)
        return () => h(Foo, { foo: foo.value })
      }
    }

    const app = createApp(App)
    app.mount('#app')
  </script>
</body>
</html>
```

I put this in `index.html` and got a copy of `vue.global.js` from the `vue-next` repo by running `yarn build` and it was created in the `packages/vue/dist` directory.

Please correct me if I am wrong but it appears you cannot use `computed` in this manner. I guess this is because you pass a string primitive to `computed` and it cannot track reactivity unless you pass a `ref` or equivalent. I do not really understand Vue's reactivity system, though, so take this with a grain of salt.

Either way if you can show me a reproduction where this works in a real app but not VTU I will spend more time investigating it. 

I greatly appreciate your contributions.

cexbrayat:
You're right I can't reproduce either. The real example I have is more complex, so maybe I have something at play that make it works otherwise. Let's close it then, and I'll reopen if needed. Thank you for investigating!

lmiller1990:
No problem - let me know if you are able to find any other bugs. I love the new reactivity API for it's excellent TS integration, but it definitely has it's trade-offs - the user needs to know how it works and it's limitations quite well. It's definitely a step forward, though, imo!

PR #88, opened 2020-04-24
repro: findComponent fails inside a Suspense

cexbrayat:
This adds a failing unit-test to showcase how `findComponent` fails if the search component is inside a `Suspense`.
I was naively expecting it to work, do you think we should support this?

### DIFF
--- a/src/utils/find.ts
+++ b/src/utils/find.ts
@@ -46,8 +46,12 @@ function findAllVNodes(vnode: VNode, selector: any): VNode[] {
   const nodes = [vnode]
   while (nodes.length) {
     const node = nodes.shift()
+    // match direct children
     aggregateChildren(nodes, node.children)
+    // match children of the wrapping component
     aggregateChildren(nodes, node.component?.subTree.children)
+    // match children if component is Suspense
+    aggregateChildren(nodes, node.suspense?.subTree.children)
     if (matches(node, selector)) {
       matchingNodes.push(node)
     }
--- a/tests/find.spec.ts
+++ b/tests/find.spec.ts
@@ -2,7 +2,6 @@ import { defineComponent, h } from 'vue'
 
 import { mount } from '../src'
 import SuspenseComponent from './components/Suspense.vue'
-import Hello from './components/Hello.vue'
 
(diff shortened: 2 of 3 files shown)


lmiller1990:
I have written tests in my other apps that work fine w/ Suspense. I think this is related to findComponent. 

We should support this.

lmiller1990:
Edit edit: the blow by blow commentary continues - `children` in a `Suspense` are not an array, but a key:value pair:

```
{
      default: [Function: renderFnWithContext],
      fallback: [Function: renderFnWithContext],
      _: 1
}
```

This should be fixable. I am worried what other edges cases are involved here, though. Excellent work catching this early. I definitely recommend asserting against the DOM where possible, as opposed to components

lmiller1990:
As a quick sanity check, I added a test for `<Teleport />` and it worked.

dobromir-hristov:
Done. Suspense is pretty much still Alpha and not complete.

lmiller1990:
wow, nice @dobromir-hristov. I tried something like that but it did not work 🤔 

Great!

lmiller1990:
Actually @dobromir-hristov I don't think this is actually going to work.

I changed the test:

```
    const wrapper = mount(SuspenseComponent)
    expect(wrapper.findComponent(AsyncComponent).exists()).toBe(false)
    await nextTick()
    expect(wrapper.findComponent(AsyncComponent).exists()).toBe(true)
```

This is failing - because both default and fallback are in the Suspense tree. Any way to know which one is visible (other than the obvious solution of the DOM).

dobromir-hristov:
hm... Will check it out.

cexbrayat:
@dobromir-hristov @lmiller1990 Where are we on this PR?

dobromir-hristov:
I played with it but got nowhere... Will try again this week.

cexbrayat:
As this was bugging me, I read the code of `vue-next` and now understand better how suspense is implemented.
Based on your feedback, I opened #168 (I did not copy your Teleport test @lmiller1990 , feel free to open a PR to add it!) if you have time to review

PR #89, opened 2020-04-24
feat: getComponent

cexbrayat:
This introduces a `getComponent` which returns a `VueWrapper` or throws.
It offers two benefits:
- a symetric API for components/elements with `findComponent/findAllComponents/getComponent`/`find/findAll/get`
- a better experience for TS users (as `findComponent` returns a union type, you can't chain certain things until you disambiguated the type manually).

I also refactored the signatures of `find` to offer proper overloads (the current version breaks tests compilation in a real app, i hope this fixes it). I added TSD tests to ensure the inference is working properly.

### DIFF
--- a/src/vue-wrapper.ts
+++ b/src/vue-wrapper.ts
@@ -112,7 +112,15 @@ export class VueWrapper<T extends ComponentPublicInstance>
     return result
   }
 
-  findComponent(selector: FindComponentSelector): VueWrapper<T> | ErrorWrapper {
+  findComponent<T extends ComponentPublicInstance>(
+    selector: new () => T
+  ): VueWrapper<T> | ErrorWrapper
+  findComponent<T extends ComponentPublicInstance>(
+    selector: FindComponentSelector
+  ): VueWrapper<T> | ErrorWrapper
+  findComponent<T extends ComponentPublicInstance>(
+    selector: any
+  ): VueWrapper<T> | ErrorWrapper {
     if (typeof selector === 'object' && 'ref' in selector) {
       const result = this.vm.$refs[selector.ref]
       return result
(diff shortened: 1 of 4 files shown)


lmiller1990:
I had this exact thought the other day - I think this makes sense. I see you didn't include `getAllComponents` - this makes sense too, there is no real use case for that.

I will wait for at least one more approval on this PR, to let others have input as well 👍 

An aside: testing types is a very cool concept I just learned about recently.

src/vue-wrapper.ts
```diff
@@ -125,6 +133,25 @@ export class VueWrapper<T extends ComponentPublicInstance>
...
+  getComponent<T extends ComponentPublicInstance>(
+    selector: FindComponentSelector
+  ): VueWrapper<T>
+  getComponent<T extends ComponentPublicInstance>(
+    selector: any
+  ): VueWrapper<T> {
+    const result = this.findComponent(selector)
+    if (result instanceof ErrorWrapper) {
+      throw new Error(
+        `Unable to find component with selector ${selector} within: ${this.html()}`
```
afontcu, line 148: This is really a small nitpick, but got bitten by it the other day 😅 the word "find" in the error message of "get" tricked me for a bit

```suggestion
        `Unable to get component with selector ${selector} within: ${this.html()}`
```

tests/getComponent.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+    const wrapper = mount(compA)
+    jest.spyOn(wrapper, 'findComponent').mockReturnThis()
+    wrapper.getComponent('.domElement')
+    expect(wrapper.findComponent).toHaveBeenCalledWith('.domElement')
+  })
+
+  it('should throw if not found', () => {
+    const wrapper = mount(compA)
+    expect(() => wrapper.getComponent('.domElement')).toThrowError(
+      'Unable to find component with selector .domElement within: <div class="A"></div>'
```
afontcu, line 20:
```suggestion
      'Unable to get component with selector .domElement within: <div class="A"></div>'
```

src/vue-wrapper.ts
cexbrayat: Good point 🙂 And I think we should change the message for `get` as well to be coherent.

cexbrayat:
@lmiller1990 👍 Thank you for the quick review.

@afontcu I pushed a new commit with your suggestion for `getComponent`, and changed it in `get` as well.

lmiller1990:
Nice, I will do a new alpha this weekend.

Issue #92, opened 2020-04-27
Refactor ErrorWrapper to DOMWrapperError and VueWrapperError


lmiller1990:
Now we have separated find and findComponent, and considering DOMWrapper and VueWrapper have different APIs, I think ErrorWrapper should be separated accordingly, too.

I ran into this complication when writing a plugin. Better typedefs will make extending VTU easier.

lmiller1990:
#100

Issue #91, opened 2020-04-27
Export DOMWrapper


lmiller1990:
Useful for plugin authors

lmiller1990:
#100

PR #93, opened 2020-04-27
Split ErrorWrapper into DOMWrapper DOMWrapperError

lmiller1990:
#91 and #92 

`VueWrapper` and `DOMWrapper` are fairly different; using a single `ErrorWrapper` for both makes it a little difficult to have proper typings on plugins, and also makes it hard to define a single `interface` they can all implement.

This PR introduces a change where when an `ErrorWrapper` would have been returned in place of a `DOMWrapper` (eg, a failed `find`), now you get a `DOMWrapperError`. Both `DOMWrapper` and `DOMWrapperError` share an `interface`.

In a future PR I would like to do the same for `VueWrapper` - changing `ErrorWrapper` to be `VueWrapperError` and defining a single interface they can both `implement`. I want to keep the PR small so I just did DOMWrapper for now. `VueWrapper` has quite a bit more complicated typings.

This main benefit is typings; this PR will allow me to write [the following plugin](https://github.com/vuejs/vue-test-utils-next-docs/pull/15/files#r416565248) with proper type-safety, for example. 

This PR does not introduce a breaking change and there is no impact on the end user.


cexbrayat:
@lmiller1990 If I understand your question correctly, I would say:
- define the overloads in the interface
- all classes implement the interface

lmiller1990:
@cexbrayat I thought so, will do this, thanks.

lmiller1990:
@cexbrayat quick review if you have some time on your hands :) for some reason I couldn't add you as a reviewer 🤔

src/types.ts
```diff
@@ -2,17 +2,23 @@ import { Component, ComponentOptions, Directive, Plugin } from 'vue'
 
 import { DOMWrapper } from './dom-wrapper'
 import { ErrorWrapper } from './error-wrapper'
+import { DOMWrapperError } from './dom-wrapper-error'
 
-export interface WrapperAPI {
+export interface VueWrapperAPI<T> {
+  element: Element
```
lmiller1990, line 8: TODO: this in another PR

cexbrayat:
@lmiller1990 I think you can only assign users that have been appointed collaborators of the project or something like that. I had to be added to the Angular repo for this reason IIRC.
I'll take a look!

src/dom-wrapper-error.ts
```diff
@@ -0,0 +1,58 @@
+import { FindComponentSelector, DOMWrapperAPI } from './types'
+
+interface Options {
+  selector: FindComponentSelector
+}
+
+export class DOMWrapperError implements DOMWrapperAPI {
+  private selector: FindComponentSelector
```
cexbrayat, line 8: Do we need this or can it be removed? I think we would be better off, as it doesn't look used? Same for the constructor below, and we could get rid of the `Options` interface

src/types.ts
```diff
@@ -2,17 +2,23 @@ import { Component, ComponentOptions, Directive, Plugin } from 'vue'
...
+}
+
+export interface DOMWrapperAPI {
   attributes: (key?: string) => string | Record<string, string>
   classes: (className?: string) => string[] | boolean | ErrorWrapper
   readonly element: Element
   exists: () => boolean
   get<T extends Element>(selector: string): DOMWrapper<T>
-  find<T extends Element>(selector: string): DOMWrapper<T> | ErrorWrapper
+  find<T extends Element>(selector: string): DOMWrapper<T> | DOMWrapperError
```
cexbrayat, line 17: I'm wondering if we should not define all the overload here to be sure that each class properly implement them. Same for get and findAll of course

src/dom-wrapper-error.ts
```diff
@@ -0,0 +1,58 @@
...
+  element: null
```
cexbrayat, line 9: Same for element, but I guess this is part of your followup PR.

src/dom-wrapper-error.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  constructor({ selector }: Options) {
+    this.selector = selector
+  }
+
+  wrapperError(method: string): Error {
+    return Error(`Cannot call ${method} on an empty wrapper.`)
```
cexbrayat, line 16: `empty wrapper` -> `empty DOM wrapper` maybe?

src/dom-wrapper-error.ts
```diff
@@ -0,0 +1,58 @@
+import { FindComponentSelector, DOMWrapperAPI } from './types'
+
+interface Options {
+  selector: FindComponentSelector
+}
+
+export class DOMWrapperError implements DOMWrapperAPI {
```
cexbrayat, line 7: nit: add some doc on when this class is created? I'm starting to get lost between VueWrapper, VuerapperAPI, DOMWrapper, ErrorWrapper, DOmWrapperAPi 😅

src/error-wrapper.ts
```diff
@@ -1,10 +1,10 @@
-import { FindComponentSelector } from './types'
+import { FindComponentSelector, DOMWrapperAPI } from './types'
 
 interface Options {
   selector: FindComponentSelector
 }
 
-export class ErrorWrapper {
+export class ErrorWrapper implements DOMWrapperAPI {
```
cexbrayat, line 7: nit: add some docs as well?

src/types.ts
lmiller1990: This was what I was thinking too - I will make this change.

src/dom-wrapper-error.ts
lmiller1990: I don't think this is used. I will look into this, be good to remove it.

src/dom-wrapper-error.ts
lmiller1990: Yes we definitely need to update docs. The fact we are in alpha and kind of changing this a lot is kind of the reason docs are not updated - I think once we get to beta we can then start to lock down the API and really get a handle on docs (VTU for Vue 2 were not great, we should def spend a good amount of time on this).

src/dom-wrapper-error.ts
lmiller1990: done

src/error-wrapper.ts
lmiller1990: When I finish refactoring and we are happy with the API :D

lmiller1990:
I have made the changes @cexbrayat. Questions:

- so we duplicate the types for `find` etc in both the interface and the implementation? maybe not ideal since I don't expect any other wrappers to use these types except `DOMWrapperError` which just throws errors for everything. I guess this makes sense if many classes implement something, perhaps now in this case (since no others except DOMWrapper and the error wrapper will do so?)
- [this overload](https://github.com/vuejs/vue-test-utils-next/blob/f92cf3f4a7d9729a96577988002d497e0513c8fa/src/dom-wrapper.ts#L46) and the one under it do not seem to do anything - if I delete them, everything seems find and `tsd` passes too. Are we missing a test?

lmiller1990:
Let's put this on hold - I think there is a better way to do this.

PR #94, opened 2020-04-27
feat: run test using build file

### LINKED ISSUE #68: e2e tests on final bundle
We should have some additional tests to the unit tests. There is the [example app](https://github.com/vuejs/vue-test-utils-next/tree/master/examples).

Something like:

- `yarn build`
- run the test app using the final bundle

This will make sure there are no surprises for the end user.

pikax:
fix #68 

Do testing with the built files

Currently only doing tests for `cjs`, more can be added easily

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,4 +30,5 @@ jobs:
     - run: yarn build
       env:
         CI: true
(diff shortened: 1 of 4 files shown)


lmiller1990:
great, thanks!

PR #100, opened 2020-04-30
refactor: eliminate need for error wrapper class

lmiller1990:
Greatly simplify the concept of `ErrorWrapper` and eliminate the need for the `WrapperAPI` interfaces (since they are not implemented by more than one class).

It's everyone favorite kind of PR - one that has more deletions than insertions!

#91  and #92

### DIFF
--- a/src/vue-wrapper.ts
+++ b/src/vue-wrapper.ts
@@ -4,7 +4,7 @@ import { config } from './config'
 
 import { DOMWrapper } from './dom-wrapper'
 import { FindAllComponentsSelector, FindComponentSelector } from './types'
-import { createWrapperError, createVueWrapperError } from './error-wrapper'
+import { createWrapperError } from './error-wrapper'
 import { TriggerOptions } from './create-dom-event'
 import { find } from './utils/find'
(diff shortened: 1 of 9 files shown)


src/vue-wrapper.ts
```diff
@@ -109,32 +103,35 @@ export class VueWrapper<T extends ComponentPublicInstance>
...
     }
 
     const result = find(this.vm.$.subTree, selector)
-    if (!result.length) return new ErrorWrapper({ selector })
-    return createWrapper(null, result[0])
+    if (result.length) {
+      return createWrapper(null, result[0])
+    }
+
+    return createWrapperError('VueWrapper')
```
dobromir-hristov, line 134: I see we lost the selector on the error wrapper. What was it used for?

lmiller1990: Turns out it wasn't used for anything - maybe it was previously, but in the current master is it not used. So I just deleted it.

cexbrayat:
I'm late to the party, but I think this API is much more sound 👌

PR #101, opened 2020-04-30
types: Assign the correct typing for vnodes

pikax:
A more correct type would be:
```ts
type VNodeArgs = Parameters<Parameters<typeof transformVNodeArgs>[0]>[0]
```
But it's a bit harder to read

### DIFF
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -1,4 +1,4 @@
-import { transformVNodeArgs, h } from 'vue'
+import { transformVNodeArgs, h, createVNode } from 'vue'
 import { hyphenate } from '@vue/shared'
 import { matchName } from './utils/matchName'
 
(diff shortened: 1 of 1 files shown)


lmiller1990:
Will the `type VNodeArgs = Parameters<Parameters<typeof transformVNodeArgs>[0]>[0]` give us any additional benefits? If it will, we should use it. Looks (at this point) it won't, though?

pikax:
Sorry missed the email with the comment.


> Will the `type VNodeArgs = Parameters<Parameters<typeof transformVNodeArgs>[0]>[0]` give us any additional benefits? If it will, we should use it. Looks (at this point) it won't, though?

Using that will make typescript infer always the correct parameter of the function, even if in the future that method changes the parameter type, personally for a maintainability standpoint that's just much harder to maintainers to make any change in the future. I think is quite unlikely for the parameter change type in the future and if it changes we will get a type error on the tests

Issue #69, opened 2020-04-16
Mounting only the slots of stubbed child components


TheJaredWilcurt:
**Original:** https://github.com/vuejs/vue-test-utils/issues/1216

* * *

### What problem does this feature solve?

We currently have a helper file in our codebase we pull in to testing files, it contains this beauty:

```js
/**
 * Create a component stub that will render all included slots from the
 * parent component. This lets you test the slots you've included in child component
 * without having to fully mount the child component.
 *
 * * Notes on implementation *
 * There is no one place this is clearly laid out. This thread gave the starting point:
 *
 * https://github.com/vuejs/vue-test-utils/issues/85
 *
 * but modifying the function requires understanding Vue's functional components
 * and the `render` function:
 *
 * https://vuejs.org/v2/guide/render-function.html
 *
 * especially the arguments that `createElement` can take:
 *
 * https://vuejs.org/v2/guide/render-function.html#createElement-Arguments
 *
 * In short, this produces a <div> with only the child components slots, but none of
 * its other functionality.
 *
 * @return {object} The functional component definition
 */
componentStubWithSlots: function () {
  return {
    render: function (createElement) {
      return createElement('div', [Object.values(this.$slots)]);
    }
  };
}
```

It would be nice if there was a more elegant way of handling this, or at least better documented (not requiring links to 3 different places).

### What does the proposed API look like?
That helper function gets used like so:

```js
test('Show only mine checkbox adds showOnlyMine to the query params', async () => {
  const wrapper = shallow(reservationsList, {
    store,
    localVue,
    stubs: {
      RouterLink: RouterLinkStub,
      'base-table': helpers.componentStubWithSlots()
    },
    mocks: { $ga }
  });

  let checkbox = wrapper.find('[data-test="showOnlyMineCheckbox"]');
  checkbox.trigger('click');

  expect(wrapper.vm.paramsForGettingListItems)
    .toEqual({ showOnlyMine: true });
});
```

I think it would be nicer if there was something like that built in, so we could just do this:

```js
test('Show only mine checkbox adds showOnlyMine to the query params', async () => {
  const wrapper = shallow(reservationsList, {
    store,
    localVue,
    stubs: {
      RouterLink: RouterLinkStub
    },
    stubComponentSlots: [
      'base-table'
    ],
    mocks: { $ga }
  });

  let checkbox = wrapper.find('[data-test="showOnlyMineCheckbox"]');
  checkbox.trigger('click');

[...]

PR #102, opened 2020-04-30
POC - Render all stub slots via config 

dobromir-hristov:
This is what I got... and I am not sure I like it...

1. Scoped slots just cannot work with auto stubs and `shallowMount`.
1. We cannot know what slots a component has defined, we can just render everything.

Check it out, play with it if you want. I wanted Evan to give me an advice on this, but yeah...

Related to #69

### DIFF
--- a/tests/mountingOptions/stubs.global.spec.ts
+++ b/tests/mountingOptions/stubs.global.spec.ts
@@ -1,6 +1,6 @@
 import { h, ComponentOptions } from 'vue'
 
-import { mount } from '../../src'
+import { config, mount } from '../../src'
 import Hello from '../components/Hello.vue'
 import ComponentWithoutName from '../components/ComponentWithoutName.vue'
 import ComponentWithSlots from '../components/ComponentWithSlots.vue'
@@ -302,23 +302,46 @@ describe('mounting options: stubs', () => {
     expect(wrapper.html()).toBe('<foo-bar-stub></foo-bar-stub>')
   })
 
-  it('renders stub slots by default', () => {
+  describe('stub slots', () => {
     const Component = {
       name: 'Parent',
-      template: `<div><component-with-slots>
-        <template #default>Default</template>
-        <template #named>A named</template>
-        <template #noop>A not existing one</template>
-      </component-with-slots></div>`,
+      template: `
+        <div>
+          <component-with-slots>
+            <template #default>Default</template>
+            <template #named>A named</template>
+            <template #noop>A not existing one</template>
+            <template #scoped="{ boolean, string }">{{ boolean }} {{ string }}</template>
+          </component-with-slots>
+        </div>`,
       components: { ComponentWithSlots }
     }
-    const wrapper = mount(Component, {
-      global: {
-        stubs: ['ComponentWithSlots']
-      }
+
+    afterEach(() => {
+      config.renderAllStubSlots = false
     })
-    expect(wrapper.html()).toBe(
-      '<div><component-with-slots-stub>Default A named With Default Content</component-with-slots-stub></div>'
-    )
+
+    it('renders only the default stub slot by default', () => {
+      const wrapper = mount(Component, {
+        global: {
+          stubs: ['ComponentWithSlots']
+        }
+      })
+      expect(wrapper.html()).toBe(
+        `<div><component-with-slots-stub>Default</component-with-slots-stub></div>`
+      )
+    })
+
+    it('renders all provided stub slots', () => {
+      config.renderAllStubSlots = true
+      const wrapper = mount(Component, {
+        global: {
+          stubs: ['ComponentWithSlots']
+        }
+      })
+      expect(wrapper.html()).toBe(
+        '<div><component-with-slots-stub>DefaultA namedA not existing one </component-with-slots-stub></div>'
+      )
+    })
   })
 })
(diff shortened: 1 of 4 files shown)


tests/mountingOptions/stubs.global.spec.ts
```diff
@@ -300,4 +301,47 @@ describe('mounting options: stubs', () => {
...
+    })
+
+    it('renders only the default stub slot by default', () => {
+      const wrapper = mount(Component, {
+        global: {
+          stubs: ['ComponentWithSlots']
+        }
+      })
+      expect(wrapper.html()).toBe(
+        `<div><component-with-slots-stub>Default</component-with-slots-stub></div>`
```
lmiller1990, line 331: This seems in correct - passing a stub when the `renderAllStubSlots` is `false` (whcih is it in this test) should just render a stub with no children, shouldn't it? (not sure how beta works here, but if this is how it worked, it's a bug in beta).

tests/mountingOptions/stubs.global.spec.ts
```diff
@@ -300,4 +301,47 @@ describe('mounting options: stubs', () => {
...
+    })
+
+    it('renders all provided stub slots', () => {
+      config.renderAllStubSlots = true
+      const wrapper = mount(Component, {
+        global: {
+          stubs: ['ComponentWithSlots']
+        }
+      })
+      expect(wrapper.html()).toBe(
```
lmiller1990, line 342: what happens if we have

```html
<template>
   <Child>
      Child
  </Child>
</template>
```

or

```html
<template>
   <Child>
      <Another Child>Child</Another Child>
  </Child>
</template>
```

? 

What "should" happen?

An alternative would be to limit this behavior to `shallowMount` - that seems how people usually use it (once we implement shallowMount which should be quite easy with this).

Will we just get `<child-stub />`?

tests/mountingOptions/stubs.global.spec.ts
dobromir-hristov: I thought that it made sense to render the default slot always. If you have a button or something of the like, most often the default slot is what you need.

tests/mountingOptions/stubs.global.spec.ts
dobromir-hristov: I added a test for that case.

tests/mountingOptions/stubs.global.spec.ts
lmiller1990: I think this behavior be part of the `renderAllStubSlots` flag - it seems much cleaner to me to just do

- default behavior: a stub is a stub. It just stubs everything out. easy to understand
- `renderAllStubSlots`: renders the slots (default and named). also easy to understand.

We could also make the flag `renderStubSlots` (this way, `all` really does mean all slots, not just named slots).

tests/mountingOptions/stubs.global.spec.ts
lmiller1990: Cool! Is this in line with the behavior people want/requested in beta (I have no idea what is ideal/desirable; I don't use this feature).

This thread comes to mind; the last user to post has an extremely strong opinion about this: https://github.com/vuejs/vue-test-utils/issues/1307#issuecomment-617491495

dobromir-hristov:
I think we should make an RFC explaining why we are doing this and what solution we offer.
This is a feature that mostly worked before and breaking it will cause every `shallowMount` based testsuite to break. And they are not as few as we expected.

I can write up an RFC, explaining what we tried, what the limitations are and what alternatives people can use, along with what we are offering as "official" solution for the problem.

This thing should imho be in the guides later on, so people are fully aware that this way of testing is not reliable, and will lead to problems if you rely 100% on it.

lmiller1990:
Hm - we can do an RFC, sure. I feel like most of the people who have a strong opinion about rendering slots of stubbed components hold that opinion because that's how beta worked, and it will cause extra work for them to update their code-bases, not because it actually makes sense.

Could we have a flag that attempts to simulate the old behavior to let people upgrade - kind of like what you did here where a component passed to `stubs` will still render the default slot. Something like `config.vtu-beta-compat`, for example. 

I understand a lot of code-bases will break, and we do need to consider those people, and this flag (might) make it easier to upgrade. I don't think re-implementing this behavior purely because that's how it worked before is the right answer.

dobromir-hristov:
Yeah, we can't really reimplement it fully anyway. It will be behind a flag, I just want to make sure people get it.

lmiller1990:
I see two things to consider:

- new users and the path forward. I obviously like the behavior where we just stub the component and render no children. It's simple, it works.
- migrating users. We can't just screw them over and change things hugely. We should "reimplement" VTU beta as best we can. sStick it behind a flag and give people 18 months to migrate.

Best of both worlds (imo). I do not work on any HUGE Vue apps (largest is around 4k lines, not tht big, prob 1k unit tests) so I don't have the best perspective on how hard migrating will be, though. We also don't `stub` that often so I am quite biased.

lmiller1990:
This looks pretty good. Are we going to bundle a helper too?

Once that is done, I think we should take this chance to update the docs and explain how this works, and provide some examples on testing slots and stubs.

I can look into shallowMount soon, it should be pretty easy with this PR merged in.

dobromir-hristov:
Yeah ofc.

lmiller1990:
@dobromir-hristov maybe we can add this info around [here](https://github.com/vuejs/vue-test-utils-next-docs/tree/master/src/api#slots). We should also add a section about `shallow: true` in the mounting options. Finally, we did not talk about global config yet.

This can be done at any time, no rush.

Issue #104, opened 2020-05-02
Test with `async import`


lmiller1990:
This was buggy in beta, let's make sure it works here. eg; import a component async and use it.

lmiller1990:
Surprisingly it looks like this "just works".

Nothing ever "just works". I will add some tests.

PR #105, opened 2020-05-02
Add tests using async components

lmiller1990:
#104

They work as expected! Caveats:

- since they are async, you need to do `flushPromises` to make them resolve. This is pretty common in VTU. Just need good education around this. Evan does something similar in core to test them, [see here](https://github.com/vuejs/vue-next/blob/b570287dfe391d90ccf3f1ba43c95888f739a95d/packages/runtime-core/__tests__/apiAsyncComponent.spec.ts#L34)
- It errors out with `cannot find innerHTML of parentElement` if an AsyncComponent is the root component. `return this.vm.$el.parentElement` is null - I am not sure why. 🤔 
  - Work around is to wrap with a `div`. See my "basic usage" test.

Anyone see any other test cases we could add?

### DIFF
--- a/tests/components/LoadingComponent.vue
+++ b/tests/components/LoadingComponent.vue
@@ -1,0 +1,3 @@
+<template>
+  <div>Loading Component</div> 
+</template>
--- a/tests/components/LoadingComponent.vue
+++ b/tests/components/LoadingComponent.vue
@@ -1,3 +1,0 @@
-<template>
-  <div>Loading Component</div> 
-</template>
--- a/tests/features/async-components.spec.ts
+++ b/tests/features/async-components.spec.ts
@@ -1,0 +1,122 @@
+import {
+  defineAsyncComponent,
+  defineComponent,
+  h,
+  ComponentPublicInstance
+} from 'vue'
+import flushPromises from 'flush-promises'
+
+import { mount } from '../../src'
+
+// AsyncComponents are documented here: https://github.com/vuejs/rfcs/blob/async-component/active-rfcs/0026-async-component-api.md
+
+describe('defineAsyncComponent', () => {
+  it('works with the basic usage', async () => {
+    const AsyncHello = defineAsyncComponent(() =>
+      import('../components/Hello.vue')
+    )
+    const Comp = defineComponent({
+      render() {
+        return h('div', [h(AsyncHello)])
+      }
+    })
+
+    const wrapper = mount(Comp)
+    await flushPromises()
+    expect(wrapper.html()).toContain('Hello world')
+  })
+
+  it('works with options usage', async (done) => {
+    const Async = defineAsyncComponent({
+      loader: () =>
+        new Promise<any>((res) => {
+          setTimeout(() => {
+            res({
+              template: '<div>Async Component</div>'
+            })
+          }, 75)
+        }),
+      loadingComponent: {
+        template: '<div>Loading Component</div>'
+      },
+      delay: 10
+    })
+
+    const Comp = defineComponent({
+      render() {
+        return h('div', [h(Async)])
+      }
+    })
+
+    const wrapper = mount(Comp)
+
+    setTimeout(() => {
+      expect(wrapper.html()).toContain('Loading Component')
+    }, 35)
+
+    setTimeout(() => {
+      expect(wrapper.html()).toContain('Async Component')
+      done()
+    }, 100)
+  })
+
+  it('works with vue files', async () => {
+    const Async = defineAsyncComponent({
+      loader: () => import('../components/Hello.vue')
+    })
+
+    const Comp = defineComponent({
+      render() {
+        return h('div', [h(Async)])
+      }
+    })
+
+    const wrapper = mount(Comp)
+    await flushPromises()
+
+    expect(wrapper.html()).toContain('Hello world')
+  })
+
+  it('catches error and renders ErrorComponent', async () => {
+    const Async = defineAsyncComponent({
+      loader: () =>
+        new Promise<any>((res, rej) => {
+          rej('Async component failed to load.')
+        }),
+      errorComponent: {
+        template: '<div>Error Component</div>'
+      },
+      onError(error, retry, fail, attempts) {
+        fail()
+      }
+    })
+
+    const Comp = defineComponent({
+      render() {
+        return h('div', [h(Async)])
+      }
+    })
+
+    const wrapper = mount(Comp)
+    await flushPromises()
+
+    expect(wrapper.html()).toContain('Error Component')
+  })
+
+  // TODO: Find out why this does not work
+  // Is it valid to have an AsyncComponent as the root? Was it ever?
+  it.skip('works when AsyncComponent is the root', async () => {
+    const AsyncHello = defineAsyncComponent(() =>
+      import('../components/Hello.vue')
+    )
+    const Comp = defineComponent({
+      render() {
+        return h(AsyncHello)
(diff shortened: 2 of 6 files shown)


tests/features/async-components.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+      loader: () =>
+        new Promise<any>((res, rej) => {
+          rej('Async component failed to load.')
+        }),
+      errorComponent: {
+        template: '<div>Error Component</div>'
+      },
+      onError(error, retry, fail, attempts) {
+        fail()
+      }
```
afontcu, line 91: I think this `onError` is a bit redundant? The option is used to have some retry control, but here we're just failing no matter what.

afontcu: Hm, also the test is passing but it yields some warns and errors:

image

afontcu: Not sure if these error logs are somewhat expected, and we should just mock console for the test and call it a day… 🤔

lmiller1990: I had this thought too. Some other tests also throw and error. Is there any benefit to mocking the console? If printing the error is correct, I don't keep it.

afontcu: I tend to mute them to make it the output cleaner in the long run, but I don't care much either. Unless the console error is part of the expected output, then I spy it and assert the times called and so.

afontcu: (I still believe the `onError` bit can be removed)

tests/features/async-components.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+    const wrapper = mount(Comp)
+
+    setTimeout(() => {
+      expect(wrapper.html()).toContain('Loading Component')
+    }, 35)
+
+    setTimeout(() => {
+      expect(wrapper.html()).toContain('Async Component')
+      done()
+    }, 100)
```
SergiCL, line 60: Why not use [jest.useFakeTimers](https://jestjs.io/docs/en/timer-mocks)? Mocking the `setTimeout` would make the test 10 times faster.
Something like this seems to work fine, but there may be something I'm not seeing:

```suggestion
    await jest.advanceTimersByTime(35)
    expect(wrapper.html()).toContain('Loading Component')

    jest.advanceTimersByTime(100)
    await flushPromises()
    expect(wrapper.html()).toContain('Async Component')
```

tests/components/LoadingComponent.vue
```diff
@@ -0,0 +1,3 @@
+<template>
```
dobromir-hristov, line 1: Can we just use any sfc?

tests/features/async-components.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+  })
+
+  it('works with vue files', async () => {
```
dobromir-hristov, line 63: isnt this test the same as the first one?

tests/features/async-components.spec.ts
dobromir-hristov: We can have our own Vue `errorHandler`, to easily assert on it without mocking `console.error`

lmiller1990: I will make this change, thanks

tests/features/async-components.spec.ts
lmiller1990: I did not know jest had a fake timer, that is pretty cool.

tests/components/LoadingComponent.vue
lmiller1990: Sure, I can use one of the existing ones

tests/features/async-components.spec.ts
lmiller1990: Oops, it is.

I can change this one to test this usage

```
defineAsyncComponent({
  errorComponent: Hello
})
```

where `errorComponent` is a regular `import` at the top of the file
```

lmiller1990:
Great reviews everyone, I will make the requested changes and update the PR.

cexbrayat:
@lmiller1990 👍 I'm also in favor of using `jest.useFakeTimers` (I already thought so when seeing the Suspense test). Also note that `flushPromises` is not necessary, as you could write a bunch a few `nextTick` and achieve the same. It just far more convenient. I have a few tests in my app that use it, and was also wondering if we should not expose it ourselves 🤔 

You could also add a test of a component with an `async` setup, meant to be used in a `Suspense`? They are a bit cumbersome to test currently, as you need a wrapper component defining the `Suspense` but it works.

lmiller1990:
For `async setup` components, I have not found a good way to test these - in my apps I either test the component "one level up" (with `<Suspense>` or wrap the component in a `<Suspense>` in the test. I then call `await flushPromises` to make it render. 

Does anyone have any good ideas for this?

I had the idea for a `mountInSuspense` function - but it's just as easy for a user to implement that themselves. I see two options:

- have some simple FAQs and show the user how to write a helper
- provide a helper

I am leaning towards the first, if something is simple enough the user can do it in user-and, just let them do it. People should absolutely write helpers if it makes their tests more concise. That said, if this going to come up a lot, we can per-emptilvely solve it with a helper. My only fear is we end up many helper functions, which makes the library larger and more complex than it needs to be.

cexbrayat:
@lmiller1990 I tested a few async components in my app and ended up with the same conclusions :+1: I wanted to open an issue to talk about this, and gather your thoughts.

I think it would be nice to offer a helper that would allow to mount an async component easily, either a different mount, or with a mount option, or just a helper to wrap the component to test. I completely understand that this would make the library bigger and that we might not want to do it, and simply write a good cookbook on the topic.

lmiller1990:
Alternatively, if we can know ahead of time it's an `async setup` function... can we just do the work for them in the background so it "just works"?

afontcu:
> I am leaning towards the first, if something is simple enough the user can do it in user-and, just let them do it. People should absolutely write helpers if it makes their tests more concise. Give the power to the user!
> That said, if this going to come up a lot, we can preemptively solve it with a helper.

Agree 100%

lmiller1990:
I made most of the changes except `flushPromises` to `nextTick`. These are different - `nextTick` is DOM updates only, `flushPromises` does what the name suggested. Eg, in the first test we need `flushPromises` to the `defineAsyncComponent` promise is resolved. Or maybe I am missing something.

With `trigger`, `setProps` etc now returning `nextTick`, there should be no scenarios where you next to do `await nextTick` yourself, right? If you need a promise (eg async import, axios whatever) you need to use `flushPromises`.

lmiller1990:
@afontcu can you confirm which thing you are 100% agreeing to (util function or user-land)?

pikax:
> When I changed it to `nextTick` it would fail sometimes.

`nextTick` does a `microTask` but `flushPromises` is a `macroTask`

lmiller1990:
^ This was my understanding - thanks for clarifying. We will stick with `flush-promises` here for this reason.

afontcu:
> @afontcu can you confirm which thing you are 100% agreeing to (util function or user-land)?

both! I'd start with docs and let user-land figure out a way if needed, and if this comes up a lot during alpha/beta stages, then we could reconsider.

lmiller1990:
Ok - sounds good to me. Let's continue this in another issue: https://github.com/vuejs/vue-test-utils-next/issues/108

For now, unless anyone has any other comments here, I will merge this (will give 24h more for any other ideas/suggestions).

PR #107, opened 2020-05-04
feat: shallow mount

### LINKED ISSUE #6: epic: `shallow: true` mounting option (previously `shallowMount`)

dobromir-hristov:
Adds a basic implementation for `shallowMount` via `MountOptions.shallow`

I still need to add tests, but it seems to work great.

resolves #6

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,7 +1,7 @@
-import { mount } from './mount'
+import { mount, shallowMount } from './mount'
 import { RouterLinkStub } from './components/RouterLinkStub'
 import { VueWrapper } from './vue-wrapper'
 import { DOMWrapper } from './dom-wrapper'
 import { config } from './config'
 
-export { mount, RouterLinkStub, VueWrapper, DOMWrapper, config }
+export { mount, shallowMount, RouterLinkStub, VueWrapper, DOMWrapper, config }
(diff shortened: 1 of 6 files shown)


lmiller1990:
I think we move to beta once this is merged and docs are up to date.

dobromir-hristov:
I dont think so.. I dont know if there is a point to add more tests. It is just mount with stubs :P

PR #110, opened 2020-05-05
feat: improve mount typings when using object syntax and infer Data type

pikax:
This improves typing when using 

```ts
mount({
 props: {
   // ...
  }
})
```

and also adds the data type validation
```ts
const AppWithDefine = defineComponent({
  props: {
    a: {
      type: String,
      required: true
    },
    b: Number
  },
  template: ''
})



// no data provided
expectError(
  mount(AppWithDefine, {
    data() {
      return {
        myVal: 1
      }
    }
  })
)
```

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -34,8 +34,8 @@ import { stubComponents } from './stubs'
 
 type Slot = VNode | string | { render: Function }
 
-interface MountingOptions<Props, Data = {}> {
-  data?: () => Data
+interface MountingOptions<Props, Data extends object = {}> {
+  data?: () => Data extends unknown ? never : Data
   props?: Props
   attrs?: Record<string, unknown>
   slots?: {
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -35,7 +35,7 @@ import { stubComponents } from './stubs'
 type Slot = VNode | string | { render: Function }
 
 interface MountingOptions<Props, Data = {}> {
-  data?: () => unknown extends Data ? never : Partial<Data>
+  data?: () => Data extends object ? Partial<Data> : never
   props?: Props
   attrs?: Record<string, unknown>
   slots?: {
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -35,7 +35,7 @@ import { stubComponents } from './stubs'
 type Slot = VNode | string | { render: Function }
 
 interface MountingOptions<Props, Data = {}> {
-  data?: () => Data extends unknown ? never : Data
+  data?: () => unknown extends Data ? never : Partial<Data>
   props?: Props
   attrs?: Record<string, unknown>
   slots?: {
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -34,7 +34,7 @@ import { stubComponents } from './stubs'
 
 type Slot = VNode | string | { render: Function }
 
-interface MountingOptions<Props, Data extends object = {}> {
+interface MountingOptions<Props, Data = {}> {
   data?: () => Data extends unknown ? never : Data
   props?: Props
   attrs?: Record<string, unknown>
@@ -102,7 +102,7 @@ export function mount<
     E,
     EE
   >,
-  options?: MountingOptions<Props, D>
+  options?: MountingOptions<never, D>
 ): VueWrapper<
   ComponentPublicInstance<Props, RawBindings, D, C, M, E, VNodeProps & Props>
 >
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -11,7 +11,12 @@ import {
   ComponentOptionsWithArrayProps,
   ComponentOptionsWithoutProps,
   ExtractPropTypes,
-  Component
+  Component,
+  VNodeProps,
+  WritableComputedOptions,
+  SetupContext,
+  RenderFunction,
+  ComponentPropsOptions
 } from 'vue'
 
 import { config } from './config'
@@ -29,8 +34,8 @@ import { stubComponents } from './stubs'
 
 type Slot = VNode | string | { render: Function }
 
-interface MountingOptions<Props> {
-  data?: () => Record<string, unknown>
+interface MountingOptions<Props, Data = {}> {
+  data?: () => Data
   props?: Props
   attrs?: Record<string, unknown>
   slots?: {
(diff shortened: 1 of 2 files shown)


src/mount.ts
```diff
@@ -29,8 +34,8 @@ import { stubComponents } from './stubs'
 
 type Slot = VNode | string | { render: Function }
 
-interface MountingOptions<Props> {
-  data?: () => Record<string, unknown>
+interface MountingOptions<Props, Data = {}> {
+  data?: () => unknown extends Data ? never : Partial<Data>
```
lmiller1990, line 38: How can `unknown` extend `Data` 🤔 ? What is happening here? Doesn't this means we know a bit more about data (eg, it's an `object` for example?)

src/mount.ts
```diff
@@ -42,35 +47,127 @@ interface MountingOptions<Props> {
...
+  string,
+  ((...args: any[]) => any) | null
+>
+export type EmitsOptions = ObjectEmitsOptions | string[]
 
 // Component declared with defineComponent
 export function mount<TestedComponent extends ComponentPublicInstance>(
   originalComponent: { new (): TestedComponent } & Component,
-  options?: MountingOptions<TestedComponent['$props']>
+  options?: MountingOptions<TestedComponent['$props'], TestedComponent['$data']>
```
lmiller1990, line 63: This is great!

src/mount.ts
```diff
@@ -42,35 +47,127 @@ interface MountingOptions<Props> {
...
+    RawBindings,
+    D,
+    C,
+    M,
+    E,
+    VNodeProps & ExtractPropTypes<PropsOptions, false>
+  >
+>
+
+// implementation
```
lmiller1990, line 170: I love it when the typings are longer than the implementation 😆

src/mount.ts
pikax: extending object is better 😄   Thank you 👍

lmiller1990:
Maybe @cexbrayat can take a quick look 👀

src/mount.ts
```diff
@@ -42,35 +47,127 @@ interface MountingOptions<Props> {
...
-): VueWrapper<ExtractComponent<TestedComponent>>
+
+// Functional Component declared with (props, ctx)=> render
+export function mount<Props, RawBindings = object>(
+  setup: (
+    props: Readonly<Props>,
+    ctx: SetupContext
+  ) => RawBindings | RenderFunction,
+
+  options: MountingOptions<Props, {}>
```
cexbrayat, line 73: nit: as `{}` is the default value for the `Data` type, I think you can simplify to `MountingOptions<Props>`?

src/mount.ts
```diff
@@ -42,35 +47,127 @@ interface MountingOptions<Props> {
...
+): VueWrapper<
+  ComponentPublicInstance<
+    Props,
+    RawBindings,
+    {},
+    {},
+    {},
+    // public props
+    VNodeProps & Props
+  >
```
cexbrayat, line 83: :+1: for handling functional component

OOC: do we need to declare all the generics here? If we can achieve what we want with just `ComponentPublicInstance<Props>`? If so, I vote to keep it simple.
If we can't, I would love  to:
- have a comment to explain why
- have a tsd test that fails without it

src/mount.ts
```diff
@@ -42,35 +47,127 @@ interface MountingOptions<Props> {
...
 export function mount<
-  TestedComponent extends ComponentOptionsWithoutProps,
-  ComponentT extends ComponentOptionsWithoutProps & {}
+  Props = {},
+  RawBindings = {},
+  D = {},
+  C extends ComputedOptions = {},
+  M extends Record<string, Function> = {},
+  E extends EmitsOptions = Record<string, any>,
+  EE extends string = string
```
cexbrayat, line 94: Same here: if we really need all the generics, it would be great to have a test showcasing why.
Also, maybe pick more explicit names for the generic types and add comments: I struggle to follow what they refer to

test-dts/mount.d-test.ts
```diff
@@ -13,12 +13,23 @@ const AppWithDefine = defineComponent({
...
-  props: { a: 'Hello', b: 2 }
-})
-// vm is properly typed
-expectType<string>(wrapper.vm.a)
+// accept props- vm is properly typed
+expectType<string>(
+  mount(AppWithDefine, {
+    props: { a: 'Hello', b: 2 }
+  }).vm.a
+)
```
cexbrayat, line 21: nit: I think the test is more readable when splitted like it was. Same for the following changes.

test-dts/mount.d-test.ts
```diff
@@ -71,28 +82,43 @@ const AppWithArrayProps = {
...
   props: { a: 'Hello', b: 2 }
 })
 
+// cannot receive extra props
+// if they pass use object inside
+expectError(
+  mount(
+    {
+      props: ['a']
+    },
```
cexbrayat, line 104: I'm happy to see that it throws an error, but I don't get how this is different from the test case with  `AppWithArrayProps` which still succeeds? Maybe add more explanation in the comment.
Ideally, shouldn't `mount(AppWithArrayProps, {  props: { a: 'Hello', b: 2 } })` also fail?

test-dts/mount.d-test.ts
```diff
@@ -13,12 +13,23 @@ const AppWithDefine = defineComponent({
...
+// no data provided
+expectError(
+  mount(AppWithDefine, {
+    data() {
+      return {
+        myVal: 1
+      }
+    }
+  })
+)
```
cexbrayat, line 32: 👍 I think more tests with data would be awesome. Can we pass `data` to all type of components? Can we pass data that are not declared? etc. That will greatly help us maintining the lib if we have extensive coverage of the different cases.

src/mount.ts
pikax: this is basically a `copy&paste` from [defineComponent.ts](https://github.com/vuejs/vue-next/blob/master/packages/runtime-core/src/apiDefineComponent.ts#L21-L126)

To make it simpler to maintain (bring the types from that file), I can add a comment explaining.

test-dts/mount.d-test.ts
pikax: Basically when you use `mount({ props: ['a']}) is basically the same as doing `defineComponent({props:['a']})`, the ThisType will be inferred as the `$vm`, resulting in:
```ts

mount({
  props: ['a']
})
// type
type MountTyped = {
  props: ['a'] 
} 


const Comp = {
  props: ['a']
}
//type
type CompTyped = {
  props: string[]
}
```

Basically using `mount/defineComponent` we maintain the actual tuple value, when using a variable with the options, typescript will resolve the tuple as `string[]`

cexbrayat: Got it! So the previous test should also throw if the props were declared as const right? 
```ts
const AppWithArrayProps = {
  props: ['a'] as const,
  template: ''
}
```

lmiller1990:
Is this one awaiting more tests? If so, I can try and help out. LMK where to start - just go over the comments and add those tests?

cexbrayat:
@lmiller1990 Yes! I don't know if @pikax had time to start working on them... If not and if you are willing to add some, I'll gladly review them 👍

lmiller1990:
Have a few other things to work on atm (mainly docs, and vue-jest).

Rather than let this PR hang around forever, I'll rebase and get it merged. We can add the tests at some point.

src/mount.ts
pikax: the functional overload can be removed in favour of the 

```ts
// Functional component
export function mount<
  TestedComponent extends FunctionalComponent<Props>,
  Props
>(
  originalComponent: TestedComponent,
  options?: MountingOptions<Props>
): VueWrapper<ComponentPublicInstance<Props>>
```

test-dts/mount.d-test.ts
pikax: That doesn't seem to happen, if you define `as const`, it will use `props: {}` overload 🤔

lmiller1990:
Should we merge or do you guys want to add more tests before that?

lmiller1990:
I thought about this a bit more. As cool as this feature is, until we have some way to actually get the types from a `.vue` file, very few people will actually be able to really leverage it. I wonder how difficult this is going to be to get working with `vue` files 🤔

cexbrayat:
I'm OK with merging it. I think it's still better to have it than not, even if I agree with you @lmiller1990

lmiller1990:
I think we should definitely have them towards the end of beta and moving into a release candidate - for now I'm happy to merge this up and revisit when we have better IDE integration.

I am really quite excited about this - I know little to nothing about language servers, might be time to learn.

PR #115, opened 2020-05-09
Feat: Implement Scoped Slot string support

dobromir-hristov:
This PR implements the previously very convenient scoped slot syntax of providing just a template string.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -25,6 +25,7 @@ function createEntry(options) {
     input,
     external: [
       'vue',
+      '@vue/compiler-dom',
       'lodash/mergeWith',
       'lodash/isString'
     ],
@@ -41,6 +42,7 @@ function createEntry(options) {
       format,
       globals: {
         vue: 'Vue',
+        '@vue/compiler-dom': 'VueCompilerDOM',
         'lodash/mergeWith': '_.mergeWith',
         'lodash/isString': '_.isString',
       }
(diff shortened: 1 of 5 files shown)


lmiller1990:
I played with this, seems good. I wonder if we need any tests with `shallowMount` and the `renderDefaultSlot` flag?

PS: I fixed the build. We had to specify `@vue/compiler-dom` as an external dependency the rollup config. This means if you want to use VTU in a browser as an IIFE, you will need to include VueCompilerDOM globally. I think this makes sense.

dobromir-hristov:
OK yeah awesome. As I said I did not have time to spend on it, but glad you figured it out :))) 

I think we should make note in the docs that it needs to be present. Jessica's Vite toy project would need it.

Overall I am very very happy with how this turned out. It reduced the complexity by N times, everything is handled by the Vue compiler. Sure we have a hacky string to function thing, but thats fine 😆 

I can add more tests if anyone thinks of something? Maybe some not so happy paths?

lmiller1990:
I have updated the rollup config further here: https://github.com/vuejs/vue-test-utils-next/pulls

We have it listed as a peer dependency, I think this should be okay. For Vite, it will work out of the box, since all the packages in vue-next have the `module` field.

Sure, feel free to add more tests - would be happy to merge as is, as well. I'll leave this decision up to you since you own this feature 👍

znck:
We can't avoid string to function conversion, even full build of Vue uses the same thing:

https://github.com/vuejs/vue-next/blob/b725b63e5fd6b9c9875002c2c23e210293bd771b/packages/vue/src/index.ts#L66-L68

PR #116, opened 2020-05-10
fix: bundle dependencies inline and fix es build

lmiller1990:
As pointed about by @JessicaSachs, our es build was depending on cjs dependencies.

Although I like the idea of only shipping our code and asking the users to install dependencies like lodash, it does add a bit of complexity. Eg, depending if you are using VTU in a browser/node/es build, you need to include dependencies differently.

In this PR, we now inline some dependencies:

- `lodash/isString`, `lodash/mergeWith`
- `eventTypes` (for `trigger`)

Vue and compiler-sfc and compiler-dom are still peer dependencies, so the user will bring their own version of those.

This does increase the bundle size - from 900 lines to around 4000. That said, mostly you will not be shipping this library to users, and 4000 lines is before minify, so it could be smaller, if the user wants to make it smaller, so I think this trade-off is probably fine. Simple is best!

For comparison, VTU v1 is around 10k lines (also not minified).

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -25,8 +25,7 @@ function createEntry(options) {
     input,
     external: [
       'vue',
-      'lodash/mergeWith',
-      'lodash/isString'
+      '@vue/compiler-dom'
     ],
     plugins: [
       replace({
@@ -41,14 +40,11 @@ function createEntry(options) {
       format,
       globals: {
         vue: 'Vue',
+        '@vue/compiler-dom': 'VueCompilerDOM',
         'lodash/mergeWith': '_.mergeWith',
         'lodash/isString': '_.isString',
       }
     }
-  }
-
-  if (['es', 'cjs'].includes(format)) {
-    config.external.push('dom-event-types')
   }
 
   if (format === 'es') {
@@ -57,7 +53,7 @@ function createEntry(options) {
   if (format === 'cjs') {
     config.output.file = pkg.main
   }
-  console.log('file is', config.output.file)
+  console.log(`Building ${format}: ${config.output.file}`)
 
   config.plugins.push(
     ts({
@@ -78,7 +74,6 @@ function createEntry(options) {
 
 export default [
   createEntry({ format: 'es', input: 'src/index.ts', isBrowser: false }),
-  createEntry({ format: 'es', input: 'src/index.ts', isBrowser: true }),
   createEntry({ format: 'iife', input: 'src/index.ts', isBrowser: true }),
   createEntry({ format: 'cjs', input: 'src/index.ts', isBrowser: false }),
 ]
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -46,7 +46,7 @@ function createEntry(options) {
   }
 
   if (format === 'es') {
-    config.output.file = isBrowser ? pkg.browser : pkg.module
+    config.output.file = pkg.module
   }
   if (format === 'cjs') {
     config.output.file = pkg.main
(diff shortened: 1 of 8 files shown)


rollup.config.js
```diff
@@ -25,8 +25,7 @@ function createEntry(options) {
     input,
     external: [
       'vue',
-      'lodash/mergeWith',
-      'lodash/isString'
+      '@vue/compiler-dom'
```
lmiller1990, line 28: Not needed in master right now, but will be very soon for [this feature](https://github.com/vuejs/vue-test-utils-next/pull/115) when merged, so it's fine to include it here.

JessicaSachs:
Can we inline a cheap implementation of `isString` and `mergeWith`?

`isString` should be ez to write natively...

For merge with:
https://github.com/vuejs/vue-test-utils-next/blob/1f2a18f8ae11fc53bd131f4b3c0276f22a639cf7/src/utils.ts#L6

Even if we can't get rid of using a deep merge... can we at least write a one-liner reducer or inline a lighter weight deepMerge library that doesn't require ~20 helper files?

lmiller1990:
Yep @JessicaSachs was just thinking the same thing. I tried this with vite and it works great.

I will look into some more simple `mergeWith` functions.

lmiller1990:
`isString` was easy. `mergeDeep` also easy - but I found a bug in the process with our config.components. It seems the component will always use the locally registered component. Our original test was not actually testing what it should be.

I will investigate. Good thing we uncovered this bug.

lmiller1990:
Alright @JessicaSachs should be good now!

Also I tried it with Vite, it works great.

tests/config.spec.ts
```diff
@@ -15,16 +16,27 @@ describe('config', () => {
...
       props: ['msg']
     }
 
     it('allows setting components globally', () => {
-      config.global.components = { Hello }
-      const wrapper1 = mount(Component, { props: { msg: 'Wrapper1' } })
-      const wrapper2 = mount(Component, { props: { msg: 'Wrapper2' } })
-      expect(wrapper1.text()).toEqual('Wrapper1 Hello world')
-      expect(wrapper2.text()).toEqual('Wrapper2 Hello world')
+      const HelloLocal = {
```
dobromir-hristov, line 25: Should we add a test if the locally registered component  is overwritten properly?

lmiller1990:
@dobromir-hristov I guess - I just ran the tests in watch mode until they passed. Do we have some tests that use arrays? The mixin tests are passing, which I believe use arrays.

tests/config.spec.ts
lmiller1990: We have this in the other test (I think?)

I will check.

dobromir-hristov:
Yeah that plus should we deal with mixins somehow? Are mixins concatinated or overwritten?

dobromir-hristov:
Pushed a test for the mixins, which proved me right unfortunately. Arrays should be concatinated, not merged.
I dont like that the `merge` method is mutating like that the objects.

Overall we want a more specialized merge function.

`mixins` and `plugins` should be concatinated.
`config`, `mocks`, `provide`, `components`, `directives` and `stubs` should be overwritten by key, but not merged.

lmiller1990:
thanks for that @dobromir-hristov, been a bit busy these last few days. I'll will look at this now.

lmiller1990:
Ok @dobromir-hristov getting closer, this is much cleaner and closer to what we had before - just a few failing tests to fix up.

lmiller1990:
ok @dobromir-hristov we are green! Added all the tests.

dobromir-hristov:
I will check it out in the morning :)

lmiller1990:
@dobromir-hristov we good to merge? thanks for refactor/tests.

dobromir-hristov:
Think so

PR #118, opened 2020-05-13
fix: allow mounting functional components

dobromir-hristov:
This PR allows mounting and testing functional components.

I think we should add a test for a few more things, like setProps etc. 

Also `setData` makes no sense in a FunctionalComponent, for obv reasons.

`vm` is not the functional component, because there is no instance there.

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -6,6 +6,7 @@ import {
   VNodeNormalizedChildren,
   transformVNodeArgs,
   reactive,
+  FunctionalComponent,
   ComponentPublicInstance,
   ComponentOptionsWithObjectProps,
   ComponentOptionsWithArrayProps,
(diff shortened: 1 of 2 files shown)


tests/functionalComponents.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+describe('functionalComponents', () => {
+  it('mounts a functional component', () => {
+    const Foo = (props) => h('div', { class: 'foo' }, props.msg)
+
+    const wrapper = mount(Foo, {
+      props: {
+        // @ts-ignore
+        msg: 'foo'
+      }
+    })
```
pikax, line 14:
```suggestion
    const Foo = (props: { msg: string }) =>
      h('div', { class: 'foo' }, props.msg)

    const wrapper = mount(Foo, {
      props: {
        msg: 'foo'
      }
    })
```

you can described the props on the function parameter :)

src/mount.ts
```diff
@@ -53,6 +54,11 @@ type ExtractComponent<T> = T extends { new (): infer PublicInstance }
   ? PublicInstance
   : any
 
+// Functional component
+export function mount<TestedComponent extends FunctionalComponent>(
+  originalComponent: TestedComponent,
+  options?: MountingOptions<any>
+): VueWrapper<ComponentPublicInstance>
```
pikax, line 61: We will improve this on https://github.com/vuejs/vue-test-utils-next/pull/110

PR #122, opened 2020-05-16
chore: wrapper overloads signatures

cexbrayat:
This provides a better type checking than the optional parameter.

ping @JessicaSachs as we talked about that on Discord: if you want to take a quick look, my explanation will probably make more sense with the proper code :)

### DIFF
--- a/src/dom-wrapper.ts
+++ b/src/dom-wrapper.ts
@@ -24,7 +24,7 @@ export class DOMWrapper<ElementType extends Element> {
   attributes(key: string): string
   attributes(key?: string): { [key: string]: string } | string {
     const attributes = this.element.attributes
-    const attributeMap: { [key: string]: string } = {}
+    const attributeMap: Record<string, string> = {}
     for (let i = 0; i < attributes.length; i++) {
       const att = attributes.item(i)!
       attributeMap[att.localName] = att.value
--- a/src/dom-wrapper.ts
+++ b/src/dom-wrapper.ts
@@ -10,7 +10,9 @@ export class DOMWrapper<ElementType extends Element> {
     this.element = element
   }
 
-  classes(className?: string) {
+  classes(): string[]
+  classes(className: string): boolean
+  classes(className?: string): string[] | boolean {
     const classes = this.element.classList
 
     if (className) return classes.contains(className)
--- a/src/dom-wrapper.ts
+++ b/src/dom-wrapper.ts
@@ -18,11 +18,13 @@ export class DOMWrapper<ElementType extends Element> {
     return Array.from(classes)
   }
 
-  attributes(key?: string) {
+  attributes(): { [key: string]: string }
+  attributes(key: string): string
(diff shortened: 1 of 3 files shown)


src/dom-wrapper.ts
```diff
@@ -10,19 +10,23 @@ export class DOMWrapper<ElementType extends Element> {
...
     return Array.from(classes)
   }
 
-  attributes(key?: string) {
+  attributes(): { [key: string]: string }
+  attributes(key: string): string
+  attributes(key?: string): { [key: string]: string } | string {
     const attributes = this.element.attributes
-    const attributeMap = {}
+    const attributeMap: { [key: string]: string } = {}
```
lmiller1990, line 27: I tend to do `Record<string, string>`; is there any benefit to one or the other?

src/dom-wrapper.ts
```diff
@@ -10,19 +10,23 @@ export class DOMWrapper<ElementType extends Element> {
...
     for (let i = 0; i < attributes.length; i++) {
-      const att = attributes.item(i)
+      const att = attributes.item(i)!
```
lmiller1990, line 29: If the attribute is missing, won't `att.localName` give "could not find `localName` of undefined?

src/dom-wrapper.ts
pikax: `Record<string, string>` implementation is pretty much that:
```ts
type Record<K extends keyof any, T> = {
    [P in K]: T;
};
```

Using this syntax allows greater control on the value, in this case `Record<string, string>` is clearer in this case

src/dom-wrapper.ts
pikax: we are iterating the attribute list, so I don't think is possible this is `undefined` at runtime

src/dom-wrapper.ts
lmiller1990: I agree, can we make this change @cexbrayat ? thanks!

src/dom-wrapper.ts
lmiller1990: Right 🤦

pikax: I had to double check when I was reviewing, probably a forof would be a bit more explicit, but that's a bit nit picking :D

src/dom-wrapper.ts
cexbrayat: Sure, done!

cexbrayat:
@lmiller1990 Updated according to your comment, PTAL

src/dom-wrapper.ts
cexbrayat: Yeah I agree. I had to double check too when I added this `!`.

cexbrayat: It looks like we all agree, so I pushed an extra commit with a refactoring to use forof instead.

lmiller1990:
Thanks!

PR #123, opened 2020-05-18
chore: typings maintenance in mount

cexbrayat:
This commit tidies up the types in mount. The goal is to enable `strict: true` in the future on the project, and this commit fixes up a bunch of errors that the strict mode complains about.

The only change for VTU users is that `GlobalMountOptions.config` now does not allow to override `isNativeTag` as it is private and readonly in `AppConfig`: https://github.com/vuejs/vue-next/blob/71a942b25a2cad61c3d670075523c31d296c7089/packages/runtime-core/src/apiCreateApp.ts#L46-L48

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -12,12 +12,14 @@ import {
   ComponentOptionsWithArrayProps,
   ComponentOptionsWithoutProps,
   ExtractPropTypes,
-  Component
+  Component,
+  AppConfig,
+  VNodeProps
 } from 'vue'
 
 import { config } from './config'
 import { GlobalMountOptions } from './types'
-import { mergeGlobalProperties, isString } from './utils'
+import { mergeGlobalProperties } from './utils'
 import { processSlot } from './utils/compileSlots'
 import { createWrapper, VueWrapper } from './vue-wrapper'
 import { attachEmitListener } from './emitMixin'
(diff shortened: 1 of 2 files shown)


src/mount.ts
```diff
@@ -195,7 +206,10 @@ export function mount(
 
   // AppConfig
   if (global.config) {
-    for (const [k, v] of Object.entries(global.config)) {
+    for (const [k, v] of Object.entries(global.config) as [
+      keyof Omit<AppConfig, 'isNativeTag'>,
```
lmiller1990, line 210: nice!

src/mount.ts
```diff
@@ -184,8 +193,10 @@ export function mount(
   if (global?.mocks) {
     const mixin = {
       beforeCreate() {
-        for (const [k, v] of Object.entries(global.mocks)) {
-          this[k] = v
+        for (const [k, v] of Object.entries(
+          global.mocks as { [key: string]: any }
+        )) {
+          ;(this as any)[k] = v
```
lmiller1990, line 199: TS is so ugly sometimes (or maybe due to Vue's dynamic nature it cannot be avoided...)

src/types.ts
```diff
@@ -21,7 +21,7 @@ export type FindAllComponentsSelector = NameSelector | string
 
 export type GlobalMountOptions = {
   plugins?: Plugin[]
-  config?: AppConfig
+  config?: Omit<AppConfig, 'isNativeTag'> // isNativeTag is readonly, so we omit it
```
lmiller1990, line 24: so this is how you do it, awesome, was trying to figure this out

lmiller1990:
Great, thanks a lot. `strict: true` would be great.

PR #124, opened 2020-05-18
chore: shallowMount overloads

cexbrayat:
`shallowMount` now has the same signatures than `mount`, to which it delegates the heavy work.

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -245,41 +245,6 @@ export function mount(
   return createWrapper(app, App, setProps)
 }
 
-// Functional component
-export function shallowMount<TestedComponent extends FunctionalComponent>(
-  originalComponent: TestedComponent,
-  options?: MountingOptions<any>
-): VueWrapper<ComponentPublicInstance>
-// Component declared with defineComponent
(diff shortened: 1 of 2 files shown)


src/mount.ts
```diff
@@ -245,8 +245,40 @@ export function mount(
   return createWrapper(app, App, setProps)
 }
 
+// Functional component
+export function shallowMount<TestedComponent extends FunctionalComponent>(
```
lmiller1990, line 249: If these are the same can we not share the type somehow? Or is this not possible? (Eg, are we not copy-pasting the same 50 or so lines twice?)

cexbrayat: This is a good question, but I don't know how we could easily do that. @pikax maybe?

pikax: is a bit tricky and dirty, but we could have something like this: 

```ts
export const shallowMount: typeof mount = (...args) => {
  mount(args[0], { ...args[1], shallow: true })
}
```
Typescript complains about it, but we can cast the function to `unknown` first

cexbrayat: @pikax I pushed an extra commit with your suggestion 👍 TSD is happy, so it's good enough for me :)

PR #125, opened 2020-05-23
Fix: trigger not working with some dot.notation modifiers

dobromir-hristov:
## Description

This PR fixes an issue with Vue 3 handling modifiers and keyCodes.

It should now allow passing modifiers the same way as you would do with in Vue. Main fixes are general key modifiers and system key modifiers.

* `trigger('keydown.meta.shift.c')` - will trigger a `cmd` + `shift` + `c` event listener.
* `trigger('keydown.13')` - will trigger on `enter` only, because keyCode is no longer supported.
* `trigger('click.right')` - will trigger a righclick `contextmenu` event. 
* `trigger('click.middle')` - will trigger a middle click event. 

### Context

While writing the docs for form handling I asked myself if we support `trigger('keydown.enter')` syntax, as I am sure we do in VTU 1, hence we have a dedicated test even - https://github.com/vuejs/vue-test-utils/blob/dev/test/specs/wrapper/trigger.spec.js#L73

However when I re-implemented the same test in VTU 2, it failed. I tried debugging, the trigger code is pretty much the same as VTU 1. I am not sure what is going on here :/

### DIFF
--- a/src/create-dom-event.ts
+++ b/src/create-dom-event.ts
@@ -14,7 +14,7 @@ interface EventParams {
 }
 
 // modifiers to keep an eye on
-const ignorableKeyModifiers = ['stop', 'prevent', 'self', 'exact', 'exact']
+const ignorableKeyModifiers = ['stop', 'prevent', 'self', 'exact']
 const systemKeyModifiers = ['ctrl', 'shift', 'alt', 'meta']
 const mouseKeyModifiers = ['left', 'middle', 'right']
 
@@ -97,6 +97,7 @@ function getEventProperties(eventParams: EventParams) {
   }
 
   // convert `shift, ctrl` to `shiftKey, ctrlKey`
+  // allows trigger('keydown.shift.ctrl.n') directly
   const systemModifiersMeta = systemModifiers.reduce((all, key) => {
     all[`${key}Key`] = true
     return all
--- a/src/create-dom-event.ts
+++ b/src/create-dom-event.ts
@@ -9,9 +9,45 @@ interface TriggerOptions {
 
 interface EventParams {
   eventType: string
-  modifier: string
-  meta: any
+  modifiers: string[]
   options?: TriggerOptions
+}
+
+// modifiers to keep an eye on
+const ignorableKeyModifiers = ['stop', 'prevent', 'self', 'exact', 'exact']
+const systemKeyModifiers = ['ctrl', 'shift', 'alt', 'meta']
+const mouseKeyModifiers = ['left', 'middle', 'right']
+
+/**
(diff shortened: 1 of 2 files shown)


dobromir-hristov:
Ok well this is apparently `keyCode` is no logner supported in Vue - https://codepen.io/dobromir/pen/vYNPyPx?editors=1011

src/create-dom-event.ts
```diff
@@ -33,52 +69,89 @@ export const keyCodesByKeyName = {
...
 
 function createDOMEvent(eventString: String, options?: TriggerOptions) {
-  const [eventType, modifier] = eventString.split('.')
-  const meta = eventTypes[eventType] || {
-    eventInterface: 'Event',
-    cancelable: true,
-    bubbles: true
-  }
+  // split eventString like `keydown.ctrl.shift.c` into `keydown` and array of modifiers
+  const [eventType, ...modifiers] = eventString.split('.')
```
lmiller1990, line 148: nice, this is cool

tests/trigger.spec.ts
```diff
@@ -50,6 +50,48 @@ describe('trigger', () => {
...
+
+      expect(handler).toHaveBeenCalledTimes(1)
+      expect(handler.mock.calls[0][0].button).toBe(1)
+      expect(handler.mock.calls[0][0].type).toBe('mouseup')
+    })
+
+    it('works with meta and key modifiers', async () => {
+      const handler = jest.fn()
+      const Component = {
+        template: '<div @click.meta.right="handler"/>',
```
lmiller1990, line 84: wow! I did not know you could combine modifiers like this.

src/create-dom-event.ts
```diff
@@ -9,11 +9,47 @@ interface TriggerOptions {
...
 interface EventParams {
   eventType: string
-  modifier: string
-  meta: any
+  modifiers: string[]
   options?: TriggerOptions
 }
 
+// modifiers to keep an eye on
+const ignorableKeyModifiers = ['stop', 'prevent', 'self', 'exact', 'exact']
```
lmiller1990, line 17: Where did this list come from? somewhere in Vue 3?

dobromir-hristov: Yeah, from the vOn transformer, they are used to wrap the callback

dobromir-hristov:
So I found the issue while prototyping the docs for form handling. We have complicated cases at work which I got inspired from to dig deeper.

The main issue was that Vue 3 did not seem to have keyCode anymore, so tests would fail with just eventStrings

lmiller1990:
Looks good to me. I'll approve, hit merge when you are happy.

PR #126, opened 2020-05-23
chore(config): proper typings for plugins

cexbrayat:
Adds typings to plugins, still in the quest for switching the TS config to `strict`


lmiller1990:
Thanks for all these!

PR #128, opened 2020-05-23
chore(emitMixin): proper typings

cexbrayat:
Adds typings to `emitMixin`, still in the quest for switching the TS config to `strict`

### DIFF
--- a/src/emitMixin.ts
+++ b/src/emitMixin.ts
@@ -4,9 +4,9 @@ export const attachEmitListener = () => {
   return {
     beforeCreate() {
       let events: Record<string, unknown[]> = {}
-      this.__emitted = events
-      const originalEmit = getCurrentInstance().emit
-      getCurrentInstance().emit = (event: string, ...args: unknown[]) => {
+      ;(this as any).__emitted = events
+      const originalEmit = getCurrentInstance()!.emit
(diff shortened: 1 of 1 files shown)


src/emitMixin.ts
```diff
@@ -4,9 +4,9 @@ export const attachEmitListener = () => {
   return {
     beforeCreate() {
       let events: Record<string, unknown[]> = {}
-      this.__emitted = events
-      const originalEmit = getCurrentInstance().emit
-      getCurrentInstance().emit = (event: string, ...args: unknown[]) => {
+      ;(this as any).__emitted = events
+      const originalEmit = getCurrentInstance()!.emit
```
lmiller1990, line 8: why do we need `!` here?

cexbrayat: `getCurrentInstance` returns this instance or null  (see https://github.com/vuejs/vue-next/blob/38f2d23a607cd7077da189ac274a3a0ad542cc1f/packages/runtime-core/src/component.ts#L400) So TS is not happy if we don't check if it's not null. As the mixin is only called by us, I think we're safe with just `!`.

lmiller1990: Ok!

PR #129, opened 2020-05-23
chore(find): proper typings

cexbrayat:
Adds some typings to the find functions, and handle a case surfaced by TS (but that I did not manage to reproduce via unit tests, so it might never happen).

### DIFF
--- a/src/utils/find.ts
+++ b/src/utils/find.ts
@@ -1,4 +1,9 @@
-import { VNode, ComponentPublicInstance } from 'vue'
+import {
+  ComponentPublicInstance,
+  VNode,
+  VNodeArrayChildren,
+  VNodeNormalizedChildren
+} from 'vue'
 import { FindAllComponentsSelector } from '../types'
 import { matchName } from './matchName'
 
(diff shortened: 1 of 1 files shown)


src/utils/find.ts
```diff
@@ -28,26 +33,43 @@ function matches(node: VNode, selector: FindAllComponentsSelector): boolean {
...
-    ;[...children].reverse().forEach((n: VNode) => {
-      nodes.unshift(n)
-    })
+    const reversedChildren = [...children]
+      .reverse()
+      .filter(notNullOrUndefinedOrVoid)
+    console.log(reversedChildren)
+    reversedChildren.forEach(
+      (node: string | number | boolean | VNodeArrayChildren | VNode) => {
+        nodes.unshift(node)
```
cexbrayat, line 55: I need help here, please.
If I follow correctly until that point, `children` is of type `VNodeNormalizedChildren`.
That means its elements, after filtering the potential null/undefined ones (which I added to be safe), can be either `string | number | boolean | VNodeArrayChildren | VNode` (see https://github.com/vuejs/vue-next/blob/24168bbb333727417ddcb310c615afc55ee923ad/packages/runtime-core/src/vnode.ts#L85-L92) . The current code considers it is always a VNode. Can we be sure of that? 
If yes, I cast force the cast and add a comment explaining why.
If no, we need to handle the other cases, but I'm not sure what needs to be done.
cc @dobromir-hristov

dobromir-hristov: I think I was just collecting all children, then in the `match` I am filtering away the unnecessary. 
Children can be other things, but I did not yet find such cases cases. 

`String` means its just a plain DOM element. 
`VNode` is Vue component instance.
Boolean I have no idea when 😆 
`VNodeArrayChildren` is probably for multi root components? idk.. :/

cexbrayat: Right, I did not manage to reproduce a case where we have `VNodeArrayChildren` either. To be safe, I added a branch if we end up with it, and call aggregateChildren on them. We should be safe in all cases.

cexbrayat:
The PR is now up for review.

src/utils/find.ts
```diff
@@ -28,26 +33,55 @@ function matches(node: VNode, selector: FindAllComponentsSelector): boolean {
   return false
 }
 
+/**
+ * Filters out the null, undefined and primitive values,
+ * to only keep VNode and VNodeArrayChildren values
+ * @param value
+ */
+function nodesAsObject<Node>(
```
lmiller1990, line 41: Is this actually useful? Sees much more complex - what is the benefit of this function? Isn't this effectively `any`?

cexbrayat: It is different from `any`.

It is used in a `filter` in `aggregateChildren` to remove the potential null objects in the children array, and also removes the primitive values which aren't interesting. 

It is also a type guard, so TS knows that when this function is used in the filter, the array contains only non null values, and non primitive values after filtering.  To my knowledged, this is still too hard for TS to infer these kinds of things by itself, and I believe it's still the recommended way to do it.

PR #133, opened 2020-05-30
feat: add setValue for setting v-model on Vue component form elements

dobromir-hristov:
## Description

Abstracts setting  `v-model` value of a Vue component instance. Useful for more complicated form inputs.

More about `v-model` on components - https://github.com/vuejs/rfcs/blob/master/active-rfcs/0011-v-model-api-change.md#detailed-design

### API

```js
const comp = wrapper.findComponent(Foo)
// for just v-model
comp.setValue(someValue)
// for v-model:foo
comp.setValue(someValue, 'foo')
```

## Reasoning

While writing the docs for forms, I realized that in the past, when I had to deal with complex Vue form components it is not as easy to set values on them, so that you can test various things.

Think multiselects, editable tables, uploads and so on, where a simple `find('input').setValue()` wont work, and looking for an appropriate class to find by (if any) is also not the right approach, as that may change at the developer's will.


lmiller1990:
Interesting one. You decided on `setModelValue` instead of just `setValue` since it's a little different?

I think this feature makes sense. I like 👍

lmiller1990:
Update for anyone reading: we talked about it in discord and the consensus is that we will bundle this with `setValue` for one glorious, unified way of interacting with `v-model`.

dobromir-hristov:
Renamed.

lmiller1990:
LGTM!

We need to update docs with this. I'll put more effort into getting docs into this repo - that would make it much easier to sync with the current source.

PR #134, opened 2020-06-01
repro: watcher not triggered on setProps

cexbrayat:
This is a minimal repro of something I ran into. It does work in the real app, but not in the unit test, so I _think_ `setProps` is not triggering the `watcherEffect`.

It is entirely possible that I am doing something weird here, we can close if that's the case :)

### DIFF
--- a/tests/setProps.spec.ts
+++ b/tests/setProps.spec.ts
@@ -1,4 +1,4 @@
-import { defineComponent, h, computed } from 'vue'
+import { defineComponent, h, computed, ref, watchEffect } from 'vue'
 
 import { mount } from '../src'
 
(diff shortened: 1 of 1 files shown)


tests/setProps.spec.ts
```diff
@@ -143,4 +143,58 @@ describe('setProps', () => {
...
+    expect(wrapper.html()).toContain('Hello')
+
+    // it should hide the alert when clicking on it
+    await wrapper.get('button').trigger('click')
+    expect(wrapper.html()).not.toContain('Hello')
+    expect(wrapper.vm.isDismissed).toBe(true)
+
+    // it should display the alert again if the parent component wants to (triggers the props watcher)
+    await wrapper.setProps({ shown: true })
+    expect(wrapper.vm.isDismissed).toBe(false)
```
cexbrayat, line 197: The test is failing here: `isDismissed` should be false, as a watcher resets its value when `props.show` is truthy. But it is still `true` at this point in the test.

lmiller1990:
setProps + watch was buggy in VTU v1 as well. There is a few issues about it still open.

I wonder why this is happening. I will dive into this sometime this week! Thanks for the repro.

tests/setProps.spec.ts
```diff
@@ -143,4 +143,58 @@ describe('setProps', () => {
...
+    expect(wrapper.find('button').exists()).toBe(true)
+    expect(wrapper.html()).toContain('Hello')
+
+    // it should hide the alert when clicking on it
+    await wrapper.get('button').trigger('click')
+    expect(wrapper.html()).not.toContain('Hello')
+    expect(wrapper.vm.isDismissed).toBe(true)
+
+    // it should display the alert again if the parent component wants to (triggers the props watcher)
+    await wrapper.setProps({ shown: true })
```
lmiller1990, line 196: should be 

```
await wrapper.setProps({ show: true })
```

not `shown`. But it still fails anyway.

tests/setProps.spec.ts
lmiller1990: I don't think this actually makes sense. `show` is `true` at the start of the test. it never becomes `false`. so doing `setProps({ show: true })` will not cause anything to happen, since the value of `show` is not changing, right?

At least this would be my expectation - `watchEffect` should only run when a value changes, right

cexbrayat: I think you're right 👍 I'll take another look and come with a better repro

PR #136, opened 2020-06-03
chore(stubs): proper typings

cexbrayat:
Also removes a `ts-ignore`.

### DIFF
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -1,29 +1,37 @@
-import { transformVNodeArgs, h, createVNode } from 'vue'
+import {
+  transformVNodeArgs,
+  h,
+  ComponentPublicInstance,
+  Slots,
+  ComponentOptions,
+  defineComponent,
+  VNodeProps,
+  VNodeTypes
+} from 'vue'
 import { hyphenate } from './utils/vueShared'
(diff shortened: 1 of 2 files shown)


src/stubs.ts
```diff
@@ -1,29 +1,37 @@
...
-const createStub = ({ name, props }: IStubOptions) => {
+const createStub = ({ name, props }: StubOptions): ComponentOptions => {
   const anonName = 'anonymous-stub'
   const tag = name ? `${hyphenate(name)}-stub` : anonName
 
-  const render = (ctx) => {
-    return h(tag, {}, getSlots(ctx))
+  const render = (ctx: ComponentPublicInstance) => {
+    // TS is not happy with this signature, so we cast the Slots as any
+    return h(tag, {}, getSlots(ctx) as any)
```
cexbrayat, line 31: TS is not happy here, as there is no signature of `h` matching `h(string, {}, Slots)`. But it works, so I added `as any` even if I don't like it...

cexbrayat: After digging a bit into it, I think this should be fixed in `vue-next`, and I submitted a PR: https://github.com/vuejs/vue-next/pull/1293
We can merge as is for now, and I'll remove the `as any` cast if Evan accepts my PR.

Issue #144, opened 2020-06-25
TS error with `findComponent(Foo).vm.$emit)`


lmiller1990:
This code: `await wrapper.findComponent(Hello).vm.$emit('greet')`

```
(property) $emit: unknown
This expression is not callable.
  Type '{}' has no call signatures.ts(2349)
```

Shim:

```
declare module '*.vue' {
  export default any
}
```

![image](https://user-images.githubusercontent.com/19196536/85719280-a0aa8900-b732-11ea-892b-41039967a25e.png)

lmiller1990:
This will be fixed when https://github.com/vuejs/vue-next/pull/1498 is merged and the next beta comes out

PR #145, opened 2020-06-25
Repro: add failing types test for wrapper.vm.$emit

lmiller1990:
Repro #144 

The code works fine if you `@ts-ignore`. Just types.

Any ideas? @pikax @cexbrayat maybe? 🤔 Oddly enough it knows that `emit` exists but does not think it is callable.

`vm.$.emit` is defined and works great, but that is the internal instance.

EDIT: we have this problem in the codebase too: https://github.com/vuejs/vue-test-utils-next/blob/c4787821f100524f195fd2365f42dd816634f07d/src/vue-wrapper.ts#L202

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,3 @@
+declare module '*.vue' {
+  export default any
+}
--- a/test-dts/wrapper.d-test.ts
+++ b/test-dts/wrapper.d-test.ts
@@ -1,6 +1,7 @@
 import { expectType } from 'tsd'
 import { defineComponent } from 'vue'
 import { mount } from '../src'
+import Hello from '../tests/components/Hello.vue'
 
 const AppWithDefine = defineComponent({
   template: ''
@@ -111,3 +112,6 @@ expectType<boolean>(domWrapper.classes('class'))
 // props
 expectType<{ [key: string]: any }>(wrapper.props())
 expectType<any>(wrapper.props('prop'))
+
+// vm
+expectType<Function>(wrapper.findComponent(Hello).vm.$emit)


cexbrayat:
I need to take a closer look, but I think this is expected as the shim for `*.vue` is `any`. So `$emit` becomes `unknown` (TS doesn't know it exists, but `any` allows anything).

You need to properly shim `*vue`, and that should work, I use it in various projects:
```typescript
declare module '*.vue' {
  import { defineComponent } from 'vue';
  export default defineComponent();
}
```
Not ideal typing, but should be better than `any`.
I really hope to get rid of all the `@ts-ignore` of the codebase, I need to find time to get back to it!

lmiller1990:
If this is related to the shim, though, we would not see this problem in the main codebase would we? It is the same error there, which works entirely we TS. Not sure this is a shim issue.

`findComponent` is typed to return a `PublicComponentInstance` - I am sure this is correct, and it *does* work correctly, so I am require unsure why TS flags this type as `unknown` when we try to call it (but not when we reference it; eg `$emit` is ok`, but `$emit()` is not).

Interestingly enough you can do `vm.$.emit` and this works fine (no TS error).

cexbrayat:
See #146 
We probably have the same issue in the codebase.

lmiller1990:
Thanks. I don't think the issue in the codebase is related to the shim though - it doesn't touch `.vue` files.

Very weird. Thanks for this, though, at least I can remove a ts-ignore from my application's tests.

lmiller1990:
Fixed in latest beta

PR #147, opened 2020-06-27
tests: reproduce router problem

lmiller1990:
I am getting this odd html

```
Expected substring: "This is the home route"
Received string:    "<div> Home <a href=\"/posts\" data-testid=\"to-posts\" class=\"\">To Posts</a><!----></div>"
```

Not sure why but it will not render `<router-view />`. Hm.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,5 +79,8 @@
         "dom"
       ]
     }
+  },
+  "dependencies": {
+    "vue-router": "^4.0.0-alpha.13"
   }
 }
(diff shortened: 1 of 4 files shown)


lmiller1990:
Ok maybe not even related to VTU... https://github.com/vuejs/vue-router-next/issues/330

lmiller1990:
Everything is fine; I forgot to read the docs for the breaking changes in Vue Router next. You can find the solution [here](https://github.com/vuejs/vue-router-next/issues/330#issuecomment-650661328).

tl;dr navigation is aync. We should include some info about this in the VTU docs.

PR #150, opened 2020-06-30
repro: beta.16 regressions

cexbrayat:
I was trying to update VTU 2.x from Vue beta.15 to beta.16 (beta.17 is the same), and we are having some regressions.

It looks like an SFC with just a template tag was passed to `transformVNodeArgs` as:
`{ render: [Function: render], name: 'component-without-name' }`
and now is passed as:
`{ render: [Function: render] }`
As we need the name for stubbing/shallow mount, we have test failures.

### DIFF
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -75,6 +75,17 @@ export function stubComponents(
   shallow: boolean = false
 ) {
   transformVNodeArgs((args) => {
+    const locallyRegisteredComponents = (args[0] as any).components as
+      | Record<string, VNodeTypes>
+      | undefined
+    if (locallyRegisteredComponents) {
+      for (const registrationName in locallyRegisteredComponents) {
+        const component = locallyRegisteredComponents[registrationName]
+        if (!component['name'] && !component['displayName']) {
+          component['name'] = registrationName
+        }
+      }
+    }
     const [nodeType, props, children, patchFlag, dynamicProps] = args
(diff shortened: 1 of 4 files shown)


cexbrayat:
PR is now green with the second commit. Do not merge right now, as I need to think about it, and document properly.

src/stubs.ts
```diff
@@ -75,6 +75,17 @@ export function stubComponents(
...
 ) {
   transformVNodeArgs((args) => {
+    const locallyRegisteredComponents = (args[0] as any).components as
+      | Record<string, VNodeTypes>
+      | undefined
+    if (locallyRegisteredComponents) {
+      for (const registrationName in locallyRegisteredComponents) {
+        const component = locallyRegisteredComponents[registrationName]
+        if (!component['name'] && !component['displayName']) {
+          component['name'] = registrationName
```
lmiller1990, line 85: I wonder if this mutation is going to cause and problems 🤔

cexbrayat:
@lmiller1990 please wait before releasing. I think we can do better (at least, document and test this). Evan commented on Dicsord with a similar idea, but in way that might not lead to problems as this one. But I need to find some time to try it.

lmiller1990:
I did a release before I saw this, I thought this bug was problematic enough to warrant a release (still in alpha so yeah).

I will also try to find some time to work on this, been busy between triaging issues in V1, docs...

PR #151, opened 2020-07-01
refactor: properly find/stub anonymous component

cexbrayat:
Instead of always patching the missing name with the registration name, we now look into the parent registration only if the name is missing.
Some cases were trickier to handle (for example when we try to find a component that has been stubbed), but it looks like we are on par with what we had previously and we do not have the risk of multiple registration names erasing each other anymore.


dobromir-hristov:
Looks good!

Issue #152, opened 2020-07-04
Vue Router test utils


lmiller1990:
In writing some small apps and the VTU docs, I encounted some caveats when testing components using Vue Router. This is mainly because the [navigation is now all async](https://github.com/vuejs/vue-router-next#known-issues) (which I think it is a great design decision, it fits into Vue's async update model perfectly).

> This happens because navigations are all asynchronous now.

When I first tried writing some tests using Vue Router, I didn't realize I had to do `await router.isReady`, or the initial navigation may not be completed before the component renders. 

Also, I triggered a click on a `<router-link>` I got an error. It turns out that you need to wait for the navigation to complete (makes sense with the new async navigation model).

Here is a full snippet demonstrating what is currently required when using VueRouter in a jsdom environment (sans VTU).

```ts
it('avoids several caveats to succesfully click an <a> tag", async () => {
  const el = document.createElement('div')
  el.id = 'app'
  document.body.appendChild(el)
  const app = createApp(App)
  app.use(router)

  router.push('/')
  await router.isReady() // <- need to do this

  app.mount(el)

  const event = document.createEvent('Event')
  event.initEvent('click')
  let promise = new Promise(resolve => {
    const remove = router.afterEach(() => {
      remove() // <- need to do this
      resolve() // <- need to do this
    })
  })
  document.getElementById('post')!.dispatchEvent(event)
  await promise // <- need to do this
  console.log(document.body.outerHTML)
})
```

The user must `await isReady`, and also `await resolve` using the `afterEach` callback to ensure the navigation triggered when clicking on the `<a>` tag (from `<router-link>`) before continuing their tests. Kind of like how you need to do `await nextTick` for Vue to update the DOM.

We should bundle some goodies to make testing Vue Router a more smooth experience. @posva suggested `await routerMock.pendingNavigation()`. I think we need both good utils for testing with the actual VueRouter, and a mocked VueRouter. 

This is the kind of thing you would previously write in VTU 1.x with Vue 2.x and Vue Router 3.x:

```js
const router = createRouter()

const wrapper = mount(App. {
  global: {
    plugins: [router]
  }
})

await wrapper.find('a').trigger('click') // it "just works"
```

I see a few options, but I think the best and most simple is some additional methods on `wrapper`. Here are some ideas I had:

```js
// react testing library style, like act(() => ...
await doNavigation(() => {
  wrapper.find('a').trigger('click')
})

// or a more VTU-like like API
[...]

Issue #103, opened 2020-05-02
Add a few tests for Functional Components


dobromir-hristov:
I think we are lacking tests for functional components. Those types of components resemble allot more statefull components now.

A functional component signature is now - https://github.com/vuejs/rfcs/blob/master/active-rfcs/0031-attr-fallthrough.md#functional-components

```js
const Component = (props, { slots, attrs, emit }) => h('div', { class: ['a'] }, slots)

Component.props = { /*...*/ }
Component.inheritAttrs = false
Component.displayName = 'FunctionalComponent'
```

I can take this over and try to add test cases on as many places as I can.

We also need to test asyncFunctionalComponents - https://github.com/vuejs/rfcs/blob/async-component/active-rfcs/0026-async-component-api.md

lmiller1990:
We also haven't got anything for `defineAsyncComponent`, either. IIRC this had some weird bugs in beta, we should test it. One common case is loading routes/pages async.

lmiller1990:
We have #187 and #158 relating to bugs in functional components, so I will close this.

PR #154, opened 2020-07-12
refactor: extract the repeated logic on classes.spec.ts

cuixiaorui:
I extracted some duplicate code while reading the test file
But I don't know what you think about this code style

### DIFF
--- a/src/utils/vueShared.ts
+++ b/src/utils/vueShared.ts
@@ -19,17 +19,3 @@ const hyphenateRE = /\B([A-Z])/g
 export const hyphenate = cacheStringFunction((str: string): string => {
   return str.replace(hyphenateRE, '-$1').toLowerCase()
 })
-
-export const enum ShapeFlags {
-  ELEMENT = 1,
-  FUNCTIONAL_COMPONENT = 1 << 1,
-  STATEFUL_COMPONENT = 1 << 2,
-  TEXT_CHILDREN = 1 << 3,
-  ARRAY_CHILDREN = 1 << 4,
-  SLOTS_CHILDREN = 1 << 5,
-  TELEPORT = 1 << 6,
-  SUSPENSE = 1 << 7,
-  COMPONENT_SHOULD_KEEP_ALIVE = 1 << 8,
-  COMPONENT_KEPT_ALIVE = 1 << 9,
-  COMPONENT = ShapeFlags.STATEFUL_COMPONENT | ShapeFlags.FUNCTIONAL_COMPONENT
-}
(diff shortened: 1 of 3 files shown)


lmiller1990:
Hi! Thanks for taking the time for this PR.

## Regarding `ShapeFlags`:

The idea was for this library to have 0 dependencies. This makes the build step much more simple, as well as usage (for example, if you use vue-test-utils in the browser and have the `@vue/shared` dependency, the user needs to make sure that dependency is *also* available globally, which is not really ideal). Since `ShapeFlags` doesn't change much (if ever), I am not sure if there is much benefit in adding `@vue/shared` as a dependency dependency.

## Regarding the refactor in the tests

Although there is less code, I am not really a fan of this refactor. The component with the relevant class (`class-a`) is defined around line 8, however the relevant test is on line 47. In my opinion, each test should be "self contained", so you can look at the test and see all the relevant information, as opposed to scanning the file to find out what the component looks like.

Finally, I've found that enforcing immutability whereever possible leads to more understandable code. If you are constantly mutating and reassigning `wrapper`, it can lead to weird bugs where the value is reassigned accidentally. 

For these reasons I don't see value in merging this PR. If you would like to pick up some work, though, I have some recommendations!

Some easy issues to pick up:

- we have a mix of file names.  we have `camel-case.ts` and `camelCase.ts`. We should pick one. You can rename them and preserve the git history with `git mv`. I'd reommend `camelCase.ts`, that is what Vue core is doing. This would be a very easy improvement to make.
- We need some tests for functional components, there are not many (if any?) tests for those. See https://github.com/vuejs/vue-test-utils-next/issues/103

Some other issues that need attention that will take a bit more effort:

- Fix this PR: https://github.com/vuejs/vue-test-utils-next/pull/52. This is the `setData` feature - we don't want to use `lodash` since it makes the bundle huge, ideally we would write out own `merge` function and get all those tests to pass.
- For something more ambitious, we are exploring ways to improve testing Vue Router. Issue: https://github.com/vuejs/vue-test-utils-next/issues/152. Writing tests for the router is a bit painful now. Here is a [working example](https://github.com/lmiller1990/vuejs-composition-course/blob/835cebf03464c0f6d23c571816cd3cebb512bcb8/src/ShowPost.spec.ts#L15) of a Router test - I need to do lots of `await`. 

If you want more help on these issues, let me know and I can provide some guidance.

cuixiaorui:
Thank you very much

I love testing, I'm a TDD practitioner, I want to help improve VTU,

I'll go and look at some of the PR that I mentioned

My English may not be very good, I will ask you if there is any question

PR #155, opened 2020-07-13
chore: remove ShapeFlags

cuixiaorui:
Change to export from @vue/shared

No repeat declaration


lmiller1990:
I think we decided to avoid the `@vue/shared` dependency and copy paste instead to simplify the bundle when used as an es module, for example. Maybe @JessicaSachs has some more context on this (mentioned [here](https://github.com/vuejs/vue-test-utils-next/pull/154#issuecomment-657509627), too)?

cuixiaorui:
I did some tests on this problem

After packing with yarn build

```js
private Get hasMultipleRoots(): Boolean {

    return this. Vm. $. SubTree. ShapeFlag. = = = ShapeFlags ARRAY_CHILDREN
}
```
After the build
```ts
private Get hasMultipleRoots(): Boolean {
    return this. Vm. $. SubTree. ShapeFlag. = = = ShapeFlags ARRAY_CHILDREN
}
```
Directly changes shapeflags.array_children to a number

With the rollup Tree Shaking feature, you don't need to worry about vue-test-utils.[xxx].js getting bigger

As far as package dependency is concerned, we've already had the user install the Vue, and @vue/share is what will exist once the Vue is installed, so should we really worry about this?

lmiller1990:
I think you are correct here, I did not realize it would be able to do this. Are we able to use this in a browser (and with `vite`)? If this is the case, we should definitely merge this PR, it's a great improvement. I'll give it a test this evening. Thanks!

PR #157, opened 2020-07-15
feat: add flush promises

### LINKED ISSUE #137: Helpers: nextTick/flushPromises
Follwing up on a discussion on Discord, I wanted to open this issue to gather feedback, experiences, and opinions on how to handle async testing and the usage of `nextTick` and `flushPromises`.

## Differences between `nextTick` and `flushPromises`

I couldn't explain it any better than here: https://github.com/testing-library/react-testing-library/issues/11

## Current state of the art

We're currently returning `nextTick` on several methods.

> I think I was expecting to get away with await wrapper.something() most of the time.

by @cexbrayat. Turns out you might need an indeterminate number of `nextTick` calls to get everything up to date, so you end up using `flushPromises`. Happened the same to me while writing [docs for HTTP requests](https://github.com/vuejs/vue-test-utils-next-docs/pull/26).

Also, there's this question by @JessicaSachs that is worth checking/considering:

> `nextTick` might safely be able to be the default? I don't remember if this prohibits testing of async created hooks (I'm thinking in Vue 2)

### 1. "automagically" call `flushPromises`

The pros is that users have one less thing to worry about – and a thing which is quite low-level.

On the other hand, we might lose the ability to test intermediate states (such as "Loading states") before the promise resolved/rejected. If that's the case, this is a no-go.

### 2. Leave it as is and rely on docs

Pros is that current implementation of VTU-next would work.

Cons is that we all know that explaining something in a documentation does not translate well with "our users will know about this and will take it into account" 😅 

### 3. Create a tool (method? helper?) to wrap `nextTick` and `flushPromises`

Not sure about the _shape_ of that tool, but something along the lines of [waitFor](https://testing-library.com/docs/dom-testing-library/api-async#waitfor). Other alternatives mentioned on Discord are `wrapper.findAsync`.

---

disclaimer: this issue is meant to serve as a place to discuss alternatives. After that, if we reach a consensus, I open to write a proper RFC with the suggested proposal 👍

lmiller1990:
resolves #137

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,6 @@
     "babel-jest": "^25.2.3",
     "babel-preset-jest": "^25.2.1",
     "dom-event-types": "^1.0.0",
-    "flush-promises": "^1.0.2",
     "husky": "^4.2.3",
     "jest": "^25.1.0",
     "jsdom": "^16.2.2",
--- a/src/index.ts
+++ b/src/index.ts
@@ -3,5 +3,14 @@ import { RouterLinkStub } from './components/RouterLinkStub'
 import { VueWrapper } from './vue-wrapper'
 import { DOMWrapper } from './dom-wrapper'
 import { config } from './config'
+import { flushPromises } from './utils/flushPromises'
 
-export { mount, shallowMount, RouterLinkStub, VueWrapper, DOMWrapper, config }
+export {
+  mount,
+  shallowMount,
+  RouterLinkStub,
+  VueWrapper,
+  DOMWrapper,
+  config,
+  flushPromises
+}
--- a/src/utils/flushPromises.ts
+++ b/src/utils/flushPromises.ts
@@ -1,0 +1,7 @@
+const scheduler = typeof setImmediate === 'function' ? setImmediate : setTimeout
+
+export function flushPromises() {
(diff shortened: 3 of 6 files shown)


afontcu:
When released, we should update docs too ;) https://vuejs.github.io/vue-test-utils-next-docs/guide/http-requests.html I can do that

lmiller1990:
Sure. Kiaking did a ton of work with Vitepress, we can probably move to that soon and move the docs into this repo.

afontcu:
> Sure. Kiaking did a ton of work with Vitepress, we can probably move to that soon and move the docs into this repo.

yeah, been reading discord. It's on my list ;)

NeonWizard:
> Sure. Kiaking did a ton of work with Vitepress, we can probably move to that soon and move the docs into this repo.

Where can the docs be found for this? The link above isn't working.

afontcu:
> Where can the docs be found for this? The link above isn't working.

You mean docs for Vue Test Utils next? Here: https://next.vue-test-utils.vuejs.org

lmiller1990:
The source is here https://github.com/vuejs/vue-test-utils-next/tree/master/docs

PR #162, opened 2020-07-20
test:returns the html when mounting multiple nested components

weihaiyu:
returns the html when mounting multiple nested components


lmiller1990:
Hi! Just checking, is this fixing a bug or just adding some more tests?

lmiller1990:
I think we already have a test for this, so I am going to close this one for now. Thanks for the effort. If you are still keen on contributing let me know, there is a few outstanding issues we need to tackle, mainly around functional components.

PR #167, opened 2020-07-30
fix: find component after rc.5 refactor (and other minor fixes)

cexbrayat:
Each commit explains the reasoning behind the changes, but TL;DR: we had some compilation issue easy to fix, and a small refactor to do after the component resolution changes that Evan did in rc.5.

### DIFF
--- a/src/utils/find.ts
+++ b/src/utils/find.ts
@@ -42,7 +42,7 @@ function matches(node: VNode, selector: FindAllComponentsSelector): boolean {
 
     // if a name is missing, then check the locally registered components in the parent
     if (node.component.parent) {
-      const registry = (node.component.parent as any).components
+      const registry = (node.component.parent as any).type.components
       for (const key in registry) {
         // is it the selector
         if (!selectorName && registry[key] === selector) {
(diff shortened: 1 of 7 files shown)


src/utils/find.ts
```diff
@@ -42,7 +42,7 @@ function matches(node: VNode, selector: FindAllComponentsSelector): boolean {
 
     // if a name is missing, then check the locally registered components in the parent
     if (node.component.parent) {
-      const registry = (node.component.parent as any).components
+      const registry = (node.component.parent as any).type.components
```
lmiller1990, line 45: Nice, do we need a check to see if `type` is not null?

test-dts/shallowMount.d-test.ts
```diff
@@ -20,13 +20,12 @@ let wrapper = shallowMount(AppWithDefine, {
...
 
-// can receive extra props
-// ideally, it should not
-// but the props have type { a: string } & VNodeProps
-// which allows any property
-shallowMount(AppWithDefine, {
-  props: { a: 'Hello', c: 2 }
-})
+// can not receive extra props
+expectError(
```
lmiller1990, line 24: nice

src/utils/find.ts
cexbrayat: Good question but it is not marked as nullable in the TS definition of `ComponentInternalInstance`, so I think we're good (but I do agree we would feel safer if we had `strict: true`)

PR #168, opened 2020-07-30
feat: findComponent can find component inside a Suspense

cexbrayat:
Slightly updates the find implementation to:
- check the suspense `subtree` if it is resolved
- check the suspense `fallbackTree` otherwise

A test has been added to check if we can find a component in both cases.


lmiller1990:
@dobromir-hristov not do I - maybe it was added after. Glad we have this now!

PR #177, opened 2020-08-13
(wip) feat: move docs to main repo using Vitepress

lmiller1990:
It's coming


lmiller1990:
Problem is Jekyll ignores `_` files, we can use a Jekyll ignore file.

afontcu:
Cool! 🚀  Do we need to commit autogenerated files, though? We could move docs infra to Netlify, and just set the build command, so that Netlify handles the build process and deploys the right version.

lmiller1990:
notes for myself

```
vi node_modules/vitepress/dist/client/theme-default/components/SideBar.js
```

```
function resolveMultiSidebar(config, path, headers, depth) {
    const paths = [
        path,
        Object.keys(config)[0]
    ]
    const item = paths.map(x=>config[getPathDirName(x)]).find(Boolean)

    if (Array.isArray(item)) {
        return resolveArraySidebar(item, depth);
    }
    if (item === 'auto') {
        return resolveAutoSidebar(headers, depth);
    }
    return [];
}
```

lmiller1990:
IT KIND OF WORKS. But not really, still some 404s, and I had to hack vitepress. https://vuejs.github.io/vue-test-utils-next/guide/installation.html

lmiller1990:
Unhill battle. I will come back to this when Vite is more ready.

PR #182, opened 2020-08-20
bugfix/unmounting-on-multi-root

JeremyWuuuuu:
- Fix the logic which causes unmounting fail when it's multi-root component

### DIFF
--- a/tests/unmount.spec.ts
+++ b/tests/unmount.spec.ts
@@ -1,0 +1,32 @@
+import { mount } from '../src'
+
+const AXIOM = 'Rem is the best girl'
+
+describe('Unmount', () => {
+  it('works on single root component', () => {
+    const Component = {
+      template: `
+        <div>${AXIOM}</div>
+      `,
+      onErrorCaptured(err: Error) {
+        throw err
+      }
+    }
+    const wrapper = mount(Component)
+    expect(() => wrapper.unmount()).not.toThrowError()
+  })
+
+  it('works on multi-root component', () => {
+    const Component = {
+      template: `
+        <div>${AXIOM}</div>
+        <div>${AXIOM}</div>
+      `,
+      onErrorCaptured(err: Error) {
+        throw err
+      }
+    }
+    const wrapper = mount(Component)
+    wrapper.unmount()
+  })
+})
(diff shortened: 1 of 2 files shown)


JeremyWuuuuu:
I went through the mounting function and found out that the whole part of `parentElement.removeChild(child)` is kind of unnecessary when you eventually calling the `app.unmount(element)` method, so I simply removed the code here
https://github.com/vuejs/vue-test-utils-next/blob/e5aee88ec6437a7d28dd0a4a9d0e84bbe3a8df40/src/vueWrapper.ts#L219
and use `app.unmount(this.parentNode)` to remove mounted DOM.

JeremyWuuuuu:
@lmiller1990

tests/unmount.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+      `
+    }
+    const wrapper = mount(Component, {
+      props: {},
+      global: {
+        config: {
+          errorHandler
+        }
+      }
+    } as any) // The type checking keeps complaning about unmatched type which might be a bug
```
JeremyWuuuuu, line 22: Here I tried it with 
```ts
mount(Component, {
  global: {
    config: ...
  }
}
```
Type system seems not agree with this syntax when custom component is mounting with global options

lmiller1990: Interesting, will deal with the types later, it's outside the scope of this ticket. Great job!

tests/unmount.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+    wrapper.unmount()
+    expect(errorHandler).not.toHaveBeenCalled()
+  })
+
+  it('works on multi-root component', () => {
+    const errorHandler = jest.fn()
+    const Component = defineComponent({
+      template: `
+        <div>${AXIOM}</div>
```
lmiller1990, line 31: I don't think we need `AXIOM`, it doesn't really add much to the test. I am not sure who Rem is.

Can we just have `<div /><div />` or something?

JeremyWuuuuu: Yep, sure

Issue #180, opened 2020-08-17
findComponent not working as expected for components defined with `<template>`


lmiller1990:
```js
it.only('works', () => {
    const ComponentA = {
      name: 'ComponentA',
      template: `<div class="comp-a"><slot /></div>`
    }
    const ComponentB = {
      name: 'ComponentB',
      template: `<div class="comp-b"><slot /></div>`
    }
    const Main = {
      components: { ComponentA, ComponentB },
      template: `
        <ComponentA>
          <ComponentB>1</ComponentB>
          <ComponentB>2</ComponentB>
          <ComponentB>3</ComponentB>
        </ComponentA>
      `
    }

    const App = {
      components: { Main },
      template: '<Main />'
    }

    const wrapper = mount(App)

    wrapper.getComponent(ComponentB)
```

Cannot find `ComponentB`. If you do `mount(Main)` instead, it works.

@dobromir-hristov do you have some ideas? I spent some time but I still am not super clear on how everything is stored in the `subTree` property 🤔

dobromir-hristov:
I will check later today

lmiller1990:
Spent a few hours on this, recursion hurts my head. TS also makes this someway annoying to work on (probably because I am bad at it, haha).

Let me know if you come up with anything.

lmiller1990:
Edit: seems like `node.component.subTree.children` contains a mysterious `default` entry which is a render function and nothing else. I guess Vue realizes `<App>` is just rendering another component and is doing something (converting to a functional component maybe?) I have not seen the `default` entry in `subTree` before.

lmiller1990:
So... seems something is different with render functions vs templates.

```js
  it('works', () => {
    const ComponentA = {
      name: 'ComponentA',
      template: `<div class="comp-a"><slot /></div>`
    }
    const ComponentB = {
      name: 'ComponentB',
      template: `<div class="comp-b"><slot /></div>`
    }
    const Main = {
      name: 'Main',
      components: { ComponentA, ComponentB },

      /** Works with render function. Does NOT work with template */
      render() {
        return h(ComponentA, h(ComponentB, '1'))
      }
      // template: `
      //   <component-a>
      //     <component-b>1</component-b>
      //   </component-a>
      // `
    }

    const App = {
      components: { Main },
      name: 'App',
      render() {
        return h(Main)
      }
    }

    const wrapper = mount(App)

[...]

PR #184, opened 2020-08-23
WIP: findComponent edge cases

lmiller1990:
Hi all! This is a bug write-up and I would like some feedback from everyone.

TL;DR is `findComponent` seems very difficult support. I could be wrong but either way here is my findings and some options.

The current implementation very buggy (to the point it isn't really useful). See this issue for some context: https://github.com/vuejs/vue-test-utils-next/issues/180

I rewrote the `findVNodes` function to covert some of the edge cases. How it works: basically all components have the concept of a `subTree`, which stores the components VDOM state. We recurse down each level via `subTree` or `children`. It works for some cases, but not for others. Here are the cases I cannot find a solution for:

- functional components. No instance, so you cannot `findComponent` them, or at least when you do find them you cannot get the instance, which means no `VueWrapper`. I found a fix for this in this PR. But, since there is no instance, you cannot access `.vm` and all the stuff you would expect (like props etc). We could return a `DOMWrapper`, but this feels wrong. Users will expect `findComponent` to find a component. The whole point of splitting `find` and  findComponent` was to emphasis the difference in the two.
- slots and inline templates. These is the real blocker here! You can see how I access slot content in this this PR.  Basically, when using slots and components with inline `templates`, the `subTree` is different and there seems not way to get the instance via recursing down.

Some (not all) these issues seem to go away when we use SFC (via `vue-jest`). I guess we need to dig deeper into how Vue compiles things. The original issue in #180 is fixed by using SFCs. 🤷‍♂️ Interesting. Other are not.

Here are my proposed solutions:

- add an additional wrapper. This would something like `FunctionalWrapper` that returns the non-instance with some limited methods we can support (eg, `html()` and `attributes` for static HTML attributes). Not ideal.
- dive deeper and use `@vue/compiler-sfc` to compile all non-SFC components to SFCs. This seems it might work. This is how `vue-jest` works and pre-compiled SFCs seem not to have the same caveats. This would be a long term goal, at least if I need to write the code (I also need to focus on other bugs, vue-jest, VTU v1...).
- (**my favorite solution**) *until* we figure out how to solve the problems for non SFC components, put a warning that says "findComponent currently only supports SFCs. Non SFC support is experimental. We are working on supporting all manner of components". The reason I like this is so I don't have to keep telling people about work-arounds and triaging issues about the same thing. I'd much prefer a library with limitations that works really well, than one without limitations that doesn't. It feels more polished.

Tagging all the people who have expressed interested in contributing to and improving VTU in the last few weeks. Please give me your input!

Here is a screenshot explaining the problem:

image

image

@afontcu @dobromir-hristov @JeremyWuuuuu @JessicaSachs @cexbrayat @Jokcy @AngeloGulina @AtofStryker

### DIFF
--- a/tests/findComponent.spec.ts
+++ b/tests/findComponent.spec.ts
@@ -1,4 +1,4 @@
-import { defineComponent, nextTick } from 'vue'
+import { defineComponent, nextTick, h } from 'vue'
 import { mount } from '../src'
 import Hello from './components/Hello.vue'
 import ComponentWithoutName from './components/ComponentWithoutName.vue'
@@ -37,7 +37,46 @@ const compA = defineComponent({
   components: { compB, Hello }
 })
 
+const createRenderComponent = (
+  name: string,
+  props: Record<string, string> = {},
+  children?: any
+) => ({
+  name,
+  render() {
+    return h('div', props, children)
+  }
+})
+
 describe('findComponent', () => {
+  it('works with simple render functions', () => {
+    const Child = {
+      name: 'Child',
+      render() {
+        return h('div', { id: 'child' })
+      }
+    }
+    const Parent = {
+      name: 'Parent',
+      render() {
+        return h(Child)
+      }
+    }
+
+    const wrapper = mount(Parent)
+    wrapper.getComponent(Child)
+  })
+
+  it('works with simple render functions many layers deep', () => {
+    const A = createRenderComponent('A', { id: 'a' })
+    const B = createRenderComponent('B', { id: 'b' }, h(A))
+    const C = createRenderComponent('C', { id: 'c' }, h(B))
+    const Parent = createRenderComponent('Parent', { id: 'parent' }, h(C))
+
+    const wrapper = mount(Parent)
+    wrapper.getComponent(A)
+  })
+
   it('does not find plain dom elements', () => {
     const wrapper = mount(compA)
     expect(wrapper.findComponent('.domElement').exists()).toBeFalsy()
@@ -171,18 +210,16 @@ describe('findComponent', () => {
     expect(wrapper.findComponent(Hello).unmount).toThrowError()
   })
 
-  // https://github.com/vuejs/vue-test-utils-next/issues/173
-  const ComponentA = {
-    name: 'ComponentA',
-    template: `<div><slot></slot></div>`
-  }
-
-  const ComponentB = {
-    name: 'ComponentB',
-    template: '<div><slot></slot></div>'
-  }
-
   it('finds nested components and obtains expected html and innerText', () => {
+    const ComponentA = {
+      name: 'ComponentA',
+      template: `<div><slot></slot></div>`
+    }
+    const ComponentB = {
+      name: 'ComponentB',
+      template: '<div><slot></slot></div>'
+    }
+
     const wrapper = mount({
       components: {
         ComponentA,
@@ -201,6 +238,15 @@ describe('findComponent', () => {
   })
 
   it('finds nested components and obtains expected html and innerText', () => {
+    const ComponentA = {
+      name: 'ComponentA',
+      template: `<div><slot></slot></div>`
+    }
+    const ComponentB = {
+      name: 'ComponentB',
+      template: '<div><slot></slot></div>'
+    }
+
     const wrapper = mount({
       components: {
         ComponentA,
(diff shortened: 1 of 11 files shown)


Jokcy:
By now, I am trying to undestand the main logic of VTU, so I copied the `findAllVnodes` function to my demo code and debug it. I find in the `ComponentA, ComponentB` demo, if I add some logic to `findAllVnodes` I can find all vnode and proxy I needed:
```ts
 aggregateChildren(nodes, node.children)
    if (node.component) {
      // match children of the wrapping component
      aggregateChildren(nodes, node.component.subTree.children)
      aggregateChildren(nodes, [node.component.subTree])
      if (node.component.subTree.component) {
        aggregateChildren(
          nodes,
          node.component.subTree.component.subTree.children,
        )
      }
    }
```
the main difference is I tried ` node.component.subTree.component.subTree.children` and `node.component.subTree`, now I can find App, Main, Comp1 and Comp2.
I don't know whether it will help, I still trying hard to understand the code.

lmiller1990:
Interesting - you can't always assume `component` will exist, or `subTree` or `children`, but still good progress.

I think a recursive solution (like my PR) will make this much easier to understand. Either way - my next question is what do you get when you find `Comp2`? (I think `ComponentB` in your example? Are you using `template`, `render` or `setup`? It seems to make a difference).

If you can access the `vnode`, that is a good start. Did you try with an example using `template` and `slots`?

Jokcy:
@lmiller1990 I try to explain it.
what we need to know is the difference between subTree and children
```html
<Comp><div>name</div></Comp>
```
`<div>name</div>` is children of `Comp`
```
{
  name: 'Comp',
  render() {
    return <div><slot /></div>
  }
}
```
`<div><slot /></div>` is subTree of `Comp`
So in the this demo:
```ts
const Comp1 = defineComponent({
  name: 'Comp1',
  setup() {
    return () => <div>Comp1</div>
  },
})

const Comp2 = defineComponent({
  name: 'Comp2',
  setup() {
    return () => <div>Comp2</div>
  },
})

const Main = defineComponent({
  name: 'Main',
  setup() {
    return () => (
      <div>
        <Comp1 />
        <Comp2 />
      </div>
    )
  },
})

const App = defineComponent({
  name: 'App',
  setup() {
    return () => <Main />
  },
})
```
if we render `App`, `Main` is subTree of `App`, we won't find `Main` in any children of any vnode. And also `Main` has it's own subTree, the will be in the subTree of `Main`, so if we ignore `subTree` from vnode of App, we won't find and Component here. 

I tried my demo with `setup` and `jsx`, I will try `render` later.

Jokcy:
> @lmiller1990 I try to explain it.
> what we need to know is the difference between subTree and children
> 
> ```
> <Comp><div>name</div></Comp>
> ```
> 
> `<div>name</div>` is children of `Comp`
> 
> ```
> {
>   name: 'Comp',
>   render() {
>     return <div><slot /></div>
>   }
> }
> ```
> 
> `<div><slot /></div>` is subTree of `Comp`
> So in the this demo:
> 
> ```ts
> const Comp1 = defineComponent({
>   name: 'Comp1',
>   setup() {
>     return () => <div>Comp1</div>
>   },
> })
> 
> const Comp2 = defineComponent({
>   name: 'Comp2',
>   setup() {
>     return () => <div>Comp2</div>
>   },
> })
> 
> const Main = defineComponent({
>   name: 'Main',
>   setup() {
>     return () => (
>       <div>
>         <Comp1 />
>         <Comp2 />
>       </div>
>     )
>   },
> })
> 
> const App = defineComponent({
>   name: 'App',
>   setup() {
>     return () => <Main />
>   },
> })
> ```
> 
> if we render `App`, `Main` is subTree of `App`, we won't find `Main` in any children of any vnode. And also `Main` has it's own subTree, they will be in the subTree of `Main`, so if we ignore `subTree` from vnode of App(which is vnode of `Main`), we won't find and Component here.
> 
> I tried my demo with `setup` and `jsx`, I will try `render` later.

forget it, it's not that simple...

tests/findComponent.spec.ts
```diff
@@ -37,7 +42,46 @@ const compA = defineComponent({
...
+    }
+    const Parent = {
+      name: 'Parent',
+      render() {
+        return h(Child)
+      }
+    }
+
+    const wrapper = mount(Parent)
+    wrapper.getComponent(Child)
```
AngeloGulina, line 72: I am confused by `getComponent` used in this context (and some following lines as well).

lmiller1990: I think I will abandon this PR for now since @Jokcy found a better solution.

I was just doing this as a quick test - the idea is that if the component is *not* found, it will throw an error (`findComponent` would *not*).

lmiller1990:
It seems someone smarter than me has figured this out https://github.com/vuejs/vue-test-utils-next/pull/188

I will close this for now. I am sure we are likely still missing some edge cases. Anyway, this was very frustrating, but I am glad we are finding all these bugs before Vue 3 is official released. Great job everyone who was involved in this discussion and has contributed so far!

tests/findComponent.spec.ts
AngeloGulina: I see. Just for my understanding, #188 is solving the slots case but not the functional components one?

If this is true, should we pursue your proposal of having some warning or anything else?

lmiller1990: I would say so. This would be a much better developer experience.

We could use this `__isFunctionalComponent` property I am proposing we add here https://github.com/vuejs/vue-test-utils-next/pull/185

What do you think?

AngeloGulina: I like it!

PR #188, opened 2020-08-24
fix findComponent failed for slots

Jokcy:
The main problem is fully described on https://github.com/vuejs/vue-test-utils-next/pull/184.

I try to fix this just add one line code `aggregateChildren(nodes, [node.component.subTree])` at `findAllVnodes` function.
In this test case:
```ts
it('finds all components', () => {
    const compB = defineComponent({
      name: 'compB',
      render: () => h('div', { class: 'B' }, 'TextB')
    })
    const compA = defineComponent({
      name: 'compA',
      template: '<div><slot /></div>'
    })
    const starter = defineComponent({
      name: 'starter',
      template: '<comp-a><comp-b /><comp-b /><comp-b /></comp-a>',
      components: { compA, compB }
    })
    const App = defineComponent({
      name: 'App',
      render: () => h(starter),
      components: { starter }
    })

    const wrapper = mount(App)
    // find by DOM selector
    expect(wrapper.findComponent(compA).exists()).toBe(true)
    expect(wrapper.findComponent(compB).exists()).toBe(true)
    expect(wrapper.findComponent(starter).exists()).toBe(true)
  })
```
`findComponent` will start from `AppVM.$.vnode.subTree`, which is just vnode of `starter`. It won't appear in `node.children` and `node.component.subTree.children` because it is subTree of `App`.

And the slots problem, in this case, children of `starter` vnode is :
```js
{
  default: Function
}
```
as @lmiller1990 said, we can't get vnode from it. 

But I find `component.subTree` of `starter` vnode is vnode of `CompA`, so after we `aggregateChildren` the `starter` vnode the problem solved!


Jokcy:
Actually it's very simple to understand, in this case:
```js
const compA = defineComponent({
      name: 'compA',
      template: '<div><slot /></div>'
    })
    const starter = defineComponent({
      name: 'starter',
      template: '<comp-a><comp-b /><comp-b /><comp-b /></comp-a>',
      components: { compA, compB }
    })
```
`<comp-b /><comp-b /><comp-b />` is children of `comp-a`, but they will be compile to `{ default: () => }` for better performance. If we can't use `children` to get their vnodes, we just git if from children of `div`, which `div` is subTree of `compA`.

lmiller1990:
I'll give this a test and review in the next day or two. Great job! I appreciate your work and contribution.

Jokcy:
> I'll give this a test and review in the next day or two. Great job! I appreciate your work and contribution.

OK, wait for your feedback

lmiller1990:
Oh cool, I get it now. https://github.com/vuejs/vue-test-utils-next/pull/184#issuecomment-678897322 was very useful.

Happy to merge this one - I have to ask, are there any cases we are missing? Or does this let `findComponent` work for everything?

PS: do you mind if I move this test into the `findComponent.spec.ts` file? I like to keep all the related tests in the same (makes it easier to find things).

Jokcy:
@lmiller1990 of course it's better to put cases for same api together, I PR this for quick review and test. If everything is ok now, I can commit to make code style clean.

lmiller1990:
If it's all green it's all good 👍 

It took me a while to get what this was doing:

```
if (matches(node, selector) && matchingNodes.indexOf(node) < 0) {
```

I guess with addition, it's possible we find the same node twice? W could just do `!matchingNodes.includes(node)` (not sure if there is a difference between `indexOf` and `includes`, I guess they are both checking object identity).

I also noticed there is a `uid` on components, we (might) be able to use that to keep track of which node(s) we've already found. But I think this is fine if it works.

Jokcy:
Oh right, `matchingNodes.indexOf(node)` is not needed, I add this because at first I add this code:
```
if (node.component.subTree.component) {
  aggregateChildren(
      nodes,
      node.component.subTree.component.subTree.children,
   )
}
```
this do will add same vnodes multi times. But after remove this code, I guess we don't have to worry about it.

src/utils/find.ts
```diff
@@ -123,11 +124,13 @@ function findAllVNodes(
         aggregateChildren(nodes, fallbackTree.children)
       }
     }
-    if (matches(node, selector)) {
+    if (matches(node, selector) && matchingNodes.indexOf(node) < 0) {
       matchingNodes.push(node)
     }
   }
 
+  // console.log(matchingNodes)
```
AngeloGulina, line 132: Should we maybe get rid of this before merging?

lmiller1990: Yep definitely, good to keep things clean. I think he will clean the code a little before we merge it up.

I will do a release once this is merged!

Jokcy:
> Oh right, `matchingNodes.indexOf(node)` is not needed, I add this because at first I add this code:
> 
> ```
> if (node.component.subTree.component) {
>   aggregateChildren(
>       nodes,
>       node.component.subTree.component.subTree.children,
>    )
> }
> ```
> 
> this do will add same vnodes multi times. But after remove this code, I guess we don't have to worry about it.

Ah, I made a mistake, we still need this code. In the `findAllComponents` case `wrapper.findAllComponents('.C')` will be 4 if we don't check repeat.

Jokcy:
I find the reason. In findAllComponents case:
```
const compC = defineComponent({
  name: 'ComponentC',
  template: '<div class="C">C</div>'
})
const compB = defineComponent({
  template: '<div class="B">TextBefore<comp-c/>TextAfter<comp-c/></div>',
  components: { compC }
})
const compA = defineComponent({
  template: '<div class="A"><comp-b ref="b"/><hello ref="b"/></div>',
  components: { compB, Hello }
})
```
we will go through children of `compA -> div -> comp-b -> div -> comp-c`
and when we meet `comp-b`, we check it's subTree, which is `div -> comp-c`, so same vnode of `comp-c` will add twice.

lmiller1990:
Yes I thought so - with this change we can potentially "find" the same nodes more than once, so we need the check.

The way I was handling this was using an object and the component's `uid`: https://github.com/vuejs/vue-test-utils-next/pull/184/files#diff-217a6dd71dcfdf610f0eb74f47cb27c5R121

That means we don't need to check an entire array. It's slightly more efficient, but I think the way you are doing it works fine as well.

Looks good! I will give this a little test myself and then merge it up. Thanks a lot, this is great first contribution!

lmiller1990 (approved):

Jokcy:
Very happy to contribute, for me I learned a lot about vnode ant how VTU work. Your guys have done great job, hope I can do more to help.

Oh right, I guess if we need to check matchedNodes, we can do some optimize at beginning of `findAllVnodes`, something like:
```
while (nodes.length) {
    const node = nodes.shift()!
    if (matchedNodes.include(node)) continue
```
since we already added it, we definitely have aggregate it's children

tests/findComponent.spec.ts
```diff
@@ -226,4 +226,29 @@ describe('findComponent', () => {
...
+      '<slot-component><comp-b /><comp-b /><comp-b /><comp-c v-slot:b /></slot-component>',
+    components: { slotComponent, compB, compC }
+  })
+  const SlotApp = defineComponent({
+    name: 'SlotApp',
+    template: `<slot-main />`,
+    components: { SlotMain }
+  })
+
+  it('finds components with slots which will be compile to `{ default: () => [children] }`', () => {
```
AngeloGulina, line 247:
```suggestion
  it('finds components with slots which will be compiled to `{ default: () => [children] }`', () => {
```

cexbrayat (approved):
Thank you for your contribution! 

It would be great to add a few comments with the reasoning behind these changes if you can. I suspect we'll all forgot very quickly why we ended there :)

lmiller1990:
Oh @cexbrayat just saw your comment. Agreed on documenting this.

I am going to do a very detailed blog post about this at some point in the near future. It is really difficult to explain without a diagram of how the tree looks. It took many hours of logging for me to understand how the vdom works (and I still don't full get it, haha).

Jokcy:
@cexbrayat It's a great idea, I did not add a comment because is really complex to explain, and another reason is my poor english😭, so I guess a doc from @lmiller1990 will be better.

lmiller1990:
@Jokcy If you write any blog posts about this library (or Vue internals) I can proof read and correct it - you can either email me or find me in Vue land discord.

Jokcy:
@lmiller1990 I really appreciate you said that. I havn't tried to write an english post yet, maybe I should try 🤔.

Issue #194, opened 2020-08-31
Types are wrong when using slots mounting option


lmiller1990:
No idea what's going on.

Some days I wonder if TS is actually making the DX better or worse, lol. 

```
test('layout full page layout', () => {
  const wrapper = mount(Layout, {
    slots: {
      header: '<div>Header</div>',
      main: '<div>Main Content</div>' ,
      footer: {
        template: '<div>Footer</div>'
      }
    }
  })

  expect(wrapper.html()).toContain('Main Content')
})
```

![image](https://user-images.githubusercontent.com/19196536/91722215-4bc93a00-ebdd-11ea-81ce-b5c5d4c6c1ac.png)

calebbergman:
I'm running into an issue with a somewhat similar error message.

This works fine when the Component is declared within the spec file (I just discovered)

```
// example.spec.ts
import { mount} from '@vue/test-utils'

const Layout = {
  template: `
  <div>
    <button v-if="loggedIn" data-test="logoutbtn">Logout</button>
  </div>
  `,
  data: () => ({
    loggedIn: false,
  })
}

describe('Layout.vue', () => {
  it('renders button when logged in', () => {
    const wrapper = mount(Layout, {
      data() {
        return {
          loggedIn: true
        }
      },
    });
    expect(wrapper.findAll('[data-test="logoutbtn"]').length === 1).toBe(true);
  });
});
```

But fails with a `No overload matches this call` error when trying to import the Component.

```
// Layout.vue
<template>
  <div>
    <button v-if="loggedIn" data-test="logoutbtn">Logout</button>
  </div>
</template>

<script lang="ts">
import { defineComponent } from 'vue';

export default defineComponent({
  name: 'Layout',
  data: () => ({
    loggedIn: false,
  }),
});
</script>
```

```
// example.spec.ts
import { mount} from '@vue/test-utils'
import Layout from '@/components/Layout.vue';

describe('Layout.vue', () => {
  it('renders button when logged in', () => {
    const wrapper = mount(Layout, {
      data() {
        return {
          loggedIn: true
        }
      },
    });
    expect(wrapper.findAll('[data-test="logoutbtn"]').length === 1).toBe(true);
  });
});
```

```
Test suite failed to run

    TypeScript diagnostics (customize using `[jest-config].globals.ts-jest.diagnostics` option):
    tests/unit/example.spec.ts:18:27 - error TS2769: No overload matches this call.
      The last overload gave the following error.
[...]

PR #200, opened 2020-09-10
wip: data typing issue

lmiller1990:
This lets you use `data` with SFC imports as described in #194, but breaks the type safety for non SFCs. This seems unavoidable - I don't see how we can have the best of both words here, really.

I don't see a good solution for this in the short term, unless someone has figured a good way to extract type info from `.vue` files (not yet, or at least not stable yet).

Any ideas? Using `vue` files is much more common than using components defined with render functions, but I really value strong typing as well. I wonder if there is some way to not have as string type-checking for `.vue` files (using a shim). Eg: disable mounting option checks for shimmed vue files. I will do some research.

@cexbrayat @pikax would love some input.

### DIFF
--- a/src/vueShims.d.ts
+++ b/src/vueShims.d.ts
@@ -1,5 +1,5 @@
 declare module '*.vue' {
   // TODO: Figure out the typing for this
-  import Vue from 'vue'
-  export default any
+  const component: any
+  export default component
 }
--- a/test-dts/mount.d-test.ts
+++ b/test-dts/mount.d-test.ts
@@ -1,6 +1,8 @@
 import { expectError, expectType } from 'tsd'
 import { defineComponent } from 'vue'
 import { mount } from '../src'
+// @ts-ignore
+import Hello from '../tests/components/Hello.vue'
 
 const AppWithDefine = defineComponent({
   props: {
@@ -18,17 +20,6 @@ expectType<string>(
   mount(AppWithDefine, {
     props: { a: 'Hello', b: 2 }
   }).vm.a
-)
-
-// no data provided
-expectError(
-  mount(AppWithDefine, {
-    data() {
-      return {
-        myVal: 1
-      }
-    }
-  })
 )
 
 // can not receive extra props
@@ -150,6 +141,15 @@ mount(AppWithProps, {
   global: {
     config: {
       isCustomElement: (tag: string) => true
+    }
+  }
+})
+
+// SFC with data option
+mount(Hello, {
+  data() {
+    return {
+      foo: 'bar'
     }
   }
 })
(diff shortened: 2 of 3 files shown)


test-dts/mount.d-test.ts
```diff
@@ -1,6 +1,8 @@
 import { expectError, expectType } from 'tsd'
 import { defineComponent } from 'vue'
 import { mount } from '../src'
+// @ts-ignore
```
cexbrayat, line 4: Is it necessary? TBH I don't like `@ts-ignore` even less in dts tests where we're supposed to check that everything works :) I'd like this to work without it.

src/vueShims.d.ts
```diff
@@ -1,5 +1,5 @@
 declare module '*.vue' {
   // TODO: Figure out the typing for this
-  import Vue from 'vue'
-  export default any
+  const component: any
+  export default component
```
cexbrayat, line 4: unrelated: Maybe we should use the same shim as the one provided by the CLI. It is not cast as `any` I know, but that would help catch errors that developers will encounter.

cexbrayat (commented):
I think @pikax may remember why we have `never` there, as he introduced it in 9af9de2d81095c8336a68786556406cdf21a1c93 to help bump to beta.15. So he'll probably have a better insight.
If not, I'll need to take a deeper look. I'd love to keep the strong type safety if we can (as you suspect ;) )

src/vueShims.d.ts
pikax: I think `Component` is a better candidate, but needs testing

lmiller1990: From what I can tell since the shim gives you no information about your component (eg props/data/whatever) the majority of the types we have are not very useful. in my experience the vast majority of people are using `vue` files.

I think we are focusing too heavily on "correct typing" and not enough on the actual DX. At least from my point of view, I would rather not have type safety than have random ts compile errors even doing the most basic of things (like using `data`). 

I wonder if there is some way to ignore all the mounting types when the imported component is from a `vue` file? That would really be ideal, at least at this point, since you cannot get any useful information from an SFC right now.

pikax: without a typescript plugin that will handle vue files, I doubt we can extract much information from a `.vue` file.

The shim only tells the typescript: "this file import is of this type" - without any process.

lmiller1990: I think we need to merge something (this or something better?) sooner rather that later... using `data` to set a component up for testing is basically the Hello World of VTU and it doesn't work out of the box atm. I do not want to ruin the hard work that has gone into typing VTU, though. 🤔

PR #203, opened 2020-09-10
fix: trim slot when passed a string

### LINKED ISSUE #202: Passing raw HTML to a slot mounting option breaks when starting with a carriage return
Odd error with this:

```
const ComponentWithSlots = {
  template: `
    <div class="scoped">
      <slot name="scoped" v-bind="{ msg }" />
    </div>
  `,
  data() {
    return {
      msg: 'world'
    }
  }
}

test('scoped slots', () => {
  const wrapper = mount(ComponentWithSlots, {
    slots: {
      scoped: `
        <template #scoped="params">
        Hello {{ params.msg }}
        </template>
      `
    }
  })

  expect(wrapper.html()).toContain('Hello world')
})
```

Specifically:

BAD:

```
 scoped: `
        <template #scoped="params">
        Hello {{ params.msg }}
        </template>
      `
```

GOOD:

```
 scoped: `<template #scoped="params">
        Hello {{ params.msg }}
        </template>
      `
```

Probably just need to do `trim`.

lmiller1990:
resolves #202

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -236,7 +236,7 @@ export function mount(
 
         if (typeof slot === 'string') {
           // slot is most probably a scoped slot string or a plain string
-          acc[name] = (props: VNodeProps) => h(processSlot(slot), props)
+          acc[name] = (props: VNodeProps) => h(processSlot(slot.trim()), props)
           return acc
         }
 
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -236,7 +236,7 @@ export function mount(
 
         if (typeof slot === 'string') {
           // slot is most probably a scoped slot string or a plain string
-          acc[name] = (props: VNodeProps) => h(processSlot(slot.trim()), props)
+          acc[name] = (props: VNodeProps) => h(processSlot(slot), props)
           return acc
         }
 
--- a/src/utils/compileSlots.ts
+++ b/src/utils/compileSlots.ts
@@ -1,7 +1,8 @@
 import { compile } from '@vue/compiler-dom'
 import * as vue from 'vue'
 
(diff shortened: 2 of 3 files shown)


src/mount.ts
```diff
@@ -236,7 +236,7 @@ export function mount(
 
         if (typeof slot === 'string') {
           // slot is most probably a scoped slot string or a plain string
-          acc[name] = (props: VNodeProps) => h(processSlot(slot), props)
+          acc[name] = (props: VNodeProps) => h(processSlot(slot.trim()), props)
```
cexbrayat, line 239: nit: maybe `trim` should be called in `processSlot` (as if someone uses it in an other place of VTU, we don't have the same issue another time)

cexbrayat (approved):
I'm curious why this is needed. Does it mean that `compile` from core has an issue?

Otherwise, LGTM

src/mount.ts
lmiller1990: @cexbrayat I am curious too... I had a little look but it wasn't obvious why this was causing and issue. I guess no one has tried calling `compile` where the first character is a carriage return 🤔 

I wonder if this should be fixed in core.

lmiller1990: Ohh I bet this is the issue https://github.com/vuejs/vue-test-utils-next/blob/e92bdcb430b6098204d20c5ea3b9c80bd21ddba7/src/utils/compileSlots.ts#L5

`startsWith('<template')` is not the same as `startsWith('\ntemplate')`.

lmiller1990: I made your suggested change! thanks.

Issue #204, opened 2020-09-13
Support propsData


lmiller1990:
I am moving a small code base to VTU v2. One of the main breaking changes is the change from `propsData` -> `props` in the mounting options.

Eg:

```
// VTU v1

const A = {
  props: ['msg']
}

mount(A, {
  propsData: { msg: 'msg' }
})

// VTU v2

mount(A, {
  props: { msg: 'msg' }
})
```

I'd like to allow `propsData` to work as well. I'd suggest we support both for migration purposes. We won't document `propsData`, since using `props` matches the `props` key in Vue components, but we will keep both for convenience. 

The acceptance criteria for this PR are as follows:

- support the `propsData` mounting method
- pass whatever is received from `propsData` to `props` in `src/mount.ts`
- add at least one tests with `propsData`

PR #205, opened 2020-09-13
feat(mount): add support for propsData

WebDevEtc:
- Adds support for mounting with `propsData` option,
   for backwards compatibility.

See https://github.com/vuejs/vue-test-utils-next/issues/204

Added two tests to be clear what happens when both `props` and `propsData` are provided. Happy to change implementation to a different behaviour if that would be preferred. 

(No warning/error is given to the user if both `props` and `propsData` are used in the mounting options.)

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -47,6 +47,8 @@ interface MountingOptions<Props, Data = {}> {
     ? Partial<Data>
     : never
   props?: Props
+  /** @deprecated */
+  propsData?: Props
   attrs?: Record<string, unknown>
   slots?: SlotDictionary & {
     default?: Slot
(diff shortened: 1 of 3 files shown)


src/mount.ts
```diff
@@ -47,6 +47,7 @@ interface MountingOptions<Props, Data = {}> {
     ? Partial<Data>
     : never
   props?: Props
+  propsData?: Props
```
lmiller1990, line 50: Looks good, what do you think about marking this as deprecated?

https://devblogs.microsoft.com/typescript/announcing-typescript-4-0-beta/#deprecated-support

lmiller1990 (commented):
One comment! Thanks for this.

lmiller1990 (approved):

lmiller1990:
Great, thanks @WebDevEtc 

I will release in the next few days.

PR #206, opened 2020-09-16
fix: findComponent in Suspense for rc.11

cexbrayat:
https://github.com/vuejs/vue-next/pull/2099 introduced a big refactoring in how Suspense works, which broke VTU.
`SuspenseBoundary` no longer exposes the same interface, so our current implementation was not compiling anymore (and throwing at runtime).
This solves the issue and simplifies the current code, as the newly introduced `activeBranch` field now contains the... active branch of the Suspense, so we just have to check that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "@types/estree": "^0.0.42",
     "@types/jest": "^24.9.1",
     "@types/node": "12.12.35",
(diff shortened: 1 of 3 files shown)


lmiller1990 (approved):
Great catch. I will release a new version soon.

PR #207, opened 2020-09-17
chore: rc 12

lmiller1990:


package.json
```diff
@@ -38,7 +38,7 @@
     "ts-jest": "^25.0.0",
     "tsd": "0.11.0",
     "typescript": "^3.7.5",
-    "vue": "^3.0.0-rc.11",
+    "vue": "^3.0.0-rc.12",
```
cexbrayat, line 41: nit: I think it would be better to also update the compiler-sfc version above

cexbrayat (approved):

package.json
lmiller1990: oops, yes.

lmiller1990: done!

PR #212, opened 2020-09-25
Add isVisible and stub transition/transition-group by default

### LINKED ISSUE #210: Testing for visibility with transition (v-show)
Hi, I was trying to test that an element is not visible after triggering an event, but am having trouble with getting the test to pass when the element is wrapped in a transition. I wasn't able to find any related issues, apart from the fact that `isVisible` is no longer supported. I was hoping someone would be able to point me in the right direction

Consider the following structure where `state.showItem` is toggled between true/false on mouse enter/leave in a different element:

```
<transition name="fade">
    <span class="item" v-show="state.showItem">
      {{content}}
    </span>
</transition>
```

The following test fails on the second assertion:

```
await wrapper.find(".trigger-item").trigger("mouseenter");
expect(
  wrapper.find(".item").attributes("style").includes("display: none")
).toBe(false);

await wrapper.find(".trigger-item").trigger("mouseleave");
expect(
  wrapper.find(".item").attributes("style").includes("display: none")
).toBe(true);
```

But succeeds once I've removed the wrapping `transition` component. I have tried using `await nextTick` to no avail.

Perhaps I am doing something wrong?

Any advice would be appreciated.

Thanks

lmiller1990:
resolves #210 

- Add `isVisible` back. We decided to un-deprecate it in V1, porting it to V2.
- We stub transition and transition group by default in V1. We will do the same here for consistency.
- This solves the problem in #210 (see there for more context)
- Stubbing transitions avoids the problem where transitions take time (eg 0.25s) and your assertion runs before your transition has finished. In unit tests you don't really care about things like CSS transitions; you just care if things are visible or not via `v-show`.
- You can opt out on a test-by-test basis by passing `transition: false` and `transition-group: false` to `global.stubs`.
- you can opt out globally by doing `config.global.stubs = { transition: false }`. This is useful for libraries like [cypress-vue](https://github.com/bahmutov/cypress-vue-unit-test) who do not stub things, since they are not constrained by things like `nextTick` like when testing in jsdom with jest.

### DIFF
--- a/src/utils/isElementVisible.ts
+++ b/src/utils/isElementVisible.ts
@@ -1,0 +1,33 @@
+/*!
+ * isElementVisible
+ * Adapted from https://github.com/testing-library/jest-dom
+ * Licensed under the MIT License.
+ */
+
+type StylableElement = HTMLElement | SVGElement
+
+function isStyleVisible<T extends StylableElement>(element: T) {
+  const { display, visibility, opacity } = element.style
+  return (
+    display !== 'none' &&
+    visibility !== 'hidden' &&
+    visibility !== 'collapse' &&
+    opacity !== '0'
+  )
+}
+
+function isAttributeVisible<T extends StylableElement>(element: T) {
+  return (
+    !element.hasAttribute('hidden') &&
+    (element.nodeName === 'DETAILS' ? element.hasAttribute('open') : true)
+  )
+}
+
+export function isElementVisible<T extends StylableElement>(element: T) {
+  return (
+    element.nodeName !== '#comment' &&
+    isStyleVisible(element) &&
+    isAttributeVisible(element) &&
+    (!element.parentElement || isElementVisible(element.parentElement))
+  )
+}
--- a/src/utils/isElementVisible.ts
+++ b/src/utils/isElementVisible.ts
@@ -4,10 +4,13 @@
  * Licensed under the MIT License.
  */
 
-type StylableElement = HTMLElement | SVGElement
+function isStyleVisible<T extends Element>(element: T) {
+  if (!(element instanceof HTMLElement) && !(element instanceof SVGElement)) {
+    return false
+  }
 
-function isStyleVisible<T extends StylableElement>(element: T) {
   const { display, visibility, opacity } = element.style
+
   return (
     display !== 'none' &&
     visibility !== 'hidden' &&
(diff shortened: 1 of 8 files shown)


tests/isVisible.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+  const Comp = {
+    template: `<div><span v-show="show" /></div>`,
+    props: {
+      show: {
+        type: Boolean
+      }
+    }
+  }
+
+  it('returns false when element hidden via v-show', () => {
```
xanf, line 13: *suggestion*: WDYT of adding one more pair of tests `returns false when ancestor is hidden via v-show` to capture our intent - we're detecting visibility not only on entire element itself

lmiller1990: Sure, good idea.

src/utils/isElementVisible.ts
```diff
@@ -0,0 +1,36 @@
...
+    (element.nodeName === 'DETAILS' ? element.hasAttribute('open') : true)
+  )
+}
+
+export function isElementVisible<T extends Element>(element: T) {
+  return (
+    element.nodeName !== '#comment' &&
+    isStyleVisible(element) &&
+    isAttributeVisible(element) &&
+    (!element.parentElement || isElementVisible(element.parentElement))
```
cexbrayat, line 34: I guess this was battle-tested in VTU 1, but just in case here is how jQuery tests the visibility https://github.com/jquery/jquery/blob/d0ce00cdfa680f1f0c38460bc51ea14079ae8b07/src/css/hiddenVisibleSelectors.js (we implement the same logic in an Angular testing library).

cexbrayat (approved):
LGTM, nice work @lmiller1990 !

src/utils/isElementVisible.ts
lmiller1990: I just coped this from another library (same as V1).

There is even more robust visible checks out there; since the primary use case is `v-show`, which does this via `display: none`, I think this should cover the main use cases. We could make it more robust if someone has a problem. Surprisingly how had a true visibility selector can be to implement...

afontcu: https://github.com/testing-library/jest-dom/blob/master/src/to-be-visible.js#L25

tests/isVisible.spec.ts
lmiller1990: done!

PR #214, opened 2020-09-26
fix: allow finding root component

### LINKED ISSUE #211: findComponent doesn't work as expected after upgrade to 2.0.0-beta.5 for use with Vue 3
### Subject of the issue

Previously passing tests now fail. 

### Steps to reproduce

The following test setup: 
```
    const wrapper = mount(Conversation, {global: {plugins: [store], mocks: {$route}, mixins: [mixins]},  propsData: {ticket_id: 1}})
    await flushPromises()
    let conversation = wrapper.findComponent(Conversation)
    let textInput = conversation.find('textarea')
```
results in an error ```Cannot call find on an empty VueWrapper.``` even though the component was just mounted. This occurs even with the following ```Conversation.vue```: 
```
<template>
  <div class="page conversation">
    
  </div>
</template>
```

This feature works as expected post-migration to vue3, but the tests fail on this. 

### Expected behaviour

That findComponent should find the mounted component. 

### Actual behaviour

findComponent returns an empty VueWrapper.

### Possible Solution

I've tried simplifying the components, but the problem still persists.

lmiller1990:
See: https://github.com/vuejs/vue-test-utils-next/issues/211 (resolves #211)

This seems like a kind of strange thing to have, but I would like to keep back-compat as much as possible. Apparently this worked in V1 (no idea what the use case is).

I could not get this to work without giving a `name` to the component; so it's not perfect in terms of back-compat.

### DIFF
--- a/src/utils/find.ts
+++ b/src/utils/find.ts
@@ -13,7 +13,10 @@ import { matchName } from './matchName'
  * @param selector
  * @return {boolean | ((value: any) => boolean)}
  */
-function matches(node: VNode, selector: FindAllComponentsSelector): boolean {
+export function matches(
+  node: VNode,
+  selector: FindAllComponentsSelector
+): boolean {
   // do not return none Vue components
   if (!node.component) return false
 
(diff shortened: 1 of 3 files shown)


cexbrayat (approved):
Even if the name is needed, I guess this is good enough for backward comp (as it does not seem to be super useful).

afontcu (approved):

PR #220, opened 2020-10-12
fix: set default event.cancelable from dom-event-types

### LINKED ISSUE #218: @click.prevent not setting event.defaultPrevented in jest tests
### Version
3.0.0

### Reproduction link
[https://github.com/aethr/vue-prevent-test](https://github.com/aethr/vue-prevent-test)

### Steps to reproduce
Create a link with an `a` tag with a `@click.prevent='clickHandler'` event handler like:
```
  methods: {
    clickHandler(event) {
      console.log('event.defaultPrevented', event.defaultPrevented)
    }
  },
```

Click the link in the browser, console will show:
```
event.defaultPrevented true
```

Create a jest test that will trigger a click on the link like:
```
  it('does not try to follow a link', async () => {
    const wrapper = shallowMount(HelloWorld);
    wrapper.get('#link').trigger('click')
  })
```

Run jest test that trigger a click on the same element with `npm run test:unit`, console will show:
```
  console.log src/components/HelloWorld.vue:6
    event.defaultPrevented false
```

### What is expected?
Using `vue-test-utils` to trigger a click event on a link should result in the same behaviour as we see in the browser, with `event.defaultPrevented` set properly.

### What is actually happening?
Although the `prevent` modifier handler *is* being called, somehow the event itself is not being modified so that `defaultPrevented` is true.

---
My test case also includes an example of a basic jest test that sets up a similar example using pure jsdom, and we can see the correct `event.defaultPrevented` value is set, so I don't think this is an issue with jsdom.

In the app where we first saw this behaviour, we're seeing an error during one of our tests where jsdom thinks navigation is being attempted because an <a> element with `@click.prevent` is causing navigation (but only in jsdom, not in the browser). We think it is probably related to this bug.

aethr:
This resolves #218 which may be caused by a simple typo in the codebase.

It seems like a core intent of `getEventProperties` is to source default properties of different types of DOM events from the 'dom-event-types' dep and provide them as properties for new events. However, it looks like the `cancelable` property is not being added to the returned object.

This change resolves the bug I'm seeing in #218 by allowing `click` events to have `preventDefault` work correctly.

I'd be happy to add a test case to test against regression here, but it doesn't seem like a logic bug or something that's likely to be reverted if my above assumptions seem reasonable.

### DIFF
--- a/src/createDomEvent.ts
+++ b/src/createDomEvent.ts
@@ -112,7 +112,7 @@ function getEventProperties(eventParams: EventParams) {
     ...systemModifiersMeta, // shiftKey, metaKey etc
     ...options, // What the user passed in as the second argument to #trigger
     bubbles: meta.bubbles,
-    meta: meta.cancelable,
+    cancelable: meta.cancelable,
     // Any derived options should go here
     keyCode,
     code: keyCode,


afontcu (approved):
Nice catch!

lmiller1990 (approved):

PR #224, opened 2020-10-17
fix: support for DefineComponent shim

pikax:
Support for the DefineComponent type 

There's an issue with the passing the data function:
```ts
// TODO it should work
 mount(ShimedComponent, {
   data() {
     return {
       a: 1
     }
   }
})
```

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -17,7 +17,12 @@ import {
   ComponentPropsOptions,
   AppConfig,
   VNodeProps,
-  ComponentOptionsMixin
+  ComponentOptionsMixin,
+  DefineComponent,
+  MethodOptions,
+  AllowedComponentProps,
+  ComponentCustomProps,
+  ExtractDefaultPropTypes
 } from 'vue'
 
 import { config } from './config'
@@ -34,6 +39,8 @@ import {
 } from './constants'
 import { stubComponents } from './stubs'
 
+type PublicProps = VNodeProps & AllowedComponentProps & ComponentCustomProps
+
 type Slot = VNode | string | { render: Function } | Function | Component
 
 type SlotDictionary = {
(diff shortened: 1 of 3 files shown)


cexbrayat:
@pikax I gave it a try in one of my projects and it's a bit better indeed (no more issue on mount 👍).

But I still have issues with the type of `vm`.

For example, if your component looks like:
```ts
  name: 'HelloWorld',
  props: {
    msg: String,
  },
  setup() {
    return { user: reactive({ name: 'Cédric' }) };
  }
});
```

then mounting this component in a test succeeds, but `vm` does not expose `user`:

```ts
describe('HelloWorld.vue', () => {
  it('renders props.msg when passed', () => {
    const msg = 'new message'
    const wrapper = shallowMount(HelloWorld, {
      props: { msg }
    })
    expect(wrapper.text()).toMatch(msg)
    expect(wrapper.vm.user.name).toBe('Cédric');
  })
})
```

This last line does not compile with Vue 3.0.1 and this patch of VTU: 
```
tests/unit/example.spec.ts:11:28 - error TS2339: Property 'name' does not exist on type 'never'.
    11     expect(wrapper.vm.user.name).toBe('Cédric');
```

But it does compile with Vue v3.0.0. So maybe this is something to fix in vue-next... WDYT?

pikax:
Can you provide a full example?


It works on my side
image

lmiller1990:
Good to merge this? Seems this should go out asap - or do we want to wait on @cexbrayat to confirm his issue with data?

cexbrayat:
@pikax this is exactly this use case, but using a SFC 

See https://github.com/cexbrayat/vue-301-shims/commit/819bab0a9102cb12fa16ed92407ce3a92c426efe 
Even with your fix, there is still an issue with `wrapper.vm`

Issue #219, opened 2020-10-09
VTU-next renders additional markup


afontcu:
Hi folks! 👋 

When trying to port Vue Testing Lib to Vue 3 (and thus vtu-next), I've encountered several issues regarding additional markup rendered by vtu-next.

Some context: Vue Testing Lib provides `container` and `baseElement` [keys](https://testing-library.com/docs/vue-testing-library/api#container), to help testing scenarios where a div is not a valid DOM wrapper element (imagine a table with rows and columns). I first attempted to leverage attachTo ([source](https://github.com/testing-library/vue-testing-library/pull/137/files#diff-2cfdd145f61e8456ea199f8af800c145R66)), but couldn't prevent snapshots from breaking. I'm not that concerned about snapshots _per se_, but rather the implications of wrapping every rendered component with an additional div with custom attributes:

![imatge](https://user-images.githubusercontent.com/9197791/95600806-f04b6100-0a52-11eb-9669-493dac118b98.png)

I'm not entirely sure we need to mount a div with id="app" – looks like one of the first iterations we did when started working with vtu-next, but I'd say it could go now: https://github.com/vuejs/vue-test-utils-next/blob/master/src/mount.ts#L196-L197

Could we remove the additional markup? This markup will likely impact on any snapshot-based testing done with VTU, and also tools that traverse (poorly, I grant you that 😉) the DOM.

Thanks!

lmiller1990:
Hm, Vue needs *something* to mount on, right? Does testing-library use `attachTo` by default?

I agree we probably do not need the `id`.

afontcu:
> Hm, Vue needs _something_ to mount on, right? Does testing-library use `attachTo` by default?

I've tried several combinations of attaching to the container, the baseElement, or a totally unrelated element, without success. Until now (Vue 2), we attach to the `container`, which makes total sense.

Now, if I do something like the following, tests pass:

```js
const wrapper = mount(Component, { attachTo: container }) 
// here, container is the same thing it was in Vue Testing Lib for Vue 2

function unwrap(node) {
  node.replaceWith(...node.childNodes)
}

unwrap(wrapper.parentElement) // this removes the additional "data-v-app" div node
```

but still feels a bit hacky?

lmiller1990:
One difference is that Vue 2 did not have top level fragments; Vue 3 does. I think this is why we have the top level node (maybe).

lmiller1990:
[...]

PR #225, opened 2020-10-18
refactor: do not pollute markup with useless id

lmiller1990:
Related: https://github.com/vuejs/vue-test-utils-next/issues/219

@afontcu seems we can remove this id. I am not sure about the div - what should we be attaching to?

Also, does testing library use `attachTo` by default? We need *something* to mount on - if you are using attachTo, however, maybe we can use that element to mount on, instead of creating our own DIV here?

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,3 +1,2 @@
-export const MOUNT_ELEMENT_ID = 'app'
 export const MOUNT_COMPONENT_REF = 'VTU_COMPONENT'
 export const MOUNT_PARENT_NAME = 'VTU_ROOT'
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -27,11 +27,7 @@ import { processSlot } from './utils/compileSlots'
 import { createWrapper, VueWrapper } from './vueWrapper'
 import { attachEmitListener } from './emitMixin'
 import { createDataMixin } from './dataMixin'
-import {
-  MOUNT_COMPONENT_REF,
-  MOUNT_ELEMENT_ID,
-  MOUNT_PARENT_NAME
-} from './constants'
+import { MOUNT_COMPONENT_REF, MOUNT_PARENT_NAME } from './constants'
 import { stubComponents } from './stubs'
 
 type Slot = VNode | string | { render: Function } | Function | Component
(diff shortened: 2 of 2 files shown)


afontcu:
Hi!

This is the solution I came up with, which makes all related-tests pass at once: https://github.com/testing-library/vue-testing-library/pull/137/files#diff-4f8b5837e5c68e71852f1c3f41f2e98201426d45a94bd7767cddfc98ccd02fc5R66-R68

re `attachTo`, I'm currently attaching it to `container`, just as testing lib for Vue 2 (no changes there): https://github.com/testing-library/vue-testing-library/pull/137/files#diff-4f8b5837e5c68e71852f1c3f41f2e98201426d45a94bd7767cddfc98ccd02fc5R57

> if you are using attachTo, however, maybe we can use that element to mount on, instead of creating our own DIV here?

This is probably… sensible? I mean, people asked for special behavior when `document.body` is provided to `attachTo`. What's the point of letting people set a custom `attachTo` value, and then wrap it up again in a div? 🤔

lmiller1990:
I could be wrong about how attachTo is currently working, that might already be the behavior. I will check.

Either way this fix is still good. Seems like your fix in testing library is okay, is there anything else blocking the Vue 3 integration?

afontcu:
> is there anything else blocking the Vue 3 integration?

Not really, I'm a bit worried about the third `render` parameter – a way to access the Vue instance of the component under testing, and how this translates to Vue 3 concept of "app" instead of vue instances:

```js
// testing lib with 2.x
render(Component, {}, vue => vue.use(AnAwesomePlugin))
```

`options.global` works in a large chunk of situations (this is going to be quite a big breaking change, and hard to suggest a clear solution or even a codemod 😞), however some people did quite complex trickery with the Vue instance. 

Also, integration with VueRouter-next is not there yet, given that we haven't decided how to tackle it in VTU-next (AFAICT). I might go ahead and simply install it as a plugin, and see what happens in the wild.

I guess I'll release an alpha version in the next channel soon, and see if people run into some issues. This might help dogfooding VTU-next, too! :)

lmiller1990:
`localVue` changes should mostly be fine with just `global.plugins`. The `localVue` change is inevitably a big breaking change for everyone; we can't really avoid this, it's just how Vue 3 works now.

And yep - my experience so far is predicting things is hard, better just to get something out there and let people report issues.

PR #232, opened 2020-10-27
Issue 158 functional emits

### LINKED ISSUE #158: emitted cannot capture events for for functional components
hi

I found no way to test emit today when Adding tests to FunctionalComponents
```js
it.only('functional emit', () => {
    const Component = (props, ctx) => {
      return h('div', [
        h('button', {
          onClick() {
            console.log("click")
            return ctx.emit('hello', 'foo', 'bar')
          }
        })
      ])
    }
    expect(wrapper.emitted()).toEqual({})
    expect(wrapper.emitted().hello).toEqual(undefined)

    wrapper.find('button').trigger('click')
    // Here is a failure
    expect(wrapper.emitted().hello[0]).toEqual(['foo', 'bar'])

    wrapper.find('button').trigger('click')
    expect(wrapper.emitted().hello[1]).toEqual(['foo', 'bar'])
  })
```

The previous processing of the emit was done a layer of packaging in Mixin's hook function beforeCreate

But if it is functional Components it will not trigger beforeCreate

The logical implementation in vue-next is that if it is STATEFUL_COMPONENT, it will call applyOptions, and the beforeCreate that calls mixin is at that time
```js
// vue-next/runtime-cor/renderer.js

 const setupResult = isStateful
    ? setupStatefulComponent(instance, isSSR)
    : undefined

setupStatefulComponent() -> finishComponentSetup() -> applyOptions()
```

But applyOptions is not called if you are functional Components

So the beforeCreate hook for mixin will not be triggered

vue-next does not give us any hooks for Functional Components before calling renderComponentRoot

```js
setupComponent(instance) -> setupRenderEffect() -> renderComponentRoot()
```

Therefore, there is no way for us to wrap one layer for the emit

I think we should give the vue-next a PR

It is recommended to add a lifecycle hook before calling Render

such as "createdComponentInstance"

I'd like your advice

lmiller1990:
resolves #158 

Also found a problem with types, see `ts-ignore`. We don't have any types for functional components. Man, Vue component definitions are complex, and there are many.

I also bumped a bunch of dependencies.


src/mount.ts
```diff
@@ -68,11 +68,18 @@ export type ObjectEmitsOptions = Record<
...
->(
-  originalComponent: TestedComponent,
+// export function mount<
+//   TestedComponent extends FunctionalComponent<Props, E>,
+//   Props,
+//   E extends EmitsOptions = {}
+// >(
+//   originalComponent: TestedComponent,
+//   options?: MountingOptions<Props>
+// ): VueWrapper<ComponentPublicInstance<Props>>
```
cexbrayat, line 78: Is this commented code still here on purpose? Should it be removed or uncommented? Maybe add an explanation if we want to keep it around.

cexbrayat (approved):

src/mount.ts
pikax: I forgot to remove it

lmiller1990: I will remove it

PR #235, opened 2020-11-04
Add ESM browser build

zigomir:
Hi!

I'm trying to test `vue 3` using `vite` with `vue-test-utils-next`. There's a missing build for `esm-browser` because `esm-bundler` uses `@vue/compiler-dom` and not `@vue/compiler-dom/dist/compiler-dom.esm-browser` which is needed if you want to run VTU in browser using `vite`.

This PR is a showcase of how this fourth – `vue-test-utils.esm-browser.js` – build could look like.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -21,11 +21,13 @@ function createEntry(options) {
     isBrowser
   } = options
 
+  const isEsmBrowser = format === 'es' && isBrowser
+
   const config = {
     input,
     external: [
       'vue',
-      '@vue/compiler-dom',
+      isEsmBrowser ? '@vue/compiler-dom/dist/compiler-dom.esm-browser'  : '@vue/compiler-dom',
     ],
     plugins: [
       replace({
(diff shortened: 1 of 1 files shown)


lmiller1990 (approved):
I thought I had this working in Vite already 🤔 the esm-bundler build wouldn't work for you?

zigomir:
No, because `esm-bundler` directly imports `@vue/compiler-dom` which I think resolves to `compiler-dom.esm-bundler.js` which isn't browser compatible.

It would be nice for `vite` to just work, yeah. But I don't think this is possible today, as even `@vue/compiler-dom` for [example](https://github.com/vuejs/vue-next/blob/master/packages/compiler-dom/package.json) doesn't directly reference `compiler-dom.esm-browser.js`. But it builds it anyway because it's listed [here](https://github.com/vuejs/vue-next/blob/master/packages/compiler-dom/package.json#L19).

tl;dr: I don't know.

This is my [POC](https://github.com/zigomir/vite-test) on how one could run VTU next in browser.

lmiller1990:
Let's go with this for now! I will explore this more at some point. I can release this sometime next week (want to fix a few bugs).

thanks!

Issue #187, opened 2020-08-23
Functional component test fails at findComponent


cyberalien:
Simple test with functional component:

```js
import { h } from 'vue';
import { mount } from '@vue/test-utils';

function Test(props, context) {
	return h('div', {
		...props,
		innerHTML: 'This is a test',
	});
}

describe('Mounting component', () => {
	test('with wrapper', () => {
		const Wrapper = {
			components: { Test },
			template: `<Test />`,
		};

		const wrapper = mount(Wrapper, {});
		const item = wrapper.findComponent(Test);
		expect(item.exists()).toBe(true);
	});

	test('without wrapper', () => {
		const wrapper = mount(Test, {});
		const item = wrapper.findComponent(Test);
		expect(item.exists()).toBe(true);
		expect(item.html()).toStrictEqual('<div>This is a test</div>');
	});
});
```
Both tests fail at `findComponent()` that returns null.

vue: 3.0.0-rc.7
vue/test-utils: 2.0.0-beta.2, also tried with latest build from this repo

lmiller1990:
After merging this https://github.com/vuejs/vue-test-utils-next/pull/188/ I will investigate this a little. I learned a lot in #184.

I believe this was not working in VTU v1 either. I expect some limitations to by applied to a found functional component (currently we cannot detect emitted events from them, for example).

In the meantime, I will be adding some warnings to let people know this is a known issue, such as here: https://github.com/vuejs/vue-test-utils-next/pull/185

cyberalien:
Thanks!

As for VTU v1, it is working with similar functional component for Vue 2.

lmiller1990:
Fixed in #238 

This will be in the next release

PR #238, opened 2020-11-08
fix: findComponent with functional components

lmiller1990:
#187 

That was easy 🤔 ... too easy ... must have broke something? Heh.

### DIFF
--- a/src/utils/find.ts
+++ b/src/utils/find.ts
@@ -20,6 +20,10 @@ export function matches(
   // do not return none Vue components
   if (!node.component) return false
 
+  if (node.type === selector) {
+    return true
+  }
+
   if (typeof selector === 'string') {
     return node.el?.matches?.(selector)
   }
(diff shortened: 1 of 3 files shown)


src/utils/find.ts
```diff
@@ -20,6 +20,10 @@ export function matches(
   // do not return none Vue components
   if (!node.component) return false
 
+  if (node.type === selector) {
+    return true
+  }
+
```
cexbrayat, line 26: Ain't that exactly what's done below, but maybe prevented by the condition?
I think it would be better to just extend the condition so it works with functional component (with some comments if possible)

lmiller1990: Good idea, I will revisit this logic. The entire function relies on a lot of side effects and mutation (mutating the  `vnodes` array). I think we should consider rewriting it in a more immutable manner potentially 🤔

PR #240, opened 2020-11-10
Basic support for class component

lmiller1990:
We assumed if `typeof component === 'function'` is is a functional component. Not true - class components are also functions.

This fixes that. We likely need a lot more test for class component. This is a good start.

There is no docs for class component in Vue 3 (which is v8), I am just going by these: https://github.com/vuejs/vue-class-component/blob/next/test/specs/test.spec.ts to figure out what it does.


afontcu (approved):
This is great because you've added semantics to a lone if statement 👌

Stijn98s:
I have the same problem as https://stackoverflow.com/questions/64436017/vue-typescript-component-jest-testing-doesnt-render-values-in-html. Will this fix this problem with vue class components?

lmiller1990:
Should do! If you could add a test or help me test it out, that'd be great!

Stijn98s:
@lmiller1990 i will create a MR that adds the test

lmiller1990:
Thanks @Stijn98s

PR #252, opened 2020-11-20
Export MountingOptions interface

afontcu:
Having `MountingOptions` available in Vue Testing Library will improve greatly the DX of people using the library.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,5 @@
-import { mount, shallowMount, MountingOptions } from './mount'
+import { mount, shallowMount } from './mount'
+import { MountingOptions } from './types'
 import { RouterLinkStub } from './components/RouterLinkStub'
 import { VueWrapper } from './vueWrapper'
 import { DOMWrapper } from './domWrapper'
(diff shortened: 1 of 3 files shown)


dobromir-hristov (approved):

pikax (approved):

cexbrayat (approved):
👍 If it's now public, we could even add documentation on each option (they would show up in IDE)

afontcu:
> 👍 If it's now public, we could even add documentation on each option (they would show up in IDE)

Sure, that'd be great!

However, it is not related to making the interface public, right? As far as I know, people using VTU is already "seeing" the outputs of the interface.

cexbrayat:
Right, but it's a good rule of thumb that exported types should be documented. I'm fine merging it like this though, it was just nitpicking :)

afontcu:
no prob! I'll update this with some documentation :)

afontcu:
Done! Decided to move type into `types.ts` file because it grew a lot.

cexbrayat:
@afontcu 👏 thank you for taking the time to do this! LGTM

afontcu:
totally worth it! It also helped me identify several missing keys in our docs. Working on it now.

lmiller1990 (approved):
true beauty is documenting each type

scottbedard:
@afontcu Thank you for this! This makes building custom `mount` functions much nicer!

PR #253, opened 2020-11-20
Add newer node versions to CI

afontcu:
imatge
(source: https://nodejs.org/en/about/releases/)

We could drop node 10 as it is quite old and in maintenance mode, but it's fine I guess.


lmiller1990:
🎉

PR #254, opened 2020-11-21
fix(types): Fix mount and shallowMount types

### LINKED ISSUE #251: Vue tsx component type definition 'any' in typescript project issue when unit-testing.
**Problem**: I create vue 3 TSX project with default selection from vue cli, typescript, jest. When testing my component, it not show type define on my component in vs code intellisense. It show up as `any`. It not right since I am use typescript.

This issue just recently like 1 week or 2. Like 2 week ago, I create vue 3 typescript with jest enable. And it work and show up type define component on my IDE then. Now I just create new project vue 3. it not show type define on my component on my IDE. 

Imgur

**Reproducible Case**
- vs code IDE
- Create new vue 3 project with vue cli 
- select typescript, jest

Sample code with typescript, jest enable. create App.tsx and in example.spec.ts file with follow code:
```js
// App.tsx

  name: 'App',
  data() {
    return {
      timer: null
    }
  },
  methods: {
    handleSendMessage() {
      console.log('handleSendMessage()')
    }
  },
  render() {
    return (
      <div>
        App component in tsx
      </div>
    )
  }
})
// example.spec.ts
import { shallowMount } from '@vue/test-utils'
import App from '@/App' // same problem if use relative import like ../../scr/App

describe('App.vue', () => {
    it('testing', () => {
      const wrapper = shallowMount(App)
      wrapper.vm.  // <--- after . my IDE would show all property on this component like all methods, data, etc. It do not show up here.
    })
})
```

Side note: it did work before on vue 3 project. I also test on vue 2 tsx project. it work fine there.

pikax:
fix #251

This was caused because `vue-class-component` dependency was not made optional or as a dev, so the import `VueContructor` would always be `any`, making all the `mount` and `shallowMount` to be caught by that overload.

I've removed the dependency on `VueContructor` on the `mount`, made it a bit more generic, this might break if the types changes on the `vue-class-component` (which I think is unlikely) 

@ktsn can you just validate if this overload is enough?
```ts
// Class component - no props
export function mount<V>(
  originalComponent: {
    new (...args: any[]): V
    registerHooks(keys: string[]): void
  },
  options?: MountingOptions<any>
): VueWrapper<ComponentPublicInstance<V>>

// Class component - props
export function mount<V, P>(
  originalComponent: {
    new (...args: any[]): V
    props(Props: P): any
    registerHooks(keys: string[]): void
  },
  options?: MountingOptions<P>
): VueWrapper<ComponentPublicInstance<V>>

```

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -38,7 +38,6 @@ import { attachEmitListener } from './emitMixin'
 import { createDataMixin } from './dataMixin'
 import { MOUNT_COMPONENT_REF, MOUNT_PARENT_NAME } from './constants'
 import { stubComponents } from './stubs'
-import { VueConstructor } from 'vue-class-component'
 
 // NOTE this should come from `vue`
 type PublicProps = VNodeProps & AllowedComponentProps & ComponentCustomProps
@@ -74,10 +73,10 @@ export type ObjectEmitsOptions = Record<
 export type EmitsOptions = ObjectEmitsOptions | string[]
 
 // Class component
-export function mount(
-  originalComponent: VueConstructor,
+export function mount<V>(
+  originalComponent: new (...args: any[]) => V,
   options?: MountingOptions<any>
-): VueWrapper<ComponentPublicInstance<any>>
+): VueWrapper<ComponentPublicInstance<V>>
 
 // Functional component with emits
 export function mount<Props, E extends EmitsOptions = {}>(
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -71,11 +71,21 @@ export type ObjectEmitsOptions = Record<
 >
 export type EmitsOptions = ObjectEmitsOptions | string[]
 
-// Class component
+// Class component - no props
+export function mount<V>(
+  originalComponent: {
+    new (...args: any[]): V
+    registerHooks(keys: string[]): void
+  },
+  options?: MountingOptions<any>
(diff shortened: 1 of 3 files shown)


src/mount.ts
```diff
@@ -73,11 +71,24 @@ export type ObjectEmitsOptions = Record<
...
+
+// Class component - props
+export function mount<V, P>(
+  originalComponent: {
+    new (...args: any[]): V
+    props(Props: P): any
+    registerHooks(keys: string[]): void
+  },
+  options?: MountingOptions<P>
+): VueWrapper<ComponentPublicInstance<V>>
```
ktsn, line 91:
```suggestion
export function mount<V extends ComponentPublicInstance<P>, P>(
  originalComponent: {
    new (...args: any[]): V
  },
  options?: MountingOptions<P>
): VueWrapper<V>
```

As class component implements component public instance, just using the interface should be enough.

pikax: That's a bit too lenient, because the `defineComponent` will also return a valid component public instance to be used on JSX

lmiller1990: Yep, agree, having `props` intellisense is pretty cool. I don't think it's possible for `data` in class component though.

lmiller1990:
@pikax is this good to go?

pikax:
@lmiller1990 I'm good with this typing

lmiller1990 (approved):

Issue #249, opened 2020-11-19
Allow stubbing globally registered components (eg <router-view />)


cyberb:
I am using router mock and my test works fine, but I get this warning because I have a router-view:

```
console.warn node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:40
    [Vue warn]: Failed to resolve component: router-view 
      at <VueApp ref="VTU_COMPONENT" > 
      at <VTUROOT>
```

Test: https://github.com/syncloud/platform/blob/98bf6a15da825533248df78b46d8b0ac9787de35/www/tests/unit/VueApp.spec.js
Adding stubs did not help:
```
global: {
  mocks: {
    $route: mockRoute,
    $router: mockRouter
  },
  stubs: ['router-view']
}
```

cexbrayat:
Hi @cyberb 

I usually use:
```ts
import { mount, RouterLinkStub } from '@vue/test-utils';

mount(Navbar, {
    global: {
      stubs: {
        RouterLink: RouterLinkStub
      }
    }
  });
```

or just `RouterLink: true`.

Can you give this a try?

lmiller1990:
We might need to include a `RouterViewStub` too? You could try something like this:

```
stubs: {
  RouterView: true
}
```

To match the [name](https://github.com/vuejs/vue-router-next/blob/656a2d767964babbbea8f424e21dfefbf289264c/src/RouterView.ts#L37) property - this is one way we figure out how to stub things.

Alternatively, I have had luck registering a global component too (since that is what Vue Router does):

```
global: {
  components: {
     RouterView: { render() { return h('div') } }
  }
}
```

The line between `stubs` and `component` is kind of blurred. I wonder if we should make this distinction more clear? Eg `stubs` for locally registered components, `components`for global ones (like `RouterLink` and `RouterView`).

What do you guys think?

cexbrayat:
Ha yes, I misread the OP and answered with `RouterLink`, sorry @cyberb ! `RouterView: true` works 👍

@lmiller1990 I agree that we should also provide a `RouterViewStub` for convenience.

For your other question, it is indeed a bit confusing that you can use one or the other in some cases. Wouldn't it be weird to limit `components` to globally registered ones as the same name in a component is for locally registered ones?

lmiller1990:
@cexbrayat maybe... it looks like you can either use global.components or stubs with little to no difference most of the time. 🤔 

[...]

PR #260, opened 2020-11-24
issue-249 feat: stub global components

lmiller1990:
Maybe @cyberb can have some input here too.

This addresses https://github.com/vuejs/vue-test-utils-next/issues/249

At the moment `stubs` will only work for locally registered components. For global ones (like `<router-link>`) you currently must registered a component globally to stub it, eg:

```js
mount(Foo, {
  components: {
    RouterLink: {
     render() { return h('div') }
    }
  }     
})
```

Many people expect this to work (I also do)

```js
mount(Foo, {
  global: {
    stubs: {
      RouterLink: {
       render() { return h('div') }
      }
    }
  }     
})
```

This PR takes anything from `global.stubs` and registers a global stub.

### DIFF
--- a/src/components/RouterLinkStub.ts
+++ b/src/components/RouterLinkStub.ts
@@ -12,6 +12,6 @@ export const RouterLinkStub = defineComponent({
   },
 
   render() {
-    return h('a', undefined, this.$slots.default())
+    return h('a', undefined, this.$slots?.default?.())
   }
 })
--- a/tests/mountingOptions/stubs.global.spec.ts
+++ b/tests/mountingOptions/stubs.global.spec.ts
@@ -1,4 +1,4 @@
-import { h, ComponentOptions } from 'vue'
+import { h, ComponentOptions, defineComponent } from 'vue'
 
 import { config, mount } from '../../src'
 import Hello from '../components/Hello.vue'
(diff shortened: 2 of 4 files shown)


tests/mountingOptions/stubs.global.spec.ts
```diff
@@ -37,6 +37,29 @@ describe('mounting options: stubs', () => {
...
+              return h('span')
+            }
+          })
+        }
+      }
+    })
+
+    expect(wrapper.html()).toBe(
+      '<div><router-link-stub></router-link-stub><span></span></div>'
+    )
```
cexbrayat, line 60: As I'm a bit unclear on how everything is related, can you maybe add a test where we try to find the component registered as a stub? I'd like to be sure that `findComponent` and friends work in that case as well.

I think a real use-case would be to stub `RouterLink` with `RouterLinkStub`, find it, and check its props to see if the component is creating links with the appropriate values. So I'd like to make sure we cover this 😅

cexbrayat (commented):
I think it makes sense to move in that direction, it felt weird to write `components: { RouterLink: RouterLinkStub }` in tests 👍

That now leaves the question: do we need to keep `components` if we can do the same with `stubs`? But maybe we can't and I'm missing something.

tests/mountingOptions/stubs.global.spec.ts
lmiller1990: Good idea

lmiller1990:
@cexbrayat the global `components` option probably isn't really useful. That said, `global` is supposedmirro what you can do with `app`, eg `app.component`, `app.mixin` etc, so I think it's okay to have it. I don't expect people to use it, though (how often does someone register a global component? Only plugins and that sort of thing, I guess).

src/components/RouterLinkStub.ts
```diff
@@ -12,6 +12,6 @@ export const RouterLinkStub = defineComponent({
   },
 
   render() {
-    return h('a', undefined, this.$slots.default())
+    return h('a', undefined, this.$slots?.default?.())
```
lmiller1990, line 15: It is possible to use router-link `<router-link />` (however unlikely)

tests/mountingOptions/stubs.global.spec.ts
lmiller1990: Done: https://github.com/vuejs/vue-test-utils-next/pull/260/commits/20c76940652cb64a4dab4b6bfd4ce45e4ca30ea9

Is this what you mean?

cexbrayat: Perfect, thanks Lachlan!

lmiller1990: Cool, want to give me a big old ✅

cexbrayat (approved):

Issue #263, opened 2020-11-30
Not working with Vue 3.0.3 and latest <script setup>


lmiller1990:
Related: https://github.com/vuejs/vue-jest/issues/304

Seems 3.0.3 breaks test utils. The return value either [here](https://github.com/vuejs/vue-test-utils-next/blob/ea21eb1dbfe292bf2c5215bece13a6189caeb519/src/mount.ts#L427) and [here](https://github.com/vuejs/vue-test-utils-next/blob/ea21eb1dbfe292bf2c5215bece13a6189caeb519/src/mount.ts#L429) are different to normal.

I am not sure if we should fix this right now, or wait for the `<script setup>` to stop been so experimental. It's important to let people try the latest features, but I am not sure `<script setup>` is meant for production yet - thoughts?

lmiller1990:
Here is my analysis:

- new `<script setup>` introduced in 3.0.3. It breaks things.

Using `vue@3.0.2` and the same version of compiler-sfc with this component works:

```html
<script setup lang="ts">
// imported components are also directly usable in template
import { ref } from 'vue'

// write Composition API code just like in a normal setup()
// but no need to manually return everything
export const count = ref(0)
export const inc = () => { count.value++ }
</script>

<template>
  <button @click="inc">{{ count }}</button>
</template>
```

To make this work with the new `<script setup>` you just need to remove `export` from `count` and `inc` - those are automatically exposed. Doing that, then upgrading to `vue@3.0.3` breaks things.

If you do a log [here](https://github.com/vuejs/vue-test-utils-next/blob/ea21eb1dbfe292bf2c5215bece13a6189caeb519/src/mount.ts#L427) on `vue@3.0.2` you get:

```
vm { __emitted: {} }
```

And [here](https://github.com/vuejs/vue-test-utils-next/blob/ea21eb1dbfe292bf2c5215bece13a6189caeb519/src/mount.ts#L429):

```
App { count: [Getter/Setter], inc: [Getter/Setter], __emitted: {} }
```

With `vue@3.0.3`:

```
vm { __emitted: {} }
```

and:

```
App {}
```

Seems [this line](https://github.com/vuejs/vue-test-utils-next/blob/ea21eb1dbfe292bf2c5215bece13a6189caeb519/src/mount.ts#L429) is where things break. `vm.$refs[MOUNT_COMPONENT_REF]` is *not* returning the expected `vm`.

martinszeltins:
[...]

PR #264, opened 2020-11-30
Support latest <script setup>

lmiller1990:
Left some notes about debugging this [here](https://github.com/vuejs/vue-test-utils-next/issues/263).

For *some* reason getting `vm` is different for components using the new `<script setup>` rfc - I have no idea why. I was able to work around it as shown in this PR... I don't really understand why this works. That said, it *does* work 🤷‍♂️  I did a dive into the implementation but it doesn't look like anything change in core that would break this.

If anyone can figure out what is different about `<script setup>` components that makes them un-findable with `$refs` like we currently do it, that'd be great. I would like to support the latest `<script setup>` rfc, so people can give more feedback about it.


cexbrayat (approved):
Nice work tracking this down!

The fix is weird to be honest, but I don't know enough about <script setup> to offer an alternative. Might be good enough until the dust settles.

package.json
```diff
@@ -23,7 +23,7 @@
     "@types/estree": "^0.0.42",
     "@types/jest": "25.2.1",
     "@types/node": "12.12.35",
-    "@vue/compiler-sfc": "^3.0.2",
+    "@vue/compiler-sfc": "3.0.3",
```
afontcu, line 26: Are these an intended change?

lmiller1990: Yes, `<script setup>` works completely differently in 3.0.2 and 3.0.3.

So kind of a breaking change - for `<script setup>` people. Those users are on the cutting edge anyway, so they probably know enough to figure out that they need the absolute latest version.

PR #274, opened 2020-12-14
types: fix mount infer prop type

### LINKED ISSUE #237: Typescript error when mounting component with boolean prop
The following code shows a typescript error on the first mount call. It only seems to appear when adding a boolean prop.

```js

const Foo = defineComponent({
  name: 'Foo',
  props: {
    bar: Boolean,
    baz: String,
  },
  setup () {
    return () => h('div', ['hello world'])
  }
})

test('uses mounts', async () => {
  mount(Foo, {
    props: {
      baz: 'hello',
    }
  })

  mount(Foo, {
    props: {
      bar: true,
    }
  })
})
```

```ts
No overload matches this call.
  The last overload gave the following error.
    Argument of type 'DefineComponent<{ bar: BooleanConstructor; baz: StringConstructor; }, () => VNode<RendererNode, RendererElement, { [key: string]: any; }>, ... 9 more ..., { ...; }>' is not assignable to parameter of type 'ComponentOptionsWithObjectProps<readonly string[] | Readonly<ComponentObjectPropsOptions<Record<string, unknown>>>, () => VNode<RendererNode, RendererElement, { ...; }>, ... 8 more ..., { ...; } | {}>'.
      Property 'props' is missing in type 'ComponentPublicInstanceConstructor<{ $: ComponentInternalInstance; $data: {}; $props: Partial<{ bar: boolean; }> & Pick<Readonly<{ bar: boolean; } & { baz?: string; }> & VNodeProps & AllowedComponentProps & ComponentCustomProps, "baz" | ... 9 more ... | "style">; ... 10 more ...; $watch(source: string | Function, cb...' but required in type '{ props: (readonly string[] & ThisType<void>) | (Readonly<ComponentObjectPropsOptions<Record<string, unknown>>> & ThisType<void>); }'.ts(2769)
```

Here are the versions I'm using

```
"@vue/test-utils": "^2.0.0-beta.8",
"vue": "^3.0.2",
```

pikax:
fix #237 

Fix issue of not making default properties as optional.

Allow to pass extra props (same behaviour as `h`) 

Commented on tests which prevented extra props.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -4,7 +4,8 @@ import {
   Directive,
   Plugin,
   AppConfig,
-  VNode
+  VNode,
+  VNodeProps
 } from 'vue'
 
 interface RefSelector {
@@ -32,7 +33,14 @@ type SlotDictionary = {
   [key: string]: Slot
 }
 
-export interface MountingOptions<Props, Data = {}> {
+type RawProps = VNodeProps & {
+  // used to differ from a single VNode object as children
+  __v_isVNode?: never
+  // used to differ from Array children
+  [Symbol.iterator]?: never
+} & Record<string, any>
+
+export interface MountingOptions<P, Data = {}> {
   /**
    * Overrides component's default data. Must be a function.
    * @see https://vue-test-utils.vuejs.org/v2/api/#data
@@ -42,11 +50,11 @@ export interface MountingOptions<Props, Data = {}> {
    * Sets component props when mounted.
    * @see https://vue-test-utils.vuejs.org/v2/api/#props
    */
-  props?: Props
+  props?: (RawProps & P) | ({} extends P ? null : never)
   /**
    * @deprecated use `data` instead.
    */
-  propsData?: Props
+  propsData?: P
   /**
    * Sets component attributes when mounted.
    * @see https://vue-test-utils.vuejs.org/v2/api/#attrs
--- a/src/types.ts
+++ b/src/types.ts
@@ -33,6 +33,8 @@ type SlotDictionary = {
   [key: string]: Slot
 }
 
+// From vue next
+// https://github.com/vuejs/vue-next/blob/1f2a652a9d2e3bec472fb1786a4c16d6ccfa1fb1/packages/runtime-core/src/h.ts#L53-L58
 type RawProps = VNodeProps & {
   // used to differ from a single VNode object as children
   __v_isVNode?: never
@@ -40,7 +42,7 @@ type RawProps = VNodeProps & {
   [Symbol.iterator]?: never
 } & Record<string, any>
 
-export interface MountingOptions<P, Data = {}> {
+export interface MountingOptions<Props, Data = {}> {
   /**
    * Overrides component's default data. Must be a function.
    * @see https://vue-test-utils.vuejs.org/v2/api/#data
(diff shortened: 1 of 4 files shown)


src/types.ts
```diff
@@ -32,7 +33,14 @@ type SlotDictionary = {
   [key: string]: Slot
 }
 
-export interface MountingOptions<Props, Data = {}> {
+type RawProps = VNodeProps & {
+  // used to differ from a single VNode object as children
+  __v_isVNode?: never
+  // used to differ from Array children
+  [Symbol.iterator]?: never
+} & Record<string, any>
```
cexbrayat, line 41: This is from vue-next right?

Would it be possible to export it in vue-next? Or at least adding a comment to explain that it is a copy/paste would be great 👍

src/types.ts
```diff
@@ -32,7 +33,14 @@ type SlotDictionary = {
...
+
+export interface MountingOptions<P, Data = {}> {
```
cexbrayat, line 43: nit: Props is a bit more explicit than P

cexbrayat (changes requested):
LGTM.

I liked not being able to pass extra props, because it catches error like `mount(Foo, { typo: value})` if the props is `type`. But I understand the reasoning.

It would be aswome to squash the 2 commits into one to have a clearer history. And maybe remove the commented tests, or even better, fix them instead of commenting them (because we won't remember why they are commented :) ).

test-dts/shallowMount.d-test.ts
```diff
@@ -21,12 +21,12 @@ let wrapper = shallowMount(AppWithDefine, {
 // vm is properly typed
 expectType<string>(wrapper.vm.a)
 
-// can not receive extra props
-expectError(
-  shallowMount(AppWithDefine, {
-    props: { a: 'Hello', c: 2 }
-  })
-)
+// // can not receive extra props
```
lmiller1990, line 24: How come we don't need these anymore? Shouldn't these continue to error out if we infer the props?

pikax: This will make the passing props a bit more relaxed, is also the same behaviour as [h#vue-next](https://github.com/vuejs/vue-next/blob/master/packages/runtime-core/src/h.ts#L127)

lmiller1990 (approved):

lmiller1990:
@cexbrayat any more comments or can you drop a ✅

test-dts/shallowMount.d-test.ts
cexbrayat: I think it would be better to uncomment the test and remove the `expectError` then, with a comment explaining that extra props is allowed as it is for `h()`. I don't like having commented code, because we will forget why ^^. And if the tests check that extra props are allowed, we are sure to avoid regressions in the future.

pikax:
Made the changes, also added the `PublicProps` to all the overloads, but just wondering if that behaviour is expected on the test utils : 
```ts
//https://github.com/vuejs/vue-next/blob/master/packages/runtime-core/src/vnode.ts#L87-L98
export type VNodeProps = {
  key?: string | number
  ref?: VNodeRef

  // vnode hooks
  onVnodeBeforeMount?: VNodeMountHook | VNodeMountHook[]
  onVnodeMounted?: VNodeMountHook | VNodeMountHook[]
  onVnodeBeforeUpdate?: VNodeUpdateHook | VNodeUpdateHook[]
  onVnodeUpdated?: VNodeUpdateHook | VNodeUpdateHook[]
  onVnodeBeforeUnmount?: VNodeMountHook | VNodeMountHook[]
  onVnodeUnmounted?: VNodeMountHook | VNodeMountHook[]
}
```

cexbrayat (approved):
LGTM, thanks @pikax !

@lmiller1990 It would be awesome to have a release with this change so we can check if everything is fine in our respective projects 😃

lmiller1990:
@cexbrayat sure: https://github.com/vuejs/vue-test-utils-next/releases/tag/v2.0.0-beta.13

PR #277, opened 2020-12-17
fix: handle slots correctly

### LINKED ISSUE #221: Programatically accessing a slot in a test always returns null for children
I've been having a weird issue with migrating out component library over to Vue 3's options API (for now), with the intention to eventually moving each component to use the composition API.

I have a helper function that can be used to show or hide sections of a component depending on whether a given slot exists and has content. In trying to update that I noticed that in Vue Test Utils, accessing `$slots.default()` always has `children: null`, even if that slot has content, and even if that content is rendered in the test.

For example, a simple sample component such as

```html
<template>
  <div class="super-simple-component">
    <slot />
  </div>
</template>

<script>
import { defineComponent } from 'vue';

  name: 'super-simple-component',
});
</script>
```

and a test such as

```javascript

import SuperSimpleComponent from '@/components/super-simple-component';

test('Slot content example', () => {
  const wrapper = mount(SuperSimpleComponent, {
    slots: {
      default: 'Default slot content',
    },
  });

  console.log({ html: wrapper.html(), programatic: wrapper.vm.$slots.default() });
});
```

outputs

```
{
  html: '<div class="super-simple-component">Default slot content</div>',
  programatic: [{
    __v_isVNode: true,
    __v_skip: true,
    type: [Object],
    props: null,
    key: null,
    ref: null,
    scopeId: null,
    children: null,
    component: null,
    suspense: null,
    ssContent: null,
    ssFallback: null,
    dirs: null,
    transition: null,
    el: null,
    anchor: null,
    target: null,
    targetAnchor: null,
    staticCount: 0,
    shapeFlag: 4,
    patchFlag: 0,
    dynamicProps: null,
    dynamicChildren: null,
    appContext: null
  }]
}
```

I'm not sure if I'm missing a setup stage?

lmiller1990:
resolves #221 

@lewishowles quick review/test if you can, since you have a use case for this feature.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -82,3 +82,17 @@ export function isFunctionalComponent(component: any) {
 export function isObjectComponent(component: any) {
   return typeof component !== 'function' && !isClassComponent(component)
 }
+
+// https://stackoverflow.com/questions/15458876/check-if-a-string-is-html-or-not/15458987#answer-15458968
+export function isHTML(str: string) {
+  var a = document.createElement('div')
+  a.innerHTML = str
+
+  for (let c = a.childNodes, i = c.length; i--; ) {
+    if (c[i].nodeType == 1) {
+      return true
+    }
+  }
+
+  return false
+}
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -26,6 +26,7 @@ import { config } from './config'
 import { MountingOptions, Slot } from './types'
 import {
   isFunctionalComponent,
+  isHTML,
   isObjectComponent,
   mergeGlobalProperties
 } from './utils'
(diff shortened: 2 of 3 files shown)


dobromir-hristov (commented):
Heh looks ok to me :D didn't thought about this..,

src/utils.ts
```diff
@@ -82,3 +82,17 @@ export function isFunctionalComponent(component: any) {
...
+  var a = document.createElement('div')
+  a.innerHTML = str
+
+  for (let c = a.childNodes, i = c.length; i--; ) {
+    if (c[i].nodeType == 1) {
+      return true
+    }
+  }
+
+  return false
```
cexbrayat, line 97: OOC: ain't that a very expensive check that'll slow down the tests?

lmiller1990: I don't think so - it's the last case in the `reduce` function, and we return true early - so this only executes if you are passing a text slot, which is not super common from what I can tell.

I don't expect a *significant* performance impact. I have no idea how expensive this actually is, we might want to start having some automated benchmarking at some point.

Do you know of a cheaper way to do this? We could do something like check for both `<` and `>` in the string or something - that would likely solve for 99% of cases. Thoughts?

cexbrayat: You' re probably right, I missed that it was the last case, so it shouldn't matter much 👍.

PR #282, opened 2020-12-22
feat(types): infer options in plugins

### LINKED ISSUE #269: Expose typings for Plugins
I think it would be helpful to expose the `Instance` type so plugins can type the argument (https://vue-test-utils.vuejs.org/v2/guide/plugins.html#writing-a-plugin) as well as adapting the `install` function to infer the options parameter:

```ts
    install(handler: (instance: Instance) => Record<string, any>, options?: Record<string, any>): void;
```

Similar to https://github.com/vuejs/vue-next/pull/2758

posva:
Close #269

### DIFF
--- a/test-dts/plugins.d-test.ts
+++ b/test-dts/plugins.d-test.ts
@@ -38,7 +38,7 @@ config.plugins.VueWrapper.install(PluginWithOptionalOptions2)
 config.plugins.VueWrapper.install(PluginWithOptionalOptions, { msg: 'hello' })
 config.plugins.VueWrapper.install(PluginWithOptionalOptions2, { msg: 'hello' })
 
-// FIXME: not sure if this one is possible
+// uncertain if it is possible to forbid this usage
 // expectError(config.plugins.VueWrapper.install(PluginWithoutOptions, {}))
 expectError(config.plugins.VueWrapper.install(PluginWithOptions, { msg: true }))
 expectError(config.plugins.VueWrapper.install(PluginWithOptions, {}))
--- a/test-dts/plugins.d-test.ts
+++ b/test-dts/plugins.d-test.ts
@@ -1,0 +1,46 @@
+import { expectError, expectType } from 'tsd'
+import { ComponentPublicInstance } from 'vue'
+import { config, VueWrapper } from '../src'
+
+interface OptionsI {
+  msg: string
+}
+
+function PluginWithOptions(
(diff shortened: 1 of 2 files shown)


test-dts/plugins.d-test.ts
```diff
@@ -0,0 +1,46 @@
...
+
+config.plugins.VueWrapper.install(PluginWithOptions, { msg: 'Hello' })
+config.plugins.VueWrapper.install(PluginWithoutOptions)
+config.plugins.VueWrapper.install(PluginWithOptionalOptions)
+config.plugins.VueWrapper.install(PluginWithOptionalOptions2)
+config.plugins.VueWrapper.install(PluginWithOptionalOptions, { msg: 'hello' })
+config.plugins.VueWrapper.install(PluginWithOptionalOptions2, { msg: 'hello' })
+
+// FIXME: not sure if this one is possible
+// expectError(config.plugins.VueWrapper.install(PluginWithoutOptions, {}))
```
cexbrayat, line 42: I guess this is fine?
The FIXME needs to be removed before merging.

cexbrayat (approved):

test-dts/plugins.d-test.ts
posva: I think it's fine too but I thought I might have overlooked a way of doing it in TS. If there is nothing to fix, yeah, better remove it 😆

cexbrayat: We _could_ probably do something, but I'm not sure it's worth the time, the hassle and the maintenance cost 😉 Thanks for the PR!

test-dts/plugins.d-test.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+config.plugins.VueWrapper.install(PluginWithOptions, { msg: 'Hello' })
+config.plugins.VueWrapper.install(PluginWithoutOptions)
+config.plugins.VueWrapper.install(PluginWithOptionalOptions)
+config.plugins.VueWrapper.install(PluginWithOptionalOptions2)
+config.plugins.VueWrapper.install(PluginWithOptionalOptions, { msg: 'hello' })
+config.plugins.VueWrapper.install(PluginWithOptionalOptions2, { msg: 'hello' })
+
+// FIXME: not sure if this one is possible
```
posva, line 41:
```suggestion
// uncertain if it is possible to forbid this usage
```

lmiller1990 (approved):
great, thanks!

Issue #278, opened 2020-12-19
Update to 3.0.4 and 3.0.5


lmiller1990:
As of Vue 3.0.4 this test is broken: https://github.com/vuejs/vue-test-utils-next/blob/c5d31307fe2dbd9bb04c3af3287cae93920014af/tests/mountingOptions/stubs.global.spec.ts#L348

Todo: find out why.

lmiller1990:
We will just wait for 3.0.5: #285

PR #284, opened 2020-12-27
feat: make  jest debug in vscode available

HomyeeKing:


lmiller1990:
What exactly does this do? I have not used Jest debug in VS Code before. Is it normal to commit editor specific files to repos? 

I guess vue-next has it : https://github.com/vuejs/vue-next so probably fine to have it here, too.

HomyeeKing:
@lmiller1990  easy to debug the test files , but vscode only, it's up to you 
maybe it's helpful with issue like  #278 , to find problem quickly
also it's good for person like me  to read the source code

PR #295, opened 2021-01-16
docs: Port docs to this repo (and move to vitepress)

afontcu:
PR looks obviously huge but it is due to copy&paste markdown files 😄


docs/.vitepress/config.js
```diff
@@ -0,0 +1,116 @@
+/** @typedef {import('vitepress').UserConfig} UserConfig */
+
+/** @type {UserConfig} */
+const config = {
+  // base: '/v2/',
```
afontcu, line 5: Pending to figure out where will docs get deployed

lmiller1990:
Finally, have been waiting for this.

I did not look at the diff yet, the basic questions I see are:

1. how do we handle deploys? Can we update this to auto deploy to netlify as we are currently doing (if we cannot, who do we need to ping?)
2. can this be merged as-is, or does some other config need to happen (or can that happen after merging)?

afontcu:
> I did not look at the diff yet, the basic questions I see are:
>     1. how do we handle deploys? Can we update this to auto deploy to netlify as we are currently doing (if we cannot, who do we need to ping?)
>     2. can this be merged as-is, or does some other config need to happen (or can that happen after merging)?

It is deployed on Netlify so we might need to tweak config to point to this repo. AFAIK this is the only thing missing before merging. I'm not sure we still need to set `base` to `/v2/` or if we could deploy it to a subdomain to make it easier to manage (next.vue-test-utils.vuejs.org)

lmiller1990:
We might want to see what Vue Router/Vuex is doing and follow that convention

afontcu:
> We might want to see what Vue Router/Vuex is doing and follow that convention

Yeah, that's where I got the idea from :) They use subdomains:

https://next.router.vuejs.org/
https://next.vuex.vuejs.org/

afontcu:
I'll go ahead and merge this one to be ready when Netlify config changes. Any additional tweaks can be done afterwards! 😃

Issue #292, opened 2021-01-09
Better way to handle `emitted` internally


lmiller1990:
There is a hook in core for devtools. We could use that? The close integration is likely going to be better, less buggy, and make things more reliable. https://github.com/vuejs/vue-next/blob/ee266af42ab8f97011c6736aa82409a35c402e94/packages/runtime-core/src/devtools.ts#L82

afontcu:
(_I've added "internally" to the title because I first thought you were about to present a better `emitted()` API 😅_)

cuixiaorui:
I did some research on devtools

It turns out that if you use devtools, it's not that different from what we have now

Devtools is used for the [Vue DevTools](https://github.com/vuejs/vue-devtools/)

This devtools is dependent on windows.__vue_devtools_global_hook__ to inject devtools

If we use devtools to implement emitted, it needs to deal with two conditional branches
- windows.__vue_devtools_global_hook__ if it has a value (already injected by Vue devtools)
  - We need to wrap windows.__vue_devtools_global_hook__. emit this function
- windows.__vue_devtools_global_hook__ does not have a value
  - We need to build our own devTools and inject it

Or maybe I didn't think it would be better than the current one :)

lmiller1990:
Thanks @cuixiaorui, that is useful info. There are some edge cases we miss with the current approach. For example:

```js
const C = {
  setup(props, { emit }) {
    emit('foo')
  }
}
```

We cannot capture this `emit` since we install a emit listener mixin on the `beforeCreate` hook, and `setup` is called prior. Would using the devtools hook fix this bug?

cuixiaorui:
Uh-oh, I see

I gave it a try, and devtools can solve this problem

I will submit a PR later

PR #296, opened 2021-01-18
refactor: use devtools realize emit

cuixiaorui:
for #292


afontcu:
Hi! 👋  Could we add a test to cover the scenario Lachlan outlined? https://github.com/vuejs/vue-test-utils-next/issues/292#issuecomment-761895062

src/emitMixin.ts
```diff
@@ -1,34 +1,53 @@
...
-      getCurrentInstance()!.emit = (event: string, ...args: unknown[]) => {
-        events[event]
-          ? (events[event] = [...events[event], [...args]])
-          : (events[event] = [[...args]])
+const enum DevtoolsHooks {
+  COMPONENT_EMIT = 'component:emit'
+}
+
+export const attachEmitListener = (vm: ComponentPublicInstance) => {
+  let events: Record<string, unknown[]> = {}
```
cexbrayat, line 8: `let` can be a `const` here?

src/emitMixin.ts
```diff
@@ -1,34 +1,53 @@
...
+  ;(vm as any).__emitted = events
+  // use devtools capture this "emit"
```
cexbrayat, line 10: nit: `use devtools to capture this "emit"`

src/emitMixin.ts
```diff
@@ -1,34 +1,53 @@
...
+  setDevtoolsHook(createDevTools(events))
+}
+
+function createDevTools(events) {
+  const devTools: any = {
```
cexbrayat, line 15: Instead of `any`, it would be great to properly type this. Ain't this a `DevtoolsHook` type?
Maybe there are not exposed by `vue-next` though?

src/emitMixin.ts
```diff
@@ -1,34 +1,53 @@
...
+    }
+  }
+
+  return devTools
+}
+
+function recordEvent(events, event, args) {
+  events[event]
+    ? (events[event] = [...events[event], [...args]])
+    : (events[event] = [[...args]])
```
cexbrayat, line 34: Some documentation of what this does would be great (it's not obvious at first read).

src/emitMixin.ts
```diff
@@ -1,34 +1,53 @@
...
+function wrapperWarn() {
+  const consoleWarnSave = console.warn
+  console.warn = (msg: string, ...rest: unknown[]) => {
+    if (msg.includes('[Vue warn]: Component emitted event')) {
+      return
+    } else {
+      consoleWarnSave(msg, ...rest)
     }
   }
+  console.warn = consoleWarnSave
```
cexbrayat, line 52: OOC: do we still need this?

cexbrayat (commented):
Thx for the PR! I added a few comments

src/emitMixin.ts
lmiller1990: Probably not - I put this there originally because I did not realize Vue 3 now wants you to declare the components  `emits` property, I think.

src/emitMixin.ts
lmiller1990: We can get it hackily (the interface itself is not available). not sure how beneficial it'll be:

```ts
import { devtools } from 'vue'

const blah: Partial<typeof devtools> = {
  emit: () => {}
}
```

May need to be a partial unless we want to define `on` and the other hooks (we could make them no-ops).

lmiller1990:
This PR is a great way to head into release candidate stage - the more value we can get from what is built into Vue core, the better.

src/emitMixin.ts
cuixiaorui: Yes, I don't think I need to deal with console.warn anymore

users should be declared ”emits“ property

cuixiaorui:
- I have removed the interception logic for "console.warn" and processed some test cases
- Refactoring to your opinion

cexbrayat (approved):
Thanks!

lmiller1990:
@cuixiaorui I tried this spec:

```js
  it('captures an event emitted in setup', () => {
    const Comp = {
      setup(_, { emit }) {
        emit('foo')
      }
    }
    const wrapper = mount(Comp)
    expect(wrapper.emitted().foo).toBeTruthy()
  })
```

Still failing. You said you got it working [in this comment](https://github.com/vuejs/vue-test-utils-next/issues/292#issuecomment-761990349) - was that the case?

cuixiaorui:
@lmiller1990 

sorry 

I ignored the test case you mentioned before

The previous implementation really didn't pass

I reworked the implementation logic

Now all the tests have passed

---

Now that we are no longer dependent on the "vm", we can move the emit logic ahead of the "createApp"

This is a good solution :)

lmiller1990 (approved):

lmiller1990:
Great job @cuixiaorui, great contribution!

I'll do a release soon with this.

PR #308, opened 2021-01-25
docs: Add "Write components that are easy to test" page

afontcu:
Comes from https://github.com/vuejs/vue-test-utils-next-docs/pull/52

### DIFF
--- a/docs/.vitepress/config.js
+++ b/docs/.vitepress/config.js
@@ -53,6 +53,10 @@ const config = {
           {
             text: 'Passing Data to Components',
             link: '/guide/essentials/passing-data'
+          },
+          {
+            text: 'Write components that are easy to test',
+            link: '/guide/essentials/easy-to-test'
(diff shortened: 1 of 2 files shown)


docs/.vitepress/config.js
```diff
@@ -53,6 +53,10 @@ const config = {
           {
             text: 'Passing Data to Components',
             link: '/guide/essentials/passing-data'
+          },
+          {
+            text: 'Write components that are easy to test',
+            link: '/guide/essentials/easy-to-test'
```
afontcu, line 59: Not 100% sure this is the right place for this kind of content

lmiller1990: 🤔 

Not sure. Either way, should we merge this one up so it's at least available for people to use?

afontcu: sure! 🚀

PR #310, opened 2021-01-26
fix: spy on wrapper.vm with Jest

### LINKED ISSUE #309: Jest is no longer able to spy on methods
I think the wrapper no longer exposes methods on the prototype or something, because when I want to spy on a method (e.g. `wrapper.vm.send`) with `jest.spyOn(wrapper.vm, 'send')` Jest throws a type error: `TypeError: object.hasOwnProperty is not a function`.

This makes it impossible to test some component internals (like methods that format data and send it off to a state machine, in this case).

cexbrayat:
Allow to spy on `wrapper.vm` with jest.

Fixes #309 

BLocked by https://github.com/vuejs/vue-next/pull/3105

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -441,8 +441,11 @@ export function mount(
   // if not, use the return value from app.mount.
   const appRef = vm.$refs[MOUNT_COMPONENT_REF] as ComponentPublicInstance
   const $vm = Reflect.ownKeys(appRef).length ? appRef : vm
+  // we add `hasOwnProperty` so jest can spy on the proxied vm without throwing
+  $vm.hasOwnProperty = (property) => {
+    return Reflect.has($vm, property)
+  }
   console.warn = warnSave
-
   return createWrapper(app, $vm, setProps)
 }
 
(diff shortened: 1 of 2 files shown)


cexbrayat:
I have a dirty fix that I'm not super proud of, and which needs a fix in vue-next.
If someone has a better idea, please go ahead :)

pikax:
This fix won't work, because the `vm` is read only so the `jest.spyOn` will not work even with this fix, because it relies on replacing the original function with a spied version.

lmiller1990 (approved):
nice!

lmiller1990:
hm, CI failing?

afontcu:
> hm, CI failing?

I think we still need https://github.com/vuejs/vue-next/pull/3105 to be merged and released?

cexbrayat:
Yes exactly, we need my PR on vue-next to land (if that's the right move, we'll see!).

lmiller1990:
Right, true.

cexbrayat:
This patch can now land since vue-next v3.0.6 contains the necessary upstream fix.

PR is green and already reviewed, so let's merge and fix #309

PR #354, opened 2021-02-05
fix: record emitted events against each wrapper

### LINKED ISSUE #352: wrapper.emitted includes sub-component events with the same name
## Description
When testing component events, we often trigger an event on an inner sub-component to mimic user input, and observe a final event get emitted from the component under test, sometimes with a different / transformed value. However the current version of `vue-test-utils-next` makes this difficult by returning all emitted events with the same name, even if they were not emitted directly by the component under test.

This has become more of an issue with the recommendation not to use shallow mount, and the prevalence of `update:modelValue` events when using `v-model` heavily (ie in forms).

## Expected behaviour

I would expect `wrapper.emitted()` to only include events emitted by the top-level component provided as the first argument to `mount`.

## Observed behaviour

`wrapper.emitted()` includes events emitted from sub-components, adding to events observed under the same name.

## Example

```js

const RawInput = {
  template: '<input v-model="computedValue" />',
  emits: ['update:modelValue'],
  props: ['modelValue'],
  computed: {
    computedValue: {
      get() { return this.modelValue; },
      set(val) { this.$emit('update:modelValue', val); },
    },
  },
};
const CSVInput = {
  template: '<div class="container"><RawInput v-model="computedValue" /></div>',
  components: { RawInput },
  emits: ['update:modelValue'],
  props: ['modelValue'],
  computed: {
    computedValue: {
      get() { return this.modelValue.join(' '); },
      set(val) { this.$emit('update:modelValue', val.split(',')); },
    },
  },
};

test('emitted events', () => {
  const wrapper = mount(CSVInput, { props: { modelValue: [] } });
  // inner component emits 'foo,bar'
  wrapper.get('input').setValue('foo,bar');

  // we want to test outer component emits ['foo', 'bar']
  expect(wrapper.emitted('update:modelValue')).toStrictEqual([[['foo', 'bar']]]);

  // instead wrapper.emitted('update:modelValue') contains events for both
  // inner and outer components
});
```
```
Error: expect(received).toStrictEqual(expected) // deep equality

- Expected
+ Received

@@ -1,7 +1,10 @@
  Array [
    Array [
+     "foo,bar",
+   ],
+   Array [
      Array [
        "foo",
        "bar",
      ],
    ],
```

# Version

`2.0.0-rc.0`

This also appears to be a regression that was not present in earlier versions, as updating to `2.0.0-rc.0` broke about 12 tests in our test suite that were previously working with the expected behaviour.

aethr:
fix #352 

Emitted events are recorded against their component uid, so the events retrieved by `emitted()` doesn't contain child events.

Updated one of the tests to probe for this as well.

First stab at any typescript so please let me know if I've committed terrible type sins. :pray:

### DIFF
--- a/src/emit.ts
+++ b/src/emit.ts
@@ -1,20 +1,27 @@
-import { setDevtoolsHook, devtools } from 'vue'
+import { setDevtoolsHook, devtools, ComponentPublicInstance } from 'vue'
 
 const enum DevtoolsHooks {
   COMPONENT_EMIT = 'component:emit'
 }
 
-let events: Record<string, unknown[]>
+let componentEvents: Record<string, unknown[]>
+let events: Record<string, typeof componentEvents>
 
 export function emitted<T = unknown>(
+  vm: ComponentPublicInstance,
   eventName?: string
 ): T[] | Record<string, T[]> {
+  const cid = vm.$.uid
+
+  const vmEvents = (events as Record<string, Record<string, T[]>>)[cid] || {}
   if (eventName) {
-    const emitted = (events as Record<string, T[]>)[eventName]
+    const emitted = vmEvents
+      ? (vmEvents as Record<string, T[]>)[eventName]
+      : undefined
     return emitted
   }
 
-  return events as Record<string, T[]>
+  return vmEvents as Record<string, T[]>
 }
 
 export const attachEmitListener = () => {
@@ -28,20 +35,27 @@ function createDevTools(events): any {
     emit(eventType, ...payload) {
       if (eventType !== DevtoolsHooks.COMPONENT_EMIT) return
 
-      // The first argument is root component
-      // The second argument is  vm
-      // The third argument is event
-      // The fourth argument is args of event
-      recordEvent(events, payload[2], payload[3])
+      const [rootVM, componentVM, event, eventArgs] = payload
+      recordEvent(events, componentVM, event, eventArgs)
     }
   }
 
   return devTools
 }
 
-function recordEvent(events, event, args) {
+function recordEvent(events, vm, event, args): void {
+  // Functional component wrapper creates a parent component
+  let wrapperVm = vm
+  while (typeof wrapperVm.type === 'function') wrapperVm = wrapperVm.parent
+
+  const cid = wrapperVm.uid
+  if (!(cid in events)) {
+    events[cid] = {}
+  }
+  if (!(event in events[cid])) {
+    events[cid][event] = []
+  }
+
   // Record the event message sent by the emit
-  events[event]
-    ? (events[event] = [...events[event], [...args]])
-    : (events[event] = [[...args]])
+  events[cid][event].push(args)
 }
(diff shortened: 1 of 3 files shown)


src/emit.ts
```diff
@@ -28,20 +35,27 @@ function createDevTools(events): any {
...
   }
 
   return devTools
 }
 
-function recordEvent(events, event, args) {
+function recordEvent(events, vm, event, args): void {
+  // Functional component wrapper creates a parent component
+  let wrapperVm = vm
+  while (typeof wrapperVm.type === 'function') wrapperVm = wrapperVm.parent
```
lmiller1990, line 49: I wonder if this will behave weirdly with class components 🤔 those (might) also return `type: 'function'`.

What do non-functional component return for `type`?

lmiller1990 (commented):
Good catch and fix. Are you able to see if there is any weirdness with class components (if there is and it wasn't introduced here, we can probably make a new issue to deal with those).

src/emit.ts
aethr: Just from observations coming out of the test suite (again I haven't tried this branch against my real project), non-functional components have an `object` type. Haven't tried class components (and not super familiar with them since our project doesn't use TS), but I can have a crack at putting a test case in for them after some research. :)

This part of the PR definitely felt a bit hacky, but unfortunately VTU seems to have 2x components in the hierarchy for functional components (they even have the same `el` reference). During the devtool `emit` hook, the `vm` argument is the inner component (functional) but when the user calls `component.emitted()` from the test, `vm` is the outer (object) component, so the `$.uid` is different.

I'm not that familiar with VTU or functional components so not sure about the deeper reasoning here, was just my naive attempt to bash out some code that gets the tests to pass. Very happy to take a different approach but might need a bit of guidance.

aethr: Added a class-style component test and it seems fine. The `type` for class-style components is also `object`,

I'm sure this can be improved, happy to make any other changes!

lmiller1990: Cool, thanks. Don't worry about feeling hacky here, this code base has tons of weird hacks to do things you can't otherwise do (usually for good reason, like mutating props, intercept events, etc.

All the tests are green, no reason not to merge. I'll do a release at the end of the week, hoping to get a few more small fixes in.

aethr: I had a thought, even though it's already merged I thought I'd put it out there.

Since we're discouraged to use `$.uid`, I considered whether VTU might generate it's own unique id for each Wrapper (`tid` or something), and during `emit` we could traverse up the hierarchy until we find a component with `tid` set. For most components this would be the initial `vm` but for functional it would likely be the parent.

This would be guaranteed to be a `Wrapper`, we could use the `tid` instead of `$.uid` for storage and retrieval, and we wouldn't be doing opaque "what type is this" checks that might be hard to understand in 6 months.

lmiller1990: sure, this sounds like a useful thing to have. do you want to make a PR? we could use something like `nanoid`. we are trying to keep this lib dependency free but you could just copy-paste it in: https://github.com/ai/nanoid/blob/main/index.js

or just math.random, up to you

PR #361, opened 2021-02-07
 #183 - explore teleport support using the teleportTarget attribute

### LINKED ISSUE #183: Feature request: Test against teleport element
Hi, I've been trying to testing `Teleport` using `VTU` but I went through the code and I can not find any helper for testing teleported component, can we add a method to `vueWrapper` to find `teleported` children inside the component itself? I do 
believe this is something users would be constantly asking about.

## Requesting Feature
Test helper for `Teleport` API

## Current situation

We can only testing teleported element by raw document query.

nandi95:
closes #183 
Branch exploring how how teleport support can be added


lmiller1990:
That was quick! I will review it properly and try it out in the next few days. I have experimented w/ `<teleport>` a bit a while ago, interested to see how this works with the test cases I wrote.

src/vueWrapper.ts
```diff
@@ -38,7 +50,124 @@ export class VueWrapper<T extends ComponentPublicInstance> {
...
   }
 
+  /**
+   * Get all the target elements defined in the teleportTarget mount option.
+   *
+   * @private
+   *
+   * @return {Element[]|undefined}
+   */
+  private getTeleportElements(): Element[] | undefined | never {
```
cexbrayat, line 60: `| never` is necessary here?

cexbrayat: Maybe we could avoid returning undefined, and just return an empty array if there are no elements? It would avoid TS users to have to add a non-null assertion when writing a test.

src/vueWrapper.ts
```diff
@@ -38,7 +50,124 @@ export class VueWrapper<T extends ComponentPublicInstance> {
...
+   *
+   * @return {Element}
+   */
+  private getTeleportElement(
+    vm: ComponentPublicInstance = this.componentVM
+  ): Element | never {
+    const elements = this.getTeleportElements()
+    const errorMsg =
+      vm.$.type.name +
+      "'s teleport target element cannot be found. Are you sure the target exits in the DOM?"
```
cexbrayat, line 100: Is `name` always going to be available? I'm not 100% sure, so to be safe I would just remove it from the message.

src/vueWrapper.ts
```diff
@@ -38,7 +50,124 @@ export class VueWrapper<T extends ComponentPublicInstance> {
...
+
+    if (!elements || !elements.length) {
```
cexbrayat, line 102: `!elements` can be removed if `getTeleportElements` returns `Element[]` as suggested above.

src/vueWrapper.ts
```diff
@@ -38,7 +50,124 @@ export class VueWrapper<T extends ComponentPublicInstance> {
...
+   *
+   * @param {ComponentPublicInstance} vm
+   *
+   * @private
+   *
+   * @return {Element}
+   */
+  private getTeleportElement(
+    vm: ComponentPublicInstance = this.componentVM
+  ): Element | never {
```
cexbrayat, line 96: Same here: I would remove `| never`?

src/vueWrapper.ts
```diff
@@ -38,7 +50,124 @@ export class VueWrapper<T extends ComponentPublicInstance> {
...
+    const elements = this.getTeleportElements()
+    const errorMsg =
+      vm.$.type.name +
+      "'s teleport target element cannot be found. Are you sure the target exits in the DOM?"
+
+    if (!elements || !elements.length) {
+      throw new Error(errorMsg)
+    }
+
+    const index = elements.findIndex((element) => {
```
cexbrayat, line 106: we could use `find` directly no?

src/vueWrapper.ts
```diff
@@ -38,7 +50,124 @@ export class VueWrapper<T extends ComponentPublicInstance> {
...
+          return (
+            child.hasOwnProperty('__vueParentComponent') &&
+            child.__vueParentComponent.uid === vm.$.uid
+          )
+        }
+      )
+    })
+
+    if (index === -1) {
+      throw new Error(errorMsg)
```
cexbrayat, line 122: Shouldn't that be a different error message? I think it would be less confusing to distinguish the two cases.

src/vueWrapper.ts
```diff
@@ -38,7 +50,124 @@ export class VueWrapper<T extends ComponentPublicInstance> {
...
+    const elements = Array.from(parentElement.children).filter(
+      (
+        child: Element & { __vueParentComponent: ComponentInternalInstance }
+      ) => {
+        return (
+          child.hasOwnProperty('__vueParentComponent') &&
+          child.__vueParentComponent.uid === vm.$.uid
+        )
+      }
+    )
```
cexbrayat, line 142: would be nice to extract this common code (if I'm not mistaken) into a utils function.

src/vueWrapper.ts
```diff
@@ -38,7 +50,124 @@ export class VueWrapper<T extends ComponentPublicInstance> {
...
+
+    if (elements.length !== 1) {
+      // todo
+      // this is presumably an impossible path as error would have been thrown if the element not found in the getTeleportElement
+      // or vue `uid` is not that unique
+      throw new Error('¯\\_(ツ)_/¯')
```
cexbrayat, line 148: would be nice to have a better message in the end ;)

cexbrayat (commented):
I missed that this was still in draft, so there are a bunch of nitpicking comments, sorry.

Overall this looks like a nice start.

I'm unclear on a few things though. If the component teleports to something else than body (like `#other`, contained in another part of the DOM), than the option is not really useful is it? You would have to define and mount a test component containing both the component you can to test and a div with the ID `other`. And in that case, you can access the element easily with a usual `find`. Or am I missing something?

nandi95:
@cexbrayat 
Those are valid points.
Please also note the convo over the issues especially https://github.com/vuejs/vue-test-utils-next/issues/183#issuecomment-774685973. I will try the mentioned approach first if possible.

nandi95:
> If the component teleports to something else than body 

That's why the idea is that the `teleportTarget` can also take an array of items and you would need to include every target that is present in your component (even from child components if not shallowmounting)

lmiller1990:
Wow nice review @cexbrayat, before we comment/merge I think we should explore any potential edge cases and scrutinise the API, too. What do you think of this API? How have you been testing your teleport components (if you have some?)

I still didn't get a chance to thoroughly go over everything and try it out, I can do it soon.

nandi95:
I found a way to automatically resolve the teleport targets and is a lot cleaner and robust. I will push the changes as soon as it's ready.

Can you please think of edge cases that needs to be covered? And I might not though of it.
Or if you're willing describe or write some tests?

I have a test component and this should cover the testing needs as far as I can tell.
```vue
<template>
  <teleport to="body">
    <div>decoy element</div>
    <div v-if="isVisible">
      <teleport to="#teleport-target">
        <slot name="nested-teleported-slot"></slot>
      </teleport>
      <slot />
    </div>
  </teleport>
</template>

<script lang="ts">
import { ref, defineComponent } from 'vue'

  name: 'ComponentWithTeleport',

  setup: () => {
    const isVisible = ref(false);
    const toggleVisibility = () => isVisible.value = !isVisible.value;

    return { isVisible, toggleVisibility };
  }
})
</script>

```

src/vueWrapper.ts
lmiller1990: Name will not always be available, no... eg components without name, functional components

lmiller1990:
One edge case might be a template without one root node - not sure why this would be an edge case, but it has been in the past for unexpected reasons.

Also regarding @cexbrayat's comments:

> I'm unclear on a few things though. If the component teleports to something else than body (like #other, contained in another part of the DOM), than the option is not really useful is it? You would have to define and mount a test component containing both the component you can to test and a div with the ID other. And in that case, you can access the element easily with a usual find. Or am I missing something?

I also am wondering about this. Let's say you have something like

```html
<template>
  <teleport to="#modal-wrapper">Content</teleport>
</template>
```

The `<div #modal-wrapper />` does not exist, presumably. What would make this a lot more useful that we could create the elements passed to `teleportTarget` dynamically? So it will create a `<div id="modal-wrapper" />` somewhere (who cares where, not sure if we need to attach it to the DOM or not) and then you could seemlessly do `wrapper.find`. What do you guys think?

I also find it really odd to do `wrapper.find` for something that is *not* in the wrapper (that's the point of `<teleport>` after all), but the current API for VTU is "put all methods on `wrapper`" so I think there is value in this convention. If we have a random `find` exposed (as in `import { find } from '@vue/test-utils'`) we will inevitably have to expose everything else as a stand-alone method, as well, to be consistent (and people will ask for it). Not really a bad thing, but a bigger discussion than we should have right now.

afontcu:
Hi, really interesting PR and discussion! Just wanted to give my two cents on this.

While the initial idea is good and legit, I think we're close to test that `Teleport` actually works (which is already tested on core, obviously 😄).

I also see some potential edge cases with this, and even more critically, we might end up with false negatives (tests that pass even if code is broken – see creating a DOM node to render the teleport content).

With that, what if we simply allowed `Teleport` to be stubbed (opt-in) so people could test teleported components as if they were not? VTU should enable people to test teleported content, and Vue core should test that given the right node, a teleported component gets… well, teleported 😅  Anything beyond this feels like a higher level test – say Cypress.

Thoughts?

Again, thank you for your time on this one!

nandi95:
> we could create the elements passed to teleportTarget dynamically

I don't think that's a good idea, your test should explicitly add the target to the DOM otherwise just let vue throw the warn: `[Vue warn]: Failed to locate Teleport target with selector...` otherwise it's a magic that takes your tests further from a real scenario, also people would be able to access our made up target by `.parentElement`.
However this could be an option down the line: `autoTeleport` which when turned on does in fact creates the targets.

> will inevitably have to expose everything else

Yes, my aim is to not make any change to the current api. Meaning `wrapper.find('#teleported-content')` will return the expected content which was teleported.

As for @afontcu's comment:

> With that, what if we simply allowed Teleport to be stubbed (opt-in) 

I understand this not the same as the above mentioned `autoTeleport` but another option. Which is to find the targets only if a boolean option turned on, otherwise just return `<!-- teleport start --><!-- teleport end -->` as it currently does?

nandi95:
@lmiller1990 @cexbrayat 

This is still a wip PR just wondering what you think about this new approach.

tests/mountingOptions/teleportTarget.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+describe('mountingOptions.teleportTarget', () => {
+  const nestedTeleportTarget = document.createElement('div')
+  nestedTeleportTarget.id = 'teleport-target'
+  document.body.appendChild(nestedTeleportTarget)
+
+  it('should find the html after teleporting element', async () => {
+    const wrapper = mount(ComponentWithTeleport, {
+      slots: {
+        default: '<div>default slot content</div>',
+        ['nested-teleported-slot']: '<div>nested teleport-content</div>'
```
posva, line 13:
```suggestion
        'nested-teleported-slot': '<div>nested teleport-content</div>'
```

tests/mountingOptions/teleportTarget.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+      }
+    })
+
+    expect(wrapper.html()).not.toContain('slot content')
+    // @ts-expect-error
+    await wrapper.vm.toggleVisibility()
+    expect(wrapper.html()).toContain('slot content')
```
posva, line 20: It would makes sense that the test makes sure the wrapper html doesn't contain `'nested teleport-content'`

nandi95: I think this can be up for discussion, perhaps resolve all teleports or just one level?

If we're happy with this approach, I think what we need is:
 -  a consensus of what mount option(s) we need
 - loads of tests, covering all that we can think of

Then this PR can be progressed further

posva: Everything is up for discussion but if the rendered component doesn't contain the HTML, it makes sense for the tested component not to as well

nandi95: But that's the point of this PR to resolve the elements that are not in the component but has been teleported out.

Or are you saying only resolve 1 level of teleports?

lmiller1990: I think @posva is saying that if something does not appear in the html for a component (like wrapper.html()) it should not be findable with `find` etc.

I think we will need to reconsider - using `wrapper[method]` for something *not* in the wrapper's markup feels a little weird.

lmiller1990:
I am a little concerned with `expect(wrapper.html()).toContain('slot content')`. When the `slot content` has been teleported - looking at `html()`, you will *not* see your content, right? This feels like it will get confusing - the current behavior of `html` is just "show me the markup".

Reflecting on this, maybe making wrapper more featureful is not the right way to do this. Considering the actual behavior of `<teleport>` is to *remove* the element from the component and put it somewhere else.

```js
const Comp = {
  template: `<teleport :to="#somewhere">FOO</teleport>`
}

let destination = null
const wrapper = mount(Comp, {
})

tests/mountingOptions/teleportTarget.spec.ts
nandi95: I think an argument can be made that you're still interacting with content that was defined in the component. But I see your point given that the element in the wrapper doesn't actually contain the expected markup anymore.

nandi95:
To me this level of magic seemed a good trade off for developer convenience.
 
> `createTeleportDestination` could be a wrapper,

So at this point we can just do `const teleportTarget = new DOMWrapper(document.querySelector('#somewhere'))` and it would be the same as `createTeleportDestination` no? If so, this can already be done now and this PR is no longer relevant.

tests/mountingOptions/teleportTarget.spec.ts
posva: I'm saying both:

- `wrapper.html()` should not include html from teleported components
- `wrapper.find()` should not include elements from teleported components _by default_

Because that's closer to reality and therefor easier for tests

lmiller1990:
Hm, what do you think? I think this exploration was definitely worth having because it revealed some complexities, edge cases, and also started good discussions (what is intuitive/convenient for one person may not be for everyone).

`const teleportTarget = new DOMWrapper(document.querySelector('#somewhere'))` seems pretty simple - it seems like it might be worth having an entire page in the docs about `<teleport>` outlining this technique. Is there any use-cases it doesn't solve?

nandi95:
Not an issue, maybe in the future someone can implement this as a plugin if they wish so.
I'll close this PR, then a new issue can be opened to make a note on testing teleports.

Issue #369, opened 2021-02-10
Code suggestions


nandi95:
I have couple suggestions I would like to propose to improve developer experience when contributing:

 -  [x] I noticed that `VueWrapper` has all the methods of `DOMWrapper`? Why not introduce an interface so it's clear they share functionality or perhaps a base wrapper to share logic eg.: `attributes()` has the same signature and vue wrapper don't have to use the dom wrapper as it has the same element getter.
 - [x] turn on strict options in tsconfig
 -  [x] go beyond the prettier's typescript rules. There are lot of extra [rules](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin#supported-rules) that promote consistency and better code quality that isn't utilised at the moment. eg: enforce consistent type imports

cexbrayat:
Hi @nandi95 

Sure, I think we all want that and just lacked the time to improve the codebase as we were trying to fix all the issues incoming and being ready for v2 stable release. You can definitely experiment with a shared interface and we'll see. Strict config has been on my personal todolist for a while. Turning it on in one single PR requires many changes, so I was fixing one file at a time and opening a PR with the changes. But I haven't been through all the code, so your help is definitely welcome here! Same for eslint! Thank you for offering your help.

nandi95:
Yes, I can help.
As for the eslint and strict options. What would be the best way to go about this?
One pull request per option/rule? Or one PR with separate commits per option/rule?

Have you got preferences on eslint rules? Or shall I just turn them on as I see fit, then they can be discussed in the PR?

cexbrayat:
I think separate PRs are a good way to go, so we can merge them as soon as they're reviewed.

And yes, turn the option as you see fit, or if you want to check first, you can add a list here and we can discuss. Whatever you prefer!

lmiller1990:
We had a shared interface at one point but it was removed - I don't remember why exactly but there was some complexity that either couldn't be solved or made the code much more complex. If you can find a way to have a shared interface, I think that's be fine. 

Personally I am more focused on actual big fixes but general improvements, as well as the quest for strict mode, are always welcome.

PR #372, opened 2021-02-10
Added base wrapper and wrapper like interface

nandi95:
linked #369


src/BaseWrapper.ts
```diff
@@ -0,0 +1,84 @@
...
+      return hasDisabledAttribute && elementCanBeDisabled
+    }
+
+    if (this.element && !isDisabled()) {
+      const event = createDOMEvent(eventString, options)
+      this.element.dispatchEvent(event)
+    }
+
+    await nextTick()
+    return
```
nandi95, line 82: This is an API change.

src/domWrapper.ts
```diff
@@ -142,9 +110,11 @@ export class DOMWrapper<ElementType extends Element> {
       if (tagName === 'SELECT') {
         return this.trigger('change')
       }
-      this.trigger('input')
+      await this.trigger('input')
       // trigger `change` for `v-model.lazy`
-      return this.trigger('change')
+      await this.trigger('change')
+
+      return
```
nandi95, line 117: This is an API change

nandi95:
On the vueWrapper if it wasn't for the Omit 'exists' on the get method, then that could also be added to the WrapperLike interface.

lmiller1990:
I think that Omit was intentional, maybe related: https://github.com/vuejs/vue-test-utils-next/issues/176

src/domWrapper.ts
lmiller1990: This is a breaking change - you *must* await `setValue` now.

Although this change is probably good, I am not sure we want to introduce a breaking change. Let's see what others thinks.

Our docs cover when and why to use `await` extensively, this would require updating those a lot, too.

lmiller1990 (changes requested):
I think the interface/base wrapper refactor seems fine, I am not sure on the breaking change.

Any chance we can remove the breaking change commit and discuss the elsewhere? Code improvements are great and always welcome, but any change to the API needs to be very carefully considered - we are in release candidate stage which generally means "no more breaking changes". 

We also have a pattern where we never wait for nextTick internally and leave that to the user (or another lib on top, like Testing Library), so this change would be in opposition to that.

src/BaseWrapper.ts
nandi95: Actually this is fine, given this was async before and not accepting a callbeck to be passed to the next tick

src/domWrapper.ts
nandi95: reverted

src/BaseWrapper.ts
nandi95: It isn't actually. next tick returns `Promise<void>`

src/domWrapper.ts
afontcu: 👍 Let's keep this reverted for now (so the PR does not introduce more changes than expected), but maybe making `setValue` async all the time is… a good thing? 🤔

src/interfaces/WrapperLike.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export default interface WrapperLike {
+  find<K extends keyof HTMLElementTagNameMap>(
+    selector: K
+  ): DOMWrapper<HTMLElementTagNameMap[K]>
+  find<K extends keyof SVGElementTagNameMap>(
+    selector: K
+  ): DOMWrapper<SVGElementTagNameMap[K]>
+  find<T extends Element>(selector: string): DOMWrapper<T>
+  find(selector: string): DOMWrapper<Element>
```
cexbrayat, line 11: I think there should be the `get` method as well here

cexbrayat (commented):
Looks good overall

Just 2 tiny feedback

cexbrayat:
Oh and another note: we usually have commit message convention inspired by the Angular one:
- prefix by `feat`, `fix`, `refactor`, `chore`, `docs`
- use lowercase
- use imperative tense

These 2 commits could be just one with a message `refactor: introduce base wrapper`

src/interfaces/WrapperLike.ts
nandi95: I'm not sure how to implement that given the Omit on the get see: https://github.com/vuejs/vue-test-utils-next/pull/372#issuecomment-776610237
I could have a second optional type argument but that felt messy.

src/errorWrapper.ts
```diff
@@ -1,6 +1,4 @@
-export function createWrapperError<T extends object>(
-  wrapperType: 'DOMWrapper' | 'VueWrapper'
-) {
+export function createWrapperError<T extends object>(wrapperType: string) {
```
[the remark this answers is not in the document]
nandi95: I'm nut sure why I did that.

src/interfaces/WrapperLike.ts
cexbrayat: Hmm I think the Omit should be in both signatures, as `get` throws if the element is not there, so calling `exists` on it has no sense. If that's not the case that's a mistake? Or am I missing something @lmiller1990 ?

lmiller1990: I think this is probably a mistake, yes

src/BaseWrapper.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  text() {
+    return textContent(this.element)
+  }
+
+  exists() {
+    return true
+  }
+
+  public async trigger(eventString: string, options?: TriggerOptions) {
```
lmiller1990, line 47: nit: any reason this one is marked public? I think public is the default

lmiller1990:
Thanks for reverting the API change, left one comment + @cexbrayat and we should be good... you are on fire 👍

lmiller1990 (approved):

lmiller1990:
Ahhh one more tiny thing - all the files in this code-base are camel case - can we follow this convention?

lmiller1990 (approved):

lmiller1990:
This is conflicted. I cannot merge it and I don't know why 🤔

Issue #373, opened 2021-02-10
Plugins one time setup


nandi95:
I'm working on a component library that sets some global config on the `globalProperties`.
At the moment I have set this config in every mount option so it's available in the component.

Is there currently a way to set up this global config 1 time before all tests?
I guess this also applies if I were to use lets say a vue plugin that adds a global directive and doesn't have to be imported in the components.

In the docs there's an example for the [router](https://vue-test-utils.vuejs.org/v2/guide/vue-router.html#with-a-real-router) and vuex using the `global.plugins` but that's still has to be set up at every mount. (altough those 2 are might meant to be specific to the tested component so they make more sense...)

afontcu:
Hi!

I think this was fixed on https://github.com/vuejs/vue-test-utils-next/pull/365

nandi95:
That is exactly what I was looking for. Thanks!
Side question does the config [global](https://github.com/vuejs/vue-test-utils-next/blob/master/src/config.ts#L7) has to be `Required<GlobalMountOptions>` ? can we not keep it partial?
@afontcu

afontcu:
>  Side question does the config [global](https://github.com/vuejs/vue-test-utils-next/blob/master/src/config.ts#L7) has to be `Required<GlobalMountOptions>` ? can we not keep it partial?

Yeah, I'd say we could (should?) keep it partial 🤔

PR #374, opened 2021-02-10
Add test to confirm config merger

nandi95:
linked #373 

I don't see why this should be required when trying to set up a single thing in the config. The user provided config is merged with the defaults: https://github.com/vuejs/vue-test-utils-next/blob/master/src/mount.ts#L322
Set it to `Partial` in case something will be required in the `GlobalMountOptions` down the line.

Edit:
This has stemmed from a confusion on how to use the the global config. Added a test to confirm how configs get merged.
+found and fixed an issue with renderStubDefaultSlot

### DIFF
--- a/tests/config.spec.ts
+++ b/tests/config.spec.ts
@@ -47,7 +47,7 @@ describe('config', () => {
       // mount config overrides default config
       expect(comp.find('#default-slot').exists()).toBe(true)
       // user defined config overrides default config
-      expect(comp.findComponent(Hello).exists()).toBe(true);
+      expect(comp.findComponent(Hello).exists()).toBe(true)
     })
   })
 
--- a/tests/config.spec.ts
+++ b/tests/config.spec.ts
@@ -65,6 +65,34 @@ describe('config', () => {
 
       expect(comp.find('#default-slot').exists()).toBe(true)
     })
+
+    it('should evaluate given option as boolean', () => {
+      // @ts-expect-error
+      let comp = mount(ComponentWithSlots, {
+        slots: {
+          default: '<div id="default-slot" />'
+        },
+        shallow: true,
+        global: {
+          renderStubDefaultSlot: 'truthy'
+        }
+      })
+
+      expect(comp.find('#default-slot').exists()).toBe(true)
+
+      // @ts-expect-error
+      let comp = mount(ComponentWithSlots, {
+        slots: {
+          default: '<div id="default-slot" />'
+        },
+        shallow: true,
+        global: {
+          renderStubDefaultSlot: 0
+        }
+      })
+
+      expect(comp.find('#default-slot').exists()).toBe(false)
+    })
   })
 
   describe('components', () => {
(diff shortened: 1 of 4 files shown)


cexbrayat:
I don't think this is publicly exposed, and @posva changed it to required to make sure we provide a default for every option internally. See #236

nandi95:
The type isn't the config is.
https://github.com/vuejs/vue-test-utils-next/blob/master/src/config.ts#L62
https://github.com/vuejs/vue-test-utils-next/blob/master/src/index.ts#L15
And I just realised, https://github.com/vuejs/vue-test-utils-next/blob/master/src/mount.ts#L322 actually merges the user provided config meaning that it is in fact should be `Required` type.

So I propose:
There's should be a `defaultConfig` provided by vtu, a `config` exposed to the user which is deepMerged into the `defaultConfig` then the config from the mount options merged into the existsing config.
Then the `config`'s type can be set to partial like in the PR

posva:
The type should not be changed to Partial because of the example in the linked pr by @cexbrayat 

If there is a use case, `defaultConfig` could be exposed with a proper type

nandi95:
I added a test to confirm the config merges as expected. I don't know if that tast is actualy valuable.
However, it is currentlry failing on this line:
`expect(comp.find('#default-slot').exists()).toBe(true)`
so I wrote a test for the `renderStubDefaultSlot` too.

Am I misunderstanding the renderStubDefaultSlot usage or that failure is valid?
@posva

tests/config.spec.ts
```diff
@@ -19,6 +20,51 @@ describe('config', () => {
...
+          renderStubDefaultSlot: true
+        }
+      })
+
+      // mount config overrides user defined config
+      expect(comp.vm.$.appContext.config.globalProperties.myProp).toBe(2)
+      // mount config overrides default config
+      expect(comp.find('#default-slot').exists()).toBe(true)
+      // user defined config overrides default config
+      expect(comp.vm.$.appContext.components['Hello']).not.toBeUndefined()
```
lmiller1990, line 50: Any reason to assert against what is otherwise internals? We could just render `<hello />` in the component and do `wrapper.findComponent(Hello)`?

lmiller1990:
I might have missed something, are you adding types (the PR only shows a test - more tests are always welcome if they are testing things we haven't got a test for)

nandi95:
Initially I updated the config's type as can be seen in the commits. This stemmed for a confusion about using the config. And now I just added a test to confirm what should happen with the config. More on this in my previous comment.

lmiller1990:
I see. The test is failing - will you be adding another commit(s) to fix this?

nandi95:
> Am I misunderstanding the renderStubDefaultSlot usage or that failure is valid?

Yes, once this is cleared up.

lmiller1990:
It is valid. This is how the defaultSlot option work: https://next.vue-test-utils.vuejs.org/guide/advanced/stubs-shallow-mount.html#default-slots-and-shallow

lmiller1990:
This seems fine now. @cexbrayat do you see anything you'd recommend changing or shall we merge and finally achieve the strict: true dream?

cexbrayat (approved):

PR #375, opened 2021-02-10
Turned on ts strict option

nandi95:
linked #369 
This PR introduces the ts strict option

There were places where I roughly guessed the type...
There are more non-null assertion `!` and `any` type is in there then one would hope for, but it's a start.

### DIFF
--- a/src/createDomEvent.ts
+++ b/src/createDomEvent.ts
@@ -1,3 +1,4 @@
+// @ts-expect-error No DefinitelyTyped package exists
 import eventTypes from 'dom-event-types'
 
 interface TriggerOptions {
@@ -9,7 +10,7 @@ interface TriggerOptions {
 
 interface EventParams {
   eventType: string
-  modifiers: string[]
+  modifiers: KeyNameArray
   options?: TriggerOptions
 }
 
@@ -21,8 +22,8 @@ const mouseKeyModifiers = ['left', 'middle', 'right']
 /**
  * Groups modifiers into lists
  */
-function generateModifiers(modifiers: string[], isOnClick: boolean) {
-  const keyModifiers: string[] = []
+function generateModifiers(modifiers: KeyNameArray, isOnClick: boolean) {
+  const keyModifiers: KeyNameArray = []
   const systemModifiers: string[] = []
 
   for (let i = 0; i < modifiers.length; i++) {
@@ -50,6 +51,7 @@ function generateModifiers(modifiers: string[], isOnClick: boolean) {
   }
 }
 
+export type KeyNameArray = Array<keyof typeof keyCodesByKeyName>
 export const keyCodesByKeyName = {
   backspace: 8,
   tab: 9,
@@ -98,10 +100,13 @@ function getEventProperties(eventParams: EventParams) {
 
   // convert `shift, ctrl` to `shiftKey, ctrlKey`
   // allows trigger('keydown.shift.ctrl.n') directly
-  const systemModifiersMeta = systemModifiers.reduce((all, key) => {
-    all[`${key}Key`] = true
-    return all
-  }, {})
+  const systemModifiersMeta = systemModifiers.reduce(
+    (all: Record<string, boolean>, key) => {
+      all[`${key}Key`] = true
+      return all
+    },
+    {}
+  )
 
   // get the keyCode for backwards compat
   const keyCode =
@@ -148,8 +153,12 @@ function createDOMEvent(eventString: String, options?: TriggerOptions) {
   // split eventString like `keydown.ctrl.shift.c` into `keydown` and array of modifiers
   const [eventType, ...modifiers] = eventString.split('.')
 
-  const eventParams: EventParams = { eventType, modifiers, options }
-  const event: Event = createEvent(eventParams)
+  const eventParams: EventParams = {
+    eventType,
+    modifiers: modifiers as KeyNameArray,
+    options
+  }
+  const event: Event & TriggerOptions = createEvent(eventParams)
   const eventPrototype = Object.getPrototypeOf(event)
 
   // attach custom options to the event, like `relatedTarget` and so on.
--- a/src/createDomEvent.ts
+++ b/src/createDomEvent.ts
@@ -1,58 +1,8 @@
-// @ts-ignore No DefinitelyTyped package exists
+// @ts-ignore todo - No DefinitelyTyped package exists for this
 import eventTypes from 'dom-event-types'
+import { Key } from 'readline'
 
-interface TriggerOptions {
-  code?: String
-  key?: String
-  keyCode?: Number
-  [custom: string]: any
-}
-
-interface EventParams {
-  eventType: string
-  modifiers: KeyNameArray
-  options?: TriggerOptions
-}
-
-// modifiers to keep an eye on
-const ignorableKeyModifiers = ['stop', 'prevent', 'self', 'exact']
-const systemKeyModifiers = ['ctrl', 'shift', 'alt', 'meta']
-const mouseKeyModifiers = ['left', 'middle', 'right']
-
-/**
- * Groups modifiers into lists
- */
-function generateModifiers(modifiers: KeyNameArray, isOnClick: boolean) {
-  const keyModifiers: KeyNameArray = []
-  const systemModifiers: string[] = []
-
-  for (let i = 0; i < modifiers.length; i++) {
-    const modifier = modifiers[i]
-
-    // addEventListener() options, e.g. .passive & .capture, that we dont need to handle
-    if (ignorableKeyModifiers.includes(modifier)) {
-      continue
-    }
-    // modifiers that require special conversion
-    // if passed a left/right key modifier with onClick, add it here as well.
-    if (
-      systemKeyModifiers.includes(modifier) ||
-      (mouseKeyModifiers.includes(modifier) && isOnClick)
-    ) {
-      systemModifiers.push(modifier)
-    } else {
-      keyModifiers.push(modifier)
-    }
-  }
-
-  return {
-    keyModifiers,
-    systemModifiers
-  }
-}
-
-export type KeyNameArray = Array<keyof typeof keyCodesByKeyName>
-export const keyCodesByKeyName = {
+const keyCodesByKeyName = {
   backspace: 8,
   tab: 9,
   enter: 13,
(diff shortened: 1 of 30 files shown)


nandi95:
#372 Will probably resolve the only typing issue left in the src dir. (the `attributes()`, `classes()` methods)

src/emit.ts
```diff
@@ -1,4 +1,5 @@
 import { setDevtoolsHook, devtools, ComponentPublicInstance } from 'vue'
+import { ComponentInternalInstance } from '@vue/runtime-core'
```
lmiller1990, line 2: Is this not exported from `'vue'`?

src/emit.ts
```diff
@@ -10,43 +11,44 @@ let events: Record<string, typeof componentEvents>
...
-    const emitted = vmEvents
-      ? (vmEvents as Record<string, T[]>)[eventName]
-      : undefined
-    return emitted
+    return vmEvents ? (vmEvents as Record<string, T[]>)[eventName] : undefined
   }
 
   return vmEvents as Record<string, T[]>
 }
+type Events = { [id: number]: Record<string, any> }
```
lmiller1990, line 24: Not sure on the best practice here but this could be `<Record<number, Record<string, any>>`?

Also, perhaps this should actually be `unknown`?

src/emit.ts
```diff
@@ -10,43 +11,44 @@ let events: Record<string, typeof componentEvents>
...
 
 export const attachEmitListener = () => {
   events = {}
   // use devtools to capture this "emit"
   setDevtoolsHook(createDevTools(events))
 }
 
-function createDevTools(events): any {
-  const devTools: Partial<typeof devtools> = {
+function createDevTools(events: Events): any {
```
lmiller1990, line 32: instead of `as Partial` should we just change `: any` to `: Partial<typeof devtools>`

tests/emit.spec.ts
```diff
@@ -93,7 +93,7 @@ describe('emitted', () => {
       name: 'Parent',
       setup(props, { emit }) {
         return () =>
-          h(Child, { onHello: (...events) => emit('parent', ...events) })
+          h(Child, { onHello: (...events: any[]) => emit('parent', ...events) })
```
lmiller1990, line 96: I think unknown is more correct here (seems that is how it is typed in vue/core)

tests/features/async-components.spec.ts
```diff
@@ -1,17 +1,17 @@
 import { defineAsyncComponent, defineComponent, h, AppConfig } from 'vue'
 
 import { mount, flushPromises } from '../../src'
+import { ComponentPublicInstance } from '@vue/runtime-core'
```
lmiller1990, line 4: not exported from `'vue'`? Otherwise we need to add `@vue/runtime-core` as a peer dependency.

tests/features/async-components.spec.ts
```diff
@@ -1,17 +1,17 @@
...
 
 const config: AppConfig = {
   optionMergeStrategies: {},
   globalProperties: {},
   isCustomElement: (tag: string) => false,
   performance: false,
-  errorHandler: (error: Error) => {
-    if (error.message.match(/Async component failed to load./)) {
+  errorHandler: (error: unknown) => {
```
lmiller1990, line 11: How come `Error` is not the correct type here? I assume `errorHandler` always receives something extending `Error`?

tests/isVisible.spec.ts
```diff
@@ -93,9 +93,11 @@ describe('isVisible', () => {
       `,
       methods: {
         add() {
+          // @ts-expect-error
```
lmiller1990, line 96: nit: this is probably some way to type this if you use DefineComponent

tests/mountingOptions/mocks.spec.ts
```diff
@@ -37,15 +37,18 @@ describe('mocks', () => {
         </div>
       `,
       computed: {
```
lmiller1990, line 39: Does anyone know the story with Options API? I wonder what the practice is here - maybe `(this.$router as RouterRoute)` (or whatever the interface is)

I don't care too much about the typing on the unit tests either way, mainly the code in `src`... so don't worry about those ones too much, just interested

lmiller1990 (commented):
Added some comments, glad we are going to have strict mode!

tests/features/async-components.spec.ts
nandi95: indeed it is, Just my IDE defaulted to this,
`vue` has: `export * from "@vue/runtime-dom";`
fixed

src/emit.ts
nandi95: unfortunatly not, `setDevtoolsHook` expects a full [DevtoolsHook](https://github.com/vuejs/vue-next/blob/3626ff07fe5107080c52e85018070562c84b796e/packages/runtime-core/src/devtools.ts#L21) implementation which partial doesn't cover. I'm not sure what's the logic should be for the rest. I can add stub implementations.

tests/features/async-components.spec.ts
nandi95: I would have thought so but it seems to be [typed to unkown](https://github.com/vuejs/vue-next/blob/6d5b62351248780663d2612a1f483f7ea9f5e5a2/packages/runtime-core/src/apiCreateApp.ts#L58). This may well could be an oversight in the vue next repo

tests/isVisible.spec.ts
nandi95: you mean, let's rewrite this test component to composition api?

tests/features/async-components.spec.ts
cexbrayat: This is `unknown` on purpose, it changed during the alphas (IIRC JS can throw other things than error, hence the unknown type)

tests/mountingOptions/mocks.spec.ts
cexbrayat: I would say `defineCOmponent` + somthing like : 
```ts
declare module '@vue/runtime-core' {
   interface ComponentCustomProperties {
    $router: Router
  }
}
```

tests/setData.spec.ts
```diff
@@ -34,6 +34,7 @@ describe('setData', () => {
         myObject: {
           immediate: true,
           handler() {
+            // @ts-expect-error
```
cexbrayat, line 37: same here `defineComponent` may help

tests/setProps.spec.ts
```diff
@@ -65,6 +65,7 @@ describe('setProps', () => {
       },
       watch: {
         foo(val: string) {
+          // @ts-expect-error
           this.bar = val
```
cexbrayat, line 69: same here

tests/trigger.spec.ts
```diff
@@ -233,7 +233,8 @@ describe('trigger', () => {
       const wrapper = mount(Component, {})
 
       for (const keyName in keyCodesByKeyName) {
-        const keyCode = keyCodesByKeyName[keyName]
+        const keyCode =
+          keyCodesByKeyName[keyName as keyof typeof keyCodesByKeyName]
```
cexbrayat, line 237: hard to read 😅 ain't that just `number`?

cexbrayat (commented):
Nice!

I added some comments (a lot, sorry!), but this is overall a good step in the right direction

src/createDomEvent.ts
```diff
@@ -50,6 +51,7 @@ function generateModifiers(modifiers: string[], isOnClick: boolean) {
   }
 }
 
+export type KeyNameArray = Array<keyof typeof keyCodesByKeyName>
```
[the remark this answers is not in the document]
nandi95: Actually like `Array<KeyName>` better

src/emit.ts
```diff
@@ -1,27 +1,30 @@
...
 
-let componentEvents: Record<string, unknown[]>
-let events: Record<string, typeof componentEvents>
+let events: Events
 
 export function emitted<T = unknown>(
   vm: ComponentPublicInstance,
   eventName?: string
-): T[] | Record<string, T[]> {
+): undefined | Events<T>[number][string] | Events<T>[number] {
```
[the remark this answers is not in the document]
nandi95: > harder to read

I agree and you have to "hunt down" the type somewhat.
But I think an argument can be made that this way there is a single source of truth?

src/vueWrapper.ts
```diff
@@ -57,22 +62,26 @@ export class VueWrapper<T extends ComponentPublicInstance> {
   classes(): string[]
   classes(className: string): boolean
   classes(className?: string): string[] | boolean {
-    return new DOMWrapper(this.element).classes(className)
+    return new DOMWrapper(this.element).classes(className!)
```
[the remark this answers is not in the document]
nandi95: Yes indeed, I'm not sure why it isn't accepting it given as you said the signature is the same.
At the same this this will not be relevant when #375 gets merged.
Regardless in the interim I'll update it to the conditional.

src/emit.ts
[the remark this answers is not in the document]
cexbrayat: Well it should not really matter if the code is properly typed, if we change it in one place, the rest will not compile. So I would still go with the simpler one 😃

src/createDomEvent.ts
```diff
@@ -50,6 +51,7 @@ function generateModifiers(modifiers: string[], isOnClick: boolean) {
...
 export const keyCodesByKeyName = {
```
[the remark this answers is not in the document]
nandi95: don't see why not

tests/emit.spec.ts
```diff
@@ -130,10 +132,10 @@ describe('emitted', () => {
     expect(wrapper.emitted('hello')).toEqual(undefined)
 
     wrapper.find('button').trigger('click')
-    expect(wrapper.emitted('hello')[0]).toEqual(['foo', 'bar'])
+    expect((wrapper.emitted('hello') as unknown[])[0]).toEqual(['foo', 'bar'])
```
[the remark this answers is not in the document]
nandi95: The undefined outcome still remains if you give it a type

tests/mountingOptions/data.spec.ts
```diff
@@ -11,7 +11,8 @@ describe('mounting options: data', () => {
           bar: 'bar'
         }
       },
-      render() {
+      render(): Function {
+        // @ts-expect-error
```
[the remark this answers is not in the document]
nandi95: nice

tests/trigger.spec.ts
nandi95: not quite, it's a string, union of keys of the `keyCodesByKeyName` to be exact
Now I imported the `KeyName` type instead.

tests/mountingOptions/mocks.spec.ts
nandi95: `defineComponent()` and `mount()`'s typing is actual magic 🔥

lmiller1990 (approved):
Looks good to me, we can likely continue improving the typing as time goes by

PR #376, opened 2021-02-10
fix: remove the browser entry from package.json

elevatebart:
When this entry is in, webpack never picks up on the module one,
which makes it go the wrong way. Since VTU is meant to be used in
Jest/Webpack/Rollup/Vite we shuold avoid breaking those.

I matched the exports of vue and vite here.


lmiller1990 (approved):

lmiller1990:
Thanks

PR #382, opened 2021-02-12
Add .editorconfig

nandi95:
linked #369 

These are rules just something I worked out on a package I maintain but are very much subject to discussion.
Essentially turning on all rules, then turning them off or adjusting them where needed.
Given the above also consider whether prettier is needed or not?

Edits by maintainers are welcome.

When we got a set rules agreed I can update the project to pass linting.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+root = true
+
+[*]
+charset = utf-8
+insert_final_newline = true
+end_of_line = lf
+indent_style = space
+indent_size = 2


package.json
```diff
@@ -56,19 +63,17 @@
...
+    "lint": "eslint . --ext .ts",
+    "lint:fix": "yarn lint -- --fix",
     "docs:dev": "vitepress dev docs",
     "docs:build": "vitepress build docs"
   },
   "husky": {
-    "hooks": {
-      "pre-commit": "lint-staged"
-    }
+    "hooks": {}
```
nandi95, line 72: Temporarily until  linting has been resolved

.eslintrc.js
```diff
@@ -0,0 +1,84 @@
...
+  },
+    ignorePatterns: [
+        "node_modules",
+        "**/.eslintrc.js",
+    ],
+    extends: [
+    "eslint:recommended",
+    "plugin:@typescript-eslint/all",
+    "prettier/@typescript-eslint",
+    "plugin:prettier/recommended"
```
nandi95, line 20: Note the order of these, I think the latter overwrites the former entries (if I remember correctly)

.eslintrc.js
```diff
@@ -0,0 +1,84 @@
...
+      "keyword-spacing": "off",
+      "object-curly-spacing": ["warn", "always"],
+      "comma-dangle": ["warn", "never"],
+      "max-len": ["warn", 120],
+      "eqeqeq": "error",
+
+      // https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin#supported-rules
+      "@typescript-eslint/object-curly-spacing": ["warn", "always"],
+      "@typescript-eslint/quotes": ["warn", "single"],
+      "@typescript-eslint/indent": ["warn", 4],
```
afontcu, line 42: AFAICT we've been using an indent of 2 spaces. Could we keep it that way?

.editorconfig
```diff
@@ -0,0 +1,8 @@
+root = true
+
+[*]
+charset = utf-8
+insert_final_newline = true
+end_of_line = lf
+indent_style = space
+indent_size = 4
```
cexbrayat, line 8: 4 -> 2?

.eslintrc.js
```diff
@@ -0,0 +1,84 @@
+module.exports = {
+  parser: "@typescript-eslint/parser",
+  parserOptions: {
+    tsconfigDir: __dirname,
+    project: "./tsconfig.json",
```
cexbrayat, line 5: are these options needed when the tsconfig is at the root of the project. i would have thought that's the default values.

nandi95: Unfortunately yes otherwise you get an error like:
 > `Error: Error while loading rule '@typescript-eslint/prefer-nullish-coalescing': You have used a rule which requires parserServices to be generated. You must therefore provide a value for the "parserOptions.project" property for @typescript-eslint/parser.
Occurred while linting /Users/....../Web/vue-test-utils-next/src/vueWrapper.ts`

lmiller1990:
That config file is huge! What do you think about keeping it simple and just using a bare-bones one like [vue-next](https://github.com/vuejs/vue-next/blob/master/.prettierrc)?

nandi95:
The linked issue says: 
> go beyond the prettier's typescript rules. There are lot of extra rules that promote consistency and better code quality that isn't utilised at the moment. eg: enforce consistent type imports

hence all these rules are adjusted after the `plugin:@typescript-eslint/all`
If aiming for a minimal config file, I think what was already there (`.prettierrc`) was fine.
In that case should I revert this back? Or I could leave it just remove the rules so when someone wants to adjust something, there's already a place for it.
Regardless I think at least the `.editorconfig` is an improvment.

lmiller1990:
Yeah, if that's okay. `.editorconfig` is fine to have!

PR #393, opened 2021-02-16
Bug: Emit doesn't update 2 way bound props

nandi95:


tests/props.spec.ts
```diff
@@ -82,4 +83,32 @@ describe('props', () => {
       object: {}
     })
   })
+
+  it('should return the updated props on 2 way binding', async () => {
```
nandi95, line 87: This file is probbly the wrong place for this test, please advise where it would be best.

src/emit.ts
```diff
@@ -48,7 +48,7 @@ function recordEvent(
   events: Events,
   vm: ComponentInternalInstance,
   event: string,
-  args: Events[number]
+  args: unknown[]
```
nandi95, line 51: This I typed wrong in the previous PR, fixed here

src/emit.ts
```diff
@@ -64,4 +64,16 @@ function recordEvent(
 
   // Record the event message sent by the emit
   events[cid][event].push(args)
+
+  if (event.startsWith('update:')) {
+    if (args.length !== 1) {
+      throw new Error(
+        'Two-way bound properties have to emit a single value.' +
```
nandi95, line 71:
```suggestion
        'Two-way bound properties have to emit a single value. ' +
```

lmiller1990 (approved):
I am somewhat surprised this works. We had some discussion around v-model testing, I think this is pretty good.

https://github.com/vuejs/vue-test-utils-next/discussions/279

tests/props.spec.ts
lmiller1990: Here is probably fine. Probably could re-think the test layout at some point.

lmiller1990:
We may revisit and revert this @nandi95. Based on https://github.com/vuejs/vue-test-utils-next/issues/440#issuecomment-790538633

In that case, you'd need to just change to use a `vModel` mounting option.

PR #394, opened 2021-02-17
fix: record native events against each wrapper they bubble through

### LINKED ISSUE #391: `input.trigger('compositionstart')` not work
The code works in `rc0`, but fails in `rc1`.

```ts
  test('v-model work', async () => {
    const valueRef = ref('init value')
    const wrapper = mount({
      components: { IxInput },
      template: `<ix-input v-model:value="valueRef" />`,
      setup() {
        return { valueRef }
      },
    })
    const input = wrapper.find('input')

    expect(input.element.value).toBe('init value')

    await input.setValue('input setValue')

    expect(valueRef.value).toBe('input setValue')

    valueRef.value = 'valueRef change'

    await nextTick()

    expect(input.element.value).toBe('valueRef change')

    input.element.value = '使用拼音'
    await input.trigger('compositionstart')

    expect(wrapper.emitted()).toHaveProperty('compositionStart')
    expect(valueRef.value).toBe('valueRef change')

    await input.trigger('input')

    expect(wrapper.emitted()).toHaveProperty('input')
    expect(valueRef.value).toBe('valueRef change')

    await input.trigger('compositionend')

    expect(wrapper.emitted()).toHaveProperty('compositionEnd')
    expect(valueRef.value).toBe('使用拼音')
  })
```

repo: https://github.com/IduxFE/idux

aethr:
fix #391 

This is pretty ugly on a few levels and I think it needs some cleaning up, but I think it addresses the issue. Maybe going back to the old mixin is a better idea for native events, I think I'll leave that up to you because you know the history a bit better.

I added a test that covers I think a basic exploration of the problem being solved.

Again, I'm an absolute TS novice, :pray: for me.

Looking at this code again, do we need to add a way to clear events similar to jest's `mockClear()`?

### DIFF
--- a/src/vueWrapper.ts
+++ b/src/vueWrapper.ts
@@ -12,7 +12,7 @@ import { createWrapperError } from './errorWrapper'
 import { TriggerOptions } from './createDomEvent'
 import { find, matches } from './utils/find'
 import { mergeDeep, textContent } from './utils'
-import { emitted } from './emit'
+import { emitted, recordEvent } from './emit'
 
 export class VueWrapper<T extends ComponentPublicInstance> {
   private componentVM: T
@@ -30,6 +30,9 @@ export class VueWrapper<T extends ComponentPublicInstance> {
     this.rootVM = vm?.$root
     this.componentVM = vm as T
     this.__setProps = setProps
+
+    this.attachNativeEventListener()
+
     config.plugins.VueWrapper.extend(this)
   }
 
@@ -40,6 +43,21 @@ export class VueWrapper<T extends ComponentPublicInstance> {
 
   private get parentElement(): VueElement {
     return this.vm.$el.parentElement
+  }
+
+  private attachNativeEventListener(): void {
+    const vm = this.vm
+    if (!vm) return
+
+    const element = this.element
+    for (let key in element) {
+      if (/^on/.test(key)) {
+        const eventName = key.slice(2)
+        element.addEventListener(eventName, (...args) => {
+          recordEvent(vm.$, eventName, args)
+        })
+      }
+    }
   }
 
   get element(): Element {
@@ -79,7 +97,7 @@ export class VueWrapper<T extends ComponentPublicInstance> {
   }
 
   emitted<T = unknown>(): Record<string, T[]>
-  emitted<T = unknown>(eventName?: string): undefined | T[]
+  emitted<T = unknown>(eventName: string): undefined | T[]
   emitted<T = unknown>(
     eventName?: string
   ): undefined | T[] | Record<string, T[]> {
--- a/src/vueWrapper.ts
+++ b/src/vueWrapper.ts
@@ -1,5 +1,7 @@
 import { ComponentPublicInstance, nextTick, App } from 'vue'
 import { ShapeFlags } from '@vue/shared'
+// @ts-ignore todo - No DefinitelyTyped package exists for this
+import eventTypes from 'dom-event-types'
 
 import { config } from './config'
 import { DOMWrapper } from './domWrapper'
@@ -50,13 +52,10 @@ export class VueWrapper<T extends ComponentPublicInstance> {
     if (!vm) return
 
     const element = this.element
-    for (let key in element) {
-      if (/^on/.test(key)) {
-        const eventName = key.slice(2)
-        element.addEventListener(eventName, (...args) => {
-          recordEvent(vm.$, eventName, args)
-        })
-      }
+    for (let eventName of Object.keys(eventTypes)) {
+      element.addEventListener(eventName, (...args) => {
+        recordEvent(vm.$, eventName, args)
+      })
     }
   }
 
--- a/tests/emit.spec.ts
+++ b/tests/emit.spec.ts
@@ -1,4 +1,10 @@
-import { defineComponent, FunctionalComponent, h, SetupContext } from 'vue'
+import {
+  defineComponent,
+  FunctionalComponent,
+  getCurrentInstance,
+  h,
+  SetupContext
+} from 'vue'
 import { Vue } from 'vue-class-component'
 
 import { mount } from '../src'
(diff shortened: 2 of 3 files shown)


src/vueWrapper.ts
```diff
@@ -30,6 +30,9 @@ export class VueWrapper<T extends ComponentPublicInstance> {
     this.rootVM = vm?.$root
     this.componentVM = vm as T
     this.__setProps = setProps
+
+    this.attachNativeEventListener()
+
```
aethr, line 35: Perhaps this method should be exported from `emit.ts` with `this` as an argument, the style looks very different from anything else so I tried to add a private method instead. There is definitely some bleeding of concerns.

src/vueWrapper.ts
```diff
@@ -42,6 +45,21 @@ export class VueWrapper<T extends ComponentPublicInstance> {
...
+    if (!vm) return
+
+    const element = this.element
+    for (let key in element) {
+      if (/^on/.test(key)) {
+        const eventName = key.slice(2)
+        element.addEventListener(eventName, (...args) => {
+          recordEvent(vm.$, eventName, args)
+        })
+      }
```
aethr, line 59: Pretty ugly way to add event listeners for all possible events, but it works.

Do these event listeners need to be removed on cleanup?

src/vueWrapper.ts
```diff
@@ -42,6 +45,21 @@ export class VueWrapper<T extends ComponentPublicInstance> {
     return this.vm.$el.parentElement
   }
 
+  private attachNativeEventListener(): void {
+    const vm = this.vm
+    if (!vm) return
```
aethr, line 50: Calling `this.element` breaks in some circumstances that I had trouble debugging, for some reason `findComponent` creates Wrapper objects without a vm? Anyway, if there's no vm, I guess there's nothing to bind event listeners to.

lmiller1990: I cannot imagine why this would happen 🤔  I will look at it

src/vueWrapper.ts
lmiller1990: Probably not, I am guessing when the element is cleaned up the listeners are as well

src/vueWrapper.ts
lmiller1990: Oh it makes sense, functional components can be found with `findComponent` but they do not have a `vm`

most likely someone is going to ask about capturing emitted events from functional components. let's deal with it if and when it comes up 🤷

lmiller1990 (approved):
Seems good!

lmiller1990:
Seems we need more investigation - I ran #391 against this PR and it does not pass, I get a weird output:

```
● Input.vue › v-model work

    expect(received).toHaveProperty(path)

    Expected path: "compositionStart"
    Received path: []

    Received value: {"change": [[{"isTrusted": false}]], "input": [[{"isTrusted": false}]]}

      46 |     await input.trigger('compositionstart')
      47 |
    > 48 |     expect(wrapper.emitted()).toHaveProperty('compositionStart')
         |                               ^
      49 |     expect(valueRef.value).toBe('valueRef change')
      50 |
      51 |     await input.trigger('input')

      at Object.<anonymous> (packages/components/input/__tests__/input.spec.ts:48:31)

```

lmiller1990:
I can fix this - I cannot push to your branch. Are you able to enable the setting "let maintainers push commits" (not sure the exact name of the setting).

src/vueWrapper.ts
lmiller1990: We should use 

```ts

// @ts-ignore todo - No DefinitelyTyped package exists for this
import eventTypes from 'dom-event-types'

// ... 
    const element = this.element
    for (let eventName of Object.keys(eventTypes)) {
      element.addEventListener(eventName, (...args) => {
        recordEvent(vm.$, eventName, args)
      })
    }
```

This was is missing some events like compositionstart, and also seems to work differently in FF, chrome, jsdom etc.

tests/emit.spec.ts
```diff
@@ -78,23 +78,73 @@ describe('emitted', () => {
...
+      emits: ['hi'],
+      setup(props, { emit }) {
+        return () =>
+          h('div', [h('button', { onClick: () => emit('hi', 'foo', 'bar') })])
+      }
+    })
+
     const Parent = defineComponent({
       name: 'Parent',
+      emits: ['hello'],
```
lmiller1990, line 143: Please add this test

It's the one the OP of the original issue had problems with

```ts
  it('captures composition event', async () => {
    const useCommonBindings = () => {
      const onCompositionStart = (evt: CompositionEvent) => {
        const instance = getCurrentInstance()!
        instance.emit('compositionStart', evt)
      }
      return { onCompositionStart }
    }
    const IxInput = defineComponent({
      setup() {
        return useCommonBindings()
      },
      template: `<input @compositionstart="(evt) => $emit('compositionStart', evt)" />`
    })

    const wrapper = mount({
      components: { IxInput },
      template: `<ix-input />`
    })

    await wrapper.find('input').trigger('compositionstart')

    expect(wrapper.emitted().compositionstart).not.toBe(undefined)
  })
```

aethr:
Sorry @lmiller1990 I don't see the "Allow maintainers..." option on the PR. Apparently it won't work on this PR because it's underneath an Organisation (https://github.com/isaacs/github/issues/1681), next time I'll make sure it's a user repo.

I'll make the changes you requested, and if you want to keep riffing on it we can figure out another solution. Maybe I'll re-create it as a user fork.

tests/emit.spec.ts
aethr: Added in 1ba4e59

src/vueWrapper.ts
aethr: The test I added has a functional component for the grandparent, which works because it captures events that bubbled to the element of the out wrapper, I suppose. \it might be different for functional components without their own native DOM element but I suppose in that case DOM events wouldn't expect to be captured?

lmiller1990:
Should be fine, I will give this one last test and merge it. If we need to make any tweaks we still have a few release candidates up our sleeve :) thanks for this!

aethr:
Sounds good! Thanks for your help getting it over the line.

PR #401, opened 2021-02-19
docs: Cleanup and update API docs

afontcu:
- [x] Add Signatures (TS types). I wasn't 100% sure about where to draw the line. Any feedback is appreciated.
- [x] Add missing docs (shallowMount, flushPromises…).
- [x] Make API docs more consistent.
- [x] Fix links on `MountingOptions`.
- [x] Ensure API page only shows sidebar for API methods (otherwise is way too long).

### DIFF
--- a/docs/.vitepress/config.js
+++ b/docs/.vitepress/config.js
@@ -2,7 +2,6 @@
 
 /** @type {UserConfig} */
 const config = {
-  // base: '/v2/',
   title: 'Vue Test Utils for Vue 3',
   description: 'The documentation for the official Vue Test Utils',
   locales: {
--- a/docs/.vitepress/config.js
+++ b/docs/.vitepress/config.js
@@ -8,7 +8,7 @@ const config = {
   locales: {
     '/': {
       lang: 'en-US',
-      title: 'Vue Test Utils (2.0.0-beta.14)'
+      title: 'Vue Test Utils (2.0.0-rc.0)'
     }
   },
   head: [['link', { rel: 'icon', href: `/logo.png` }]],
--- a/docs/.vitepress/config.js
+++ b/docs/.vitepress/config.js
@@ -1,13 +1,14 @@
+const package = require('../../package.json')
 /** @typedef {import('vitepress').UserConfig} UserConfig */
 
 /** @type {UserConfig} */
 const config = {
-  title: 'Vue Test Utils for Vue 3',
-  description: 'The documentation for the official Vue Test Utils',
+  title: `Vue Test Utils for Vue 3 (${package.version})`,
+  description: 'The documentation for the official Vue Test Utils for Vue 3',
   locales: {
     '/': {
       lang: 'en-US',
-      title: 'Vue Test Utils (2.0.0-rc.0)'
+      title: `Vue Test Utils for Vue 3 (${package.version})`
     }
   },
   head: [['link', { rel: 'icon', href: `/logo.png` }]],
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -1628,10 +1628,9 @@ interface MountingOptions<Props, Data = {}> {
   props?: (RawProps & Props) | ({} extends Props ? null : never)
   slots?: { [key: string]: Slot } & { default?: Slot }
   global?: GlobalMountOptions
-  shallow?: boolean
 }
 
-function mount(Component, options?: MountingOptions): VueWrapper
+function shallowMount(Component, options?: MountingOptions): VueWrapper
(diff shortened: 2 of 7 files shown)


docs/api/index.md
```diff
@@ -1088,18 +1476,21 @@ test('updates component data', async () => {
 })
 ```
 
-Notice that `setData` does not allow setting new properties that are not
-defined in the component.
-
-Also, notice that `setData` does not modify composition API `setup()` data.
+::: tip
```
nandi95, line 1479:
```suggestion
::: warning
```

docs/api/index.md
```diff
@@ -1132,108 +1527,155 @@ test('updates prop', async () => {
 })
 ```
 
-### `setValue`
+::: tip
```
nandi95, line 1530:
```suggestion
::: warning
```

docs/api/index.md
```diff
@@ -1132,108 +1527,155 @@ test('updates prop', async () => {
...
+test('setValue on input text', () => {
+  const wrapper = mount(Component)
+
+  await wrapper.find('input[type="text"]').setValue('hello!')
+  expect(wrapper.find('p').text()).toBe('Text: hello!')
+})
 ```
 
-### `text`
+::: tip
```
nandi95, line 1598:
```suggestion
::: warning
```

docs/api/index.md
```diff
@@ -1249,71 +1691,179 @@ Note that `trigger` accepts a second argument to pass options to the triggered E
 await wrapper.trigger('keydown', { keyCode: 65 })
 ```
 
-### `unmount`
+::: tip
```
nandi95, line 1694:
```suggestion
::: warning
```

docs/api/index.md
```diff
@@ -1249,71 +1691,179 @@ Note that `trigger` accepts a second argument to pass options to the triggered E
...
-## Global Config
+```ts
+vm: ComponentPublicInstance
+```
 
-### `config.global`
+**Details:**
 
-Instead of configuring global mounting options on a per-test basis, you can configure them globally. These will be used by default every time you `mount` a component. You can override the defaults by via mounting options.
+The `Vue` app instance. You can access all of the [instance methods](https://v3.vuejs.org/api/instance-methods.html) and [instance properties](https://v3.vuejs.org/api/instance-properties.html).
```
nandi95, line 1755: Like with the props section, should there be a warning here saying test against dom elements whenever is possible eg click on them and see the result, as opposed to call a method on the `vm`

docs/guide/advanced/async-suspense.md
```diff
@@ -39,11 +41,15 @@ test('increments by 1', () => {
 })
 ```
 
-Surprisingly, this fails! The reason is although `count` is increased, Vue will not update the DOM until the next "tick" or "render cycle". For this reason, the assertion will be called before Vue updates the DOM. This has to do with the concept of "macrotasks", "microtasks" and the JavaScript Event Loop. You can read more details and see a simple example [here](https://javascript.info/event-loop#macrotasks-and-microtasks).
+Surprisingly, this fails! The reason is although `count` is increased, Vue will not update the DOM until the next event loop "tick. For this reason, the assertion (`expect()...`) will be called before Vue updates the DOM.
```
nandi95, line 44: `event loop "tick.`
as in `ticks` or `aka tick` or `/next tick`.?

nandi95 (commented):
I was bored so just reviewed for the lulz
¯\\_(ツ)_/¯

docs/api/index.md
afontcu: sure! done. nice suggestion

docs/.vitepress/config.js
```diff
@@ -2,13 +2,12 @@
...
 /** @type {UserConfig} */
 const config = {
-  // base: '/v2/',
   title: 'Vue Test Utils for Vue 3',
   description: 'The documentation for the official Vue Test Utils',
   locales: {
     '/': {
       lang: 'en-US',
-      title: 'Vue Test Utils (2.0.0-beta.14)'
+      title: 'Vue Test Utils (2.0.0-rc.0)'
```
lmiller1990, line 10: We could actually import `package.json` here to make sure this is always up to date

lmiller1990 (approved):
Left one comment - up to you if you'd like to make this change.

I will leave merging this to you when you are happy 👍  thanks!

docs/.vitepress/config.js
afontcu: sure, done!

PR #408, opened 2021-02-22
fix: Handle exists() after unmounting

afontcu:
I never use it, but I believe `exists()` should return false after unmounting the component.

(_disclaimer: still trying to find the best signaling that the element is gone_)

### DIFF
--- a/src/vueWrapper.ts
+++ b/src/vueWrapper.ts
@@ -87,6 +87,11 @@ export class VueWrapper<T extends ComponentPublicInstance>
     return emitted(this.vm, eventName)
   }
 
+  exists(): boolean {
+    if (!this.parentElement) return false
+    return true
+  }
+
   html() {
     // cover cases like <Suspense>, multiple root nodes.
     if (this.parentElement['__vue_app__']) {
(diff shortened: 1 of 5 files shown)


src/vueWrapper.ts
```diff
@@ -87,6 +87,11 @@ export class VueWrapper<T extends ComponentPublicInstance>
     return emitted(this.vm, eventName)
   }
 
+  exists(): boolean {
+    if (!this.parentElement) return false
+    return true
```
lmiller1990, line 92: You could just do `return !!this.parentElement`

lmiller1990 (approved):
One comment!

src/vueWrapper.ts
afontcu: yeah, actually some tests related to functional components were failing, and I opened up the PR to make sure it wasn't a local issue – [turns out they actually fail](https://github.com/vuejs/vue-test-utils-next/runs/1955184111?check_suite_focus=true#step:6:142)

lmiller1990: These failures are introduced by this PR though, right? Do we just need a check somewhere?

afontcu: yeah! I'll take a look soon enough, looks like functional components don't have `this.vm` available.

afontcu: okay, gave it a go. Looks like this is not a failure introduced by this PR, but rather an issue with using `findComponent()` on a functional component:

```js
const Foo = () => h('div')

const wrapper = mount(Foo)

console.log(wrapper.findComponent(Foo))
/*
yields

    VueWrapper {
      wrapperElement: undefined,
      __app: null,
      rootVM: undefined,
      componentVM: null,
      __setProps: undefined
    }

*/
```

and then any method that relies on `vm` or similar fails (being that `exists()` or `text()` or…)

lmiller1990: This (might) be an expected limitation... no instance, so `findComponent` won't work? This is very confusing, maybe open an issue and we can try and figure something out?

Yet another problems with testing implementation details 🤷‍♂️ either this is definitely worth investigating/fixing.

PR #437, opened 2021-03-03
fix: attach emit listener after createApp is called

### LINKED ISSUE #436: emitted() not catching component events
Using `vue@3.0.2` and `@vue/test-utils@2.0.0-beta.8`, the following works fine.

```ts

const Foo = defineComponent({
  name: 'Foo',
  emits: ['foo'],
  setup (_, ctx) {
    return () => h('div', {
      onClick: () => {
        ctx.emit('foo', 'bar')
      },
    }, 'hello world')
  }
})

test('uses mounts', async () => {
  const wrapper = mount(Foo)

  console.log(wrapper.emitted())

  expect(wrapper.emitted('foo')).toHaveLength(1)
})
```

```bash
PASS  src/App.spec.ts
  ✓ uses mounts (21 ms)

  console.log
    { foo: [ [ 'bar' ] ] }

      at src/App.spec.ts:21:11

Test Suites: 1 passed, 1 total
Tests:       1 passed, 1 total
Snapshots:   0 total
Time:        2.29 s
Ran all test suites.
```

But using `vue@3.0.7` and `@vue/test-utils@2.0.0-rc.2` results in this

```bash
 FAIL  src/App.spec.ts
  ✕ uses mounts (18 ms)

  ● uses mounts

    expect(received).toHaveLength(expected)

    Matcher error: received value must have a length property whose value must be a number

    Received has value: undefined

      21 |   console.log(wrapper.emitted())
      22 |
    > 23 |   expect(wrapper.emitted('foo')).toHaveLength(1)
         |                                  ^
      24 | })
      25 |

      at src/App.spec.ts:23:34
      at fulfilled (src/App.spec.ts:5:58)

  console.log
    { click: [ [ [MouseEvent] ] ] }

      at src/App.spec.ts:21:11

Test Suites: 1 failed, 1 total
Tests:       1 failed, 1 total
Snapshots:   0 total
Time:        2.307 s
Ran all test suites.
```

A cursory investigation seems to point to the fact that this call https://github.com/vuejs/vue-test-utils-next/blob/master/src/emit.ts#L33 gets made _before_ this one https://github.com/vuejs/vue-next/blob/master/packages/runtime-core/src/renderer.ts#L442 which results in the test utils value being overwritten by `undefined`.

lmiller1990:
Also updating to 3.0.7.

resolves #436.

@nekosaur quick review and if possible can you grab this branch, build it and see if it fixes the problem in your app?


cexbrayat (approved):
👍 awesome, I gave it a try in one of my projects, and it fixes the issue. Thanks for taking the time to look into it!

nekosaur:
Thank for the quick fix!

PR #438, opened 2021-03-03
fix: stubbing RouterLink

### LINKED ISSUE #425: Cannot stub RouterLink using RouterLinkStub in tests when PascalCase is used for RouterLink in component
Repro repository: https://github.com/snoozbuster/vtu-router-link-casing-issue

When RouterLink is used in PascalCase in a component template, as in this simple example:

```vue
<template>
  <RouterLink class="link" :to="{ name: name }">
    <slot>
      A test link
    </slot>
  </RouterLink>
</template>

<script>
export default {
  name: "MyLink",
  props: {
    name: {
      type: String,
      required: true,
    },
  },
};
</script>
```

Then a test which uses `RouterLinkStub` will not pass:

```js
import { RouterLinkStub, shallowMount } from "@vue/test-utils";
import MyLink from "@/components/MyLink.vue";

describe("MyLink.vue", () => {
  it("should pass", () => {
    const name = "some-route";

    const wrapper = shallowMount(MyLink, {
      props: {
        name,
      },
      global: {
        stubs: {
          RouterLink: RouterLinkStub,
        },
      },
    });

    expect(wrapper.getComponent(".link").props("to")).toEqual({ name });
  });
});
```

Output:

```bash
> vue-cli-service test:unit

 FAIL  tests/unit/example.spec.js
  MyLink.vue
    ✕ should pass (25ms)

  ● MyLink.vue › should pass

    expect(received).toEqual(expected) // deep equality

    Expected: {"name": "some-route"}
    Received: undefined

      17 |     });
      18 |
    > 19 |     expect(wrapper.getComponent(".link").props("to")).toEqual({ name });
         |                                                       ^
      20 |   });
      21 | });
      22 |

      at Object.<anonymous> (tests/unit/example.spec.js:19:55)

  console.warn node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:40
    [Vue warn]: Failed to resolve component: RouterLink 
      at <MyLink name="some-route" ref="VTU_COMPONENT" > 
      at <VTUROOT>

Test Suites: 1 failed, 1 total
Tests:       1 failed, 1 total
Snapshots:   0 total
Time:        1.796s, estimated 2s
```

Changing the `<RouterLink>` in `MyLink.vue` to `<router-link>` solves the issue, but is a really sorry workaround.

lmiller1990:
resolves #425

Not sure if this ideal, pretty messy but this is one way to fix the problem. 🤔  hyphenating stubs by default messes things up.

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -414,8 +414,8 @@ export function mount(
   // components, too, such as <router-link> and <router-view>
   // so we register those globally.
   // ref: https://github.com/vuejs/vue-test-utils-next/issues/249
-  // we register both the hyphenated version and non-hyphenated
-  // version for good measure.
+  // we register the component as named in the stubs to avoid not being
+  // able to resolve the component later due to casing
   // ref: https://github.com/vuejs/vue-test-utils-next/issues/425
   if (global?.stubs) {
     for (const [name, stub] of Object.entries(global.stubs)) {
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -425,13 +425,19 @@ export function mount(
         // default stub.
         app.component(tag, stubbed)
 
-        if (!name.includes('-') && !['transition', 'transition-group'].includes(name)) {
+        if (
+          !name.includes('-') &&
+          !['transition', 'transition-group'].includes(name)
+        ) {
           app.component(name, stubbed)
         }
       } else {
         // user has provided a custom implementation.
         app.component(tag, stub)
-        if (!name.includes('-') && !['transition', 'transition-group'].includes(name)) {
+        if (
+          !name.includes('-') &&
+          !['transition', 'transition-group'].includes(name)
+        ) {
           app.component(name, stub)
         }
       }
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -413,16 +413,27 @@ export function mount(
   // users expect stubs to work with globally registered
   // components, too, such as <router-link> and <router-view>
   // so we register those globally.
-  // https://github.com/vuejs/vue-test-utils-next/issues/249
+  // ref: https://github.com/vuejs/vue-test-utils-next/issues/249
+  // we register both the hyphenated version and non-hyphenated
+  // version for good measure.
+  // ref: https://github.com/vuejs/vue-test-utils-next/issues/425
   if (global?.stubs) {
     for (const [name, stub] of Object.entries(global.stubs)) {
       const tag = hyphenate(name)
       if (stub === true) {
+        const stubbed = createStub({ name, props: {} })
         // default stub.
-        app.component(tag, createStub({ name, props: {} }))
+        app.component(tag, stubbed)
+
+        if (!name.includes('-') && !['transition', 'transition-group'].includes(name)) {
+          app.component(name, stubbed)
+        }
       } else {
         // user has provided a custom implementation.
         app.component(tag, stub)
+        if (!name.includes('-') && !['transition', 'transition-group'].includes(name)) {
+          app.component(name, stub)
+        }
       }
     }
   }
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -419,27 +419,13 @@ export function mount(
   // ref: https://github.com/vuejs/vue-test-utils-next/issues/425
   if (global?.stubs) {
     for (const [name, stub] of Object.entries(global.stubs)) {
-      const tag = hyphenate(name)
       if (stub === true) {
         const stubbed = createStub({ name, props: {} })
         // default stub.
-        app.component(tag, stubbed)
-
-        if (
-          !name.includes('-') &&
-          !['transition', 'transition-group'].includes(name)
-        ) {
-          app.component(name, stubbed)
-        }
+        app.component(name, stubbed)
       } else {
         // user has provided a custom implementation.
-        app.component(tag, stub)
-        if (
-          !name.includes('-') &&
-          !['transition', 'transition-group'].includes(name)
-        ) {
-          app.component(name, stub)
-        }
+        app.component(name, stub)
       }
     }
   }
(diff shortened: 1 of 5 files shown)


tests/getComponent.spec.ts
```diff
@@ -1,5 +1,6 @@
 import { defineComponent } from 'vue'
-import { mount } from '../src'
+import { mount, MountingOptions, RouterLinkStub, shallowMount } from '../src'
+import Issue425 from './components/Issue425.vue'
```
afontcu, line 3: I love this idea

src/mount.ts
```diff
@@ -413,16 +413,33 @@ export function mount(
...
         // default stub.
-        app.component(tag, createStub({ name, props: {} }))
+        app.component(tag, stubbed)
+
+        if (
+          !name.includes('-') &&
+          !['transition', 'transition-group'].includes(name)
+        ) {
+          app.component(name, stubbed)
+        }
```
afontcu, line 433: Could this logic be abstracte into a helper function that conveys meaning?

afontcu (approved):

src/mount.ts
```diff
@@ -413,16 +413,33 @@ export function mount(
...
       } else {
         // user has provided a custom implementation.
         app.component(tag, stub)
+        if (
+          !name.includes('-') &&
+          !['transition', 'transition-group'].includes(name)
+        ) {
+          app.component(name, stub)
+        }
```
cexbrayat, line 442: Can't we simply register the stub with the provided key in stubs? That's what we do for components (and I realize that's why I used this in my apps instead of stubs) and it should work fine?

Something like (untested):
```ts
 for (const [name, stub] of Object.entries(global.stubs)) {
  if (stub === true) {
    // default stub.
    app.component(name, createStub({ name, props: {} }))
  } else {
    // user has provided a custom implementation.
    app.component(name, stub)
  }
```

tests/components/Issue425.vue
```diff
@@ -0,0 +1,21 @@
+<template>
+<div>
+  <RouterLink class="link" :to="{ name: name }">
+    <slot>
+      A test link
+    </slot>
+  </RouterLink>
```
snoozbuster, line 7: dust: +1 indentation on these lines

src/mount.ts
```diff
@@ -413,16 +413,33 @@ export function mount(
...
   // components, too, such as <router-link> and <router-view>
   // so we register those globally.
-  // https://github.com/vuejs/vue-test-utils-next/issues/249
+  // ref: https://github.com/vuejs/vue-test-utils-next/issues/249
+  // we register both the hyphenated version and non-hyphenated
+  // version for good measure.
+  // ref: https://github.com/vuejs/vue-test-utils-next/issues/425
   if (global?.stubs) {
     for (const [name, stub] of Object.entries(global.stubs)) {
       const tag = hyphenate(name)
```
snoozbuster, line 422: IIRC, when a component is registered in PascalCase, it works in the template regardless of whether the template uses kebab-case or PascalCase. maybe we could solve this without double registration by replacing this with a `toPascalCase()`?

src/mount.ts
lmiller1990: Sure, I will do this

src/mount.ts
lmiller1990: That is what I thought, but it was not working - seems there is something else at work here 🤔

src/mount.ts
lmiller1990: Hmm neat idea, I will give this a try

src/mount.ts
lmiller1990: See comment below - I think we can just pascal case everything and it will "just work"

snoozbuster:
Question: how does `config.renderStubDefaultSlot` work? I was having trouble getting that to work in my project as well (where we are using 100% PascalCase) and now that I have the additional context of this PR, I am wondering if that also wasn't working because the stubs with the default slot rendering were registered as kebab-case...

lmiller1990:
`renderDefaultSlot` is a flag for back-compat with a somewhat unintended feature in v1. Basically in this lib when you stub a component, it's entirely is stubbed - including and slots it might otherwise render.

That flag can force the default slot to be rendered. There is an example [documented here](https://next.vue-test-utils.vuejs.org/api/#global) (note the `shallow: true` mounting option, which is the same as `shallowMount`.

It looks like if you use `components` instead of `stubs` your test works as-is - this bug only seems to apply to stubs (which will be fixed in this PR).

lmiller1990:
Fixed, I just removed `hyphenate` and it works like suggested. Pls re-review @snoozbuster @afontcu @cexbrayat.

afontcu (approved):

src/mount.ts
```diff
@@ -413,16 +413,19 @@ export function mount(
   // users expect stubs to work with globally registered
   // components, too, such as <router-link> and <router-view>
   // so we register those globally.
-  // https://github.com/vuejs/vue-test-utils-next/issues/249
+  // ref: https://github.com/vuejs/vue-test-utils-next/issues/249
+  // we register both the hyphenated version and non-hyphenated
+  // version for good measure.
```
snoozbuster, line 418: comment is no longer accurate. suggest something like: 

```suggestion
  // we register the component as named in the stubs to avoid not being
  // able to resolve the component later due to casing
```

might also be worth adding a note in the docs about the casing being expected to match

cexbrayat (approved):
That's what I had in mind 👍 
There is a comment that needs to be fixed.

And it would be great to squash all this into a single commit before merging 😉

lmiller1990:
@cexbrayat added the comment thanks to @snoozbuster.

I will use the squash/merge feature on GH to squash all the commits.

lmiller1990:
It's merged. I will probably do a release this weekend, we should get a few more things fixed by then. Thanks all!

snoozbuster:
@lmiller1990 could we release sooner than this weekend? some of the tests I am trying to write are getting stuck on this issue, and I don't want to put in a workaround that I know I will have to come back and fix very soon.

lmiller1990:
Alright, please wait for 24 hours or so.

Issue #362, opened 2021-02-07
Props of component stubs are not reactive


UrsMetz:
### Subject of the issue

When stubbing inner components in a test (e.g. using shallow mounting) the `props` of the stubs are not reactive (where as the `props` of a non-stub inner component are).
One example is a simple component with that passes data to an inner component and that data changes e.g. because a button is clicked, cf. example below.

### Steps to reproduce

The following unit test (using jest and Vue 3) shows the difference when using stubs and real components:
```ts
import {mount} from "@vue/test-utils";
import {defineComponent} from 'vue';

describe("Props updates of inner components", () => {
    const InnerComponent = defineComponent({
        name: "inner",
        props: {
            text: {
                type: String,
                required: true
            }
        },
        template: "<div>text</div>"
    });
    const OuterComponent = defineComponent({
        name: "outer",
        components: {InnerComponent},
        data() {
            return {
                text: 'initial'
            }
        },
        template: "<div><InnerComponent :text='text'></InnerComponent><button @click='changeText'></button></div>",
        methods: {
            changeText() {
                this.text = "new text";
            }
        }
    });

    it('non-shallow mount', async () => {
        const wrapper = mount(OuterComponent, {
            shallow: false
        });

        const innerComponent = wrapper.getComponent(InnerComponent);
        await wrapper.find('button').trigger('click');
        expect(innerComponent.props()).toHaveProperty('text', 'new text');
    });

    it('shallow mount', async () => {
        const wrapper = mount(OuterComponent, {
            shallow: true
        });

        const innerComponent = wrapper.getComponent(InnerComponent);
        await wrapper.find('button').trigger('click');
        expect(innerComponent.props()).toHaveProperty('text', 'new text');
    });
});
```
**edit**:
Moving the line
```ts
const innerComponent = wrapper.getComponent(InnerComponent);
```
after the button click makes both test run green.
**end of edit**

I can also provide a repo with the whole setup but it is basically a Vue 3 app created using the CLI (version 4.5.11).

### Expected behaviour

The props of the stubbed component should also be updated so that the expectation is true and both tests succeed.

### Actual behaviour

One the non-shallow test case succeeds the shallow one runs red.

### Possible Solution

I'd say this is a regression because with Vue 2 and "old" vue-test-utils both test would run green. So the reactive behavior of props of stubs should be resemble that of real inner components.

lmiller1990:
[...]

PR #453, opened 2021-03-10
Fix non-reactive stub props

UrsMetz:
As discussed in #362 this is a PR that includes a fix for the non-reactive props of stubs.

This is just the tests and the fix in separate commits. I have not added any comments. I still would like to discuss whether there is another less hacky way to fix this.


cexbrayat (approved):
LGTM 👍

Would you mind squashing the commits and write a commit message following the usual convention we have in the pro (starting with `fix:`, in lowercase, something like `fix: reactive props for stubs`)?

lmiller1990 (approved):

PR #454, opened 2021-03-10
feat: make isVisible more comprehensive

nandi95:
A quick addition so the method takes all styling into account.

### DIFF
--- a/src/utils/isElementVisible.ts
+++ b/src/utils/isElementVisible.ts
@@ -9,7 +9,7 @@ function isStyleVisible<T extends Element>(element: T) {
     return false
   }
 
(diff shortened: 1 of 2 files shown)


afontcu (approved):

lmiller1990 (approved):
Wow, neat!

IGx89:
Note for the future: getComputedStyle has been (re)defined to only work on connected elements ([ref](https://lists.w3.org/Archives/Public/www-style/2018May/0031.html)), so at some point in the future, when jsdom changes to match the spec, this will break. When that happens, see https://github.com/vuejs/vue-test-utils/pull/2013 for how I handled that scenario.

PR #459, opened 2021-03-13
fix: revert automatic two way sync of v-model prop values

### LINKED ISSUE #440: Emitting v-model event leads to unexpectedly updated prop value
I have a scenario where I want to test that interacting with a component (that results in emitting a v-model event) does _not_ lead to a state change in the internals of a component. 

Here is a simplified version of what I'm trying to do

```ts

const Foo = defineComponent({
  name: 'Foo',
  props: {
    foo: String,
  },
  emits: ['update:foo'],
  setup (props, ctx) {
    return () => h('div', {
      onClick: () => {
        ctx.emit('update:foo', 'baz')
      },
    }, props.foo)
  }
})

test('something', async () => {
  const wrapper = mount(Foo, {
    props: {
      foo: 'bar',
    }
  })

  expect(wrapper.text()).toEqual('bar')

  expect(wrapper.text()).toEqual('bar')
})
```

Currently this fails on the second expect with a received value of `baz`. This to me feels very strange and unintuitive. The component has emitted an event, but I have done nothing to catch it and subsequently update state to trigger a re-render, as I need to do in a real Vue app. Yet the props have been modified.

The change was introduced in this PR https://github.com/vuejs/vue-test-utils-next/pull/393. With this change, components under test do not behave in the same way as in normal operation.

edit:

Something a bit closer to real usage would be if you could perhaps do something like this, if you need the 2-way binding to be automatic.

```ts
const wrapper = mount(Foo, {
  props: {
    foo: 'bar',
    onUpdateFoo: async (foo) => {
      await wrapper.setProps({ foo })
    }
  }
})
```

Replacing the code from the PR with the following works, but I've no idea how stable it is.

```ts
var eventFnName = `on${event.split(':').map(vue.capitalize).join('')}`
if (wrapperVm.attrs[eventFnName]) {
  wrapperVm.attrs[eventFnName](...args)
}
```

nekosaur:
reverts changes made in https://github.com/vuejs/vue-test-utils-next/pull/393. two way syncing can be done using the following code

```ts
const wrapper = mount(component, {
  props: {
    modelValue: 1,
    'onUpdate:modelValue': async (modelValue: number) => wrapper.setProps({ modelValue })
  }
})
```

discussion regarding potential additional mount option for v-model usage continues in https://github.com/vuejs/vue-test-utils-next/discussions/279

closes #440

### DIFF
--- a/src/emit.ts
+++ b/src/emit.ts
@@ -64,16 +64,4 @@ export const recordEvent = (
 
   // Record the event message sent by the emit
   events[cid][event].push(args)
-
-  if (event.startsWith('update:')) {
-    if (args.length !== 1) {
-      throw new Error(
-        'Two-way bound properties have to emit a single value. ' +
-          args.length +
-          ' values given.'
-      )
-    }
-
-    vm.props[event.slice('update:'.length)] = args[0]
-  }
 }
--- a/tests/props.spec.ts
+++ b/tests/props.spec.ts
@@ -1,7 +1,7 @@
 import { mount } from '../src'
 import WithProps from './components/WithProps.vue'
 import Hello from './components/Hello.vue'
-import { defineComponent } from 'vue'
+import { defineComponent, h } from 'vue'
 
 describe('props', () => {
   it('returns a single prop applied to a component', () => {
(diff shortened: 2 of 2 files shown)


lmiller1990 (approved):

lmiller1990:
Thanks! I'll think about v-model a little more but won't delay the next release of this too much.

PR #483, opened 2021-03-27
fix(mount): avoid warning when stubbing

cexbrayat:
Now that we "properly" register stubs
(see this commit https://github.com/vuejs/vue-test-utils-next/commit/bb8023fc2dd1757f0358d3a8f2967a1064019540),
a bunch of warning are logged by Vue when a test tries to stub an already registered component.

For example, vue-router-mock register RouterView and then stubs it
(see https://github.com/posva/vue-router-mock/blob/v2/src/injections.ts#L37-L41).
So if you use vue-router-mock (or a similar pattern) in your test, Vue logs:

    [Vue warn]: Component "RouterView" has already been registered in target app.

because when a component is stubbed, we re-register it in VTU-next, and Vue warns us that it might be a mistake.

This is a quite naive fix, as it removes the already registered component before registering the stub.

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -420,6 +420,10 @@ export function mount(
   // ref: https://github.com/vuejs/vue-test-utils-next/issues/425
   if (global?.stubs) {
     for (const [name, stub] of Object.entries(global.stubs)) {
+      // removes a previously component with the same name if it exists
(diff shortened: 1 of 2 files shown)


cexbrayat:
Superseeded by #504

PR #494, opened 2021-03-29
fix: Using find after using findallcomponents does not match expectat…

llllllllllx:
Now I have fixed the problem that the use of find after using findallcomponents does not match the expectation, and the use of find after findallcomponents should be under the current node instead of the root node

### DIFF
--- a/src/vueWrapper.ts
+++ b/src/vueWrapper.ts
@@ -104,8 +104,11 @@ export class VueWrapper<T extends ComponentPublicInstance>
   ): DOMWrapper<SVGElementTagNameMap[K]>
   find<T extends Element>(selector: string): DOMWrapper<T>
   find(selector: string): DOMWrapper<Element> {
-    // force using the parentElement to allow finding the root element
-    const result = this.parentElement.querySelector(selector)
+    const result = this.parentElement['__vue_app__']
+      ? // force using the parentElement to allow finding the root element
+        this.parentElement.querySelector(selector)
+      : this.element.querySelector(selector)
+
     if (result) {
       return new DOMWrapper(result)
     }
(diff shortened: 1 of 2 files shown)


src/vueWrapper.ts
```diff
@@ -104,8 +104,11 @@ export class VueWrapper<T extends ComponentPublicInstance>
   ): DOMWrapper<SVGElementTagNameMap[K]>
   find<T extends Element>(selector: string): DOMWrapper<T>
   find(selector: string): DOMWrapper<Element> {
-    // force using the parentElement to allow finding the root element
-    const result = this.parentElement.querySelector(selector)
+    let result = this.parentElement['__vue_app__']
```
cexbrayat, line 107: nit: that should still be a const

src/vueWrapper.ts
```diff
@@ -202,7 +205,10 @@ export class VueWrapper<T extends ComponentPublicInstance>
   ): DOMWrapper<SVGElementTagNameMap[K]>[]
   findAll<T extends Element>(selector: string): DOMWrapper<T>[]
   findAll(selector: string): DOMWrapper<Element>[] {
-    const results = this.parentElement.querySelectorAll(selector)
+    let results = this.parentElement['__vue_app__']
```
cexbrayat, line 208: nit: same here

src/vueWrapper.ts
cexbrayat: Would you mind adding a comment on why we have to do this (you can even add a link to your issue)?

cexbrayat (changes requested):
LGTM, thanks for the PR 👍
I just added a few comments: if you can address them, I'll merge.

src/vueWrapper.ts
llllllllllx: this is indeed my mistake. I have modified and resubmitted the code

cexbrayat:
👍 I squashed the commits and mentioned the fixed issue. Thanks for your contribution!

PR #504, opened 2021-03-30
fix: allow global components to be stubbed

posva:
Coming from https://github.com/posva/vue-router-mock/issues/96

The warning happens because I do this: https://github.com/posva/vue-router-mock/blob/v2/src/injections.ts#L37-L41

I think this change is useful to toggle stubs with `stubs: {RouterLink: false}`

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -386,8 +386,12 @@ export function mount(
   }
 
   if (global.components) {
-    for (const key of Object.keys(global.components))
-      app.component(key, global.components[key])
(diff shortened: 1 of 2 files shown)


posva:
I'm realizing @cexbrayat already submitted a different fix at https://github.com/vuejs/vue-test-utils-next/pull/483/files. I think it's more reliable to not use the internal API of Vue if possible and avoid registering the global component when possible like in this PR

cexbrayat (approved):
I indeed prefer your fix 👍

PR #508, opened 2021-03-30
chore: bump to vue 3.0.9

cexbrayat:
There is potentially a regression in vue-next, but in the meantime let's fix the test and merge the updated version

PR #509, opened 2021-03-30
chore: remove tsd

cexbrayat:
TSD has weird regressions with vue-next 3.0.9, maybe due to TS 4.2.
Instead of figthing it, let's adopt what the other Vue projects did and replace TSD with a plain `tsc` check.

See https://github.com/vuejs/vue-next/commit/97dedebd8097116a16209664a1ca38392b964da3 for example for a similar commit on vue-next


cexbrayat:
This should unblock #508

PR #516, opened 2021-04-02
docs: Add docs and test for stubbing an async component

### LINKED ISSUE #514: Stub does not work for async component
When trying to stub an async component it still renders to original component instead of the stub component.

Version: 2.0.0-rc.4
```js
import { flushPromises, mount } from '@vue/test-utils'
import { defineAsyncComponent, defineComponent } from '@vue/runtime-core'

it('should stub async component', async () => {
  const AsyncComponent = defineComponent({
  })
  const StubComponent = defineComponent({
    template: '<span>StubComponent</span>'
  })
  const TestComponent = defineComponent({
    template: '<MyComponent/>',
    components: {
      MyComponent: defineAsyncComponent(async () => AsyncComponent)
    }
  })

  const wrapper = mount(TestComponent, {
    global: {
      stubs: {
        MyComponent: StubComponent
      }
    }
  })

  await flushPromises()

  expect(wrapper.html()).toContain('StubComponent')
})
```

Expected HTML to be `<span>StubComponent</span>` instead of `<span>AsyncComponent</span>`.

freakzlike:
Closes #514

Add section in docs for stubbing an async component and also a test case for this scenario.
It is required to set a name in your async component and use this name in `global.stubs`.

### DIFF
--- a/docs/guide/advanced/stubs-shallow-mount.md
+++ b/docs/guide/advanced/stubs-shallow-mount.md
@@ -135,6 +135,42 @@ test('shallow stubs out all child components', () => {
 If you used VTU V1, you may remember this as `shallowMount`. That method is still available, too - it's the same as writing `shallow: true`.
 :::
 
+## Stubbing an async component
+
+In case you want to stub out an async component, then make sure to provide a name for the component and use this name as stubs key.
+
+```js
+// AsyncComponent.js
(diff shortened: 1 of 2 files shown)


lmiller1990 (approved):
Nice, thank you!

PR #518, opened 2021-04-03
feat: Allow stub out async component without name and without need of async/await

freakzlike:
I have dealt some more with stubbing out an async component.
With this PR I want to allow to stub out the async component without the need to provide a name and use `async/await`.

The async component can now be stubbed out the same way as a sync component:
```js
// AsyncComponent.js
  name: 'AsyncComponent', // Not required any more
})

// App.js
const App = defineComponent({
  components: {
    MyComponent: defineAsyncComponent(() => import('./AsyncComponent'))
  },
  template: '<MyComponent/>'
})

// App.spec.js
test('stubs async component without resolving', () => {
  const wrapper = mount(App, {
    global: {
      stubs: {
        MyComponent: true
      }
    }
  })
  
  // No need to use await flushPromises()
  expect(wrapper.html()).toBe('<my-component-stub></my-component-stub>')
})
```

It can also be stubbed out like before using the name rather than the key.

```js
// AsyncComponent.js
  name: 'AsyncComponent',
})

// App.js
const App = defineComponent({
  components: {
    MyComponent: defineAsyncComponent(() => import('./AsyncComponent'))
  },
  template: '<MyComponent/>'
})

// App.spec.js
test('stubs async component with resolving', async () => {
  const wrapper = mount(App, {
    global: {
      stubs: {
        AsyncComponent: true
      }
    }
  })

  // flushPromises now required
  await flushPromises()

  expect(wrapper.html()).toBe('<async-component-stub></async-component-stub>')
})
```

I'm not pretty sure about the detection of the `AsyncComponentWrapper`. Maybe someone got a better way.
```js
const getAsyncComponentName = (
  type: VNodeTypes,
  props: ({ [key: string]: unknown } & VNodeProps) | null | undefined
): string | null => {
  // Async components will be mounted as AsyncComponentWrapper
  if (!isMountedComponent(type, props)) return null

  // AsyncComponentWrapper may only contain one component
  const components = (type as ComponentOptions).components as Record<
    string,
    ComponentOptions
  >
  if (!components || Object.keys(components).length !== 1) return null

  const asyncComponentName = Object.keys(components)[0]
  const asyncComponent = components[asyncComponentName]

  // Is AsyncComponentWrapper?
  if (
    !asyncComponent.__asyncLoader ||
    asyncComponent.name !== 'AsyncComponentWrapper'
  )
    return null

  return asyncComponentName
}
```

### DIFF
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -75,6 +75,23 @@ const resolveComponentStubByName = (
   }
 }
 
+const getComponentRegisteredName = (
+  instance: ComponentInternalInstance | null,
+  type: VNodeTypes
+): string | null => {
+  if (!instance || !instance.parent) return null
+
+  // try to infer the name based on local resolution
+  const registry = (instance.type as any).components
+  for (const key in registry) {
+    if (registry[key] === type) {
+      return key
+    }
+  }
+
+  return null
+}
+
 const isHTMLElement = (type: VNodeTypes) => typeof type === 'string'
 
 const isCommentOrFragment = (type: VNodeTypes) => typeof type === 'symbol'
@@ -139,27 +156,35 @@ export function stubComponents(
     }
 
     if (isComponent(type) || isFunctionalComponent(type)) {
-      let name = type['name'] || type['displayName']
+      const registeredName = getComponentRegisteredName(instance, type)
+      const componentName = type['name'] || type['displayName']
 
-      // if no name, then check the locally registered components in the parent
-      if (!name && instance && instance.parent) {
-        // try to infer the name based on local resolution
-        const registry = (instance.type as any).components
-        for (const key in registry) {
-          if (registry[key] === type) {
-            name = key
-            break
-          }
-        }
-      }
-      if (!name) {
+      // No name found?
+      if (!registeredName && !componentName) {
         return shallow ? ['stub'] : args
       }
 
-      const stub = resolveComponentStubByName(name, stubs)
+      let stub = null
+      let name = null
+
+      // Prio 1 using the key in locally registered components in the parent
+      if (registeredName) {
+        stub = resolveComponentStubByName(registeredName, stubs)
+        if (stub) {
+          name = registeredName
+        }
+      }
+
+      // Prio 2 using the name attribute in the component
+      if (!stub && componentName) {
+        stub = resolveComponentStubByName(componentName, stubs)
+        if (stub) {
+          name = componentName
+        }
+      }
 
       // case 2: custom implementation
-      if (typeof stub === 'object') {
+      if (stub && typeof stub === 'object') {
         // pass the props and children, for advanced stubbing
         return [stubs[name], props, children, patchFlag, dynamicProps]
       }
@@ -168,6 +193,11 @@ export function stubComponents(
       // where the signature is h(Component, props, slots)
       // case 1: default stub
       if (stub === true || shallow) {
+        // Set name when using shallow without stub
+        if (!name) {
+          name = registeredName || componentName
+        }
+
         const propsDeclaration = type?.props || {}
         const newStub = createStub({ name, propsDeclaration, props })
         stubs[name] = newStub
--- a/docs/guide/advanced/stubs-shallow-mount.md
+++ b/docs/guide/advanced/stubs-shallow-mount.md
@@ -83,6 +83,10 @@ test('stubs component', () => {
 
 This will stub out _all_ the `<FetchDataFromApi />` components in the entire render tree, regardless of what level they appear at. That's why it is in the `global` mounting option.
 
+::: tip
+To stub out you can either use the key in `components` or the name of your component. If both are given in `global.stubs` the key will be used first.
+:::
+
 ## Stubbing all children components
 
 Sometimes you might want to stub out _all_ the custom components. For example you might have a component like this:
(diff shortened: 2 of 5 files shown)


src/stubs.ts
```diff
@@ -87,6 +87,33 @@ const isMountedComponent = (
...
+): string | null => {
+  // Async components will be mounted as AsyncComponentWrapper
+  if (!isMountedComponent(type, props)) return null
+
+  // AsyncComponentWrapper may only contain one component
+  const components = (type as ComponentOptions).components as Record<
+    string,
+    ComponentOptions
+  >
+  if (!components || Object.keys(components).length !== 1) return null
```
lmiller1990, line 102: This seems like it will be a problem. For example

```ts
     it('stubs async component without name', () => {
      const TestComponent = defineComponent({
        components: {
          Foo: {
            template: '<div />'
          },
          MyComponent: AsyncComponentWithoutName,
        },
        template: '<MyComponent/>'
      })

      const wrapper = mount(TestComponent, {
        global: {
          stubs: {
            MyComponent: true
          }
        }
      })

      expect(wrapper.html()).toBe('<my-component-stub></my-component-stub>')
    })
```

Fails, because of

```ts
components: {
          Foo: {
            template: '<div />'
          },
          MyComponent: AsyncComponentWithoutName,
```

We can't rely on the `Object.keys` and getting the first element to capture the name of the component.

What we *can* do, I think, is use the parent component registry. See here: https://github.com/vuejs/vue-test-utils-next/blob/3e85d6ef3da57e5104f2b2ea1e4ca4b2f6ef75fc/src/stubs.ts#L145

I think we can match up using object identity. Something like above. Maybe...

```ts
      if (instance && instance.parent) {
        // try to infer the name based on local resolution
        const registry = (instance.type as any).components
        
        for (const [key, comp] of Object.entries(registry)) {
           if (args[0] === comp) {
             name = key
           }
         }
      }
```

That's untested, but there should be a way to find the component and get the name from the key.

lmiller1990 (commented):
Cool, looking good. I think we need to be more thorough in getting the name for un-named components, we should be able to match it using some similar techniques. Assuming the component is the first one in `components` and grabbing the name from the key doesn't seem very reliable.

If you are able to try to make this change, please do! If you are stuck I can give it a try.

freakzlike:
I have make it much cleaner now. The original issue was that it has already found the name of the `AsyncComponentWrapper` and uses this, instead of the name in the registered components.

I was wondering that it is not possible to stub out a component by using the name in the registered components rather than the name defined in the component.
```js
it('stubs component with different names', () => {
  const MyComponent = defineComponent({
    name: 'MyComponent',
    template: '<span>MyComponent</span>'
  })

  const TestComponent = defineComponent({
    components: {
      MyComponentAlias: MyComponent
    },
    template: '<MyComponentAlias/>'
  })

  const wrapper = mount(TestComponent, {
    global: {
      stubs: {
        MyComponentAlias: true, // does not work
        MyComponent: true // works
      }
    }
  })

  expect(wrapper.html()).toBe(
    '<my-component-alias-stub></my-component-alias-stub>'
  )
})
```

What do you think?
My idea was to give a list of names (name of component + name in registered components) to `resolveComponentStubByName` and return the first matching stub.

lmiller1990:
Ah, I see. I suppose we should use the key the component is registered under, since that is how you would use the component in your template. The only real usage of `name` is for better stack traces.

It looks like you accomplished it? https://github.com/vuejs/vue-test-utils-next/pull/518/files#diff-81bd0aa44ca1f02347a71cafda7db0708dcb4c2fe47e2c402db1925e2ffe49afR511

Can you clarify: are you asking for help on something, or are you asking what I think of your proposal?

I think this makes sense. The only edge case I see is someone registered a component using an alias that is used in another component as the `name`, but this seems fairly unlikely.

src/stubs.ts
```diff
@@ -87,6 +87,11 @@ const isMountedComponent = (
   props: ({ [key: string]: unknown } & VNodeProps) | null | undefined
 ) => isComponent(type) && props && props['ref'] === MOUNT_COMPONENT_REF
 
+const isAsyncComponentWrapper = (type: VNodeTypes) =>
+  isComponent(type) &&
+  type.__asyncLoader &&
+  type.name === 'AsyncComponentWrapper'
```
cexbrayat, line 93: Maybe add a comment here indicating that `AsyncComponentWrapper` is the name given by `vue-next` to `defineAsyncComponent`.

cexbrayat (approved):
LGTM, thanks for your investigation, and the thorough tests you added 👍

When you're done addressing the feedback, and have rebased and squashed all the commits, we'll be good to merge.

freakzlike:
I changed the stubs resolving to use the key in `components` (prio 1) **and** the name in the component (prio 2) to search for an entry in `global.stubs`. So there is also no need to use any special code for async components or for `AsyncComponentWrapper`

cexbrayat (changes requested):
I like the idea.

Regarding the code, I'm not of fan of using an array for the names. I think it would be easier to follow by using different variables, like `registeredName` and `componentName`.
Then instead of calling `resolveComponentStubByName` with an array, we could call it once with `registeredName` and if nothing is found, another time with `componentName`..
IIUC what you're doing, that should not change the result, but would be slightly easier to grasp while reading the code.

freakzlike:
@cexbrayat I changed the code according to your suggestion

cexbrayat (approved):
Awesome job, thanks.

PR #519, opened 2021-04-03
 fix: avoid using prefixIdenntifier in esm

elevatebart:
When using VTU within vite, the version used is esm-bundler. 
This version imports the esm-bundler of the compiler-core.

This version does not support prefixIdentifier.

I tried to make a conditional build of VTU.


src/utils/compileSlots.ts
```diff
@@ -15,10 +15,13 @@ export function processSlot(source = '', Vue = vue) {
     `<SlotWrapper v-bind="$attrs">${template}</SlotWrapper>`,
     {
       mode: 'function',
-      prefixIdentifiers: true
+      prefixIdentifiers: __BROWSER__
```
lmiller1990, line 18: I think you may need to update tsconfig (or somewhere else?) to make this compile.

lmiller1990 (commented):
Cool, once this is approved and all passing, we can merge.

lmiller1990:
Thanks @elevatebart, is there some way I can test this out to make sure it's going to work how we want it to? Do you have some way of testing (assuming you did with Vite etc)?

elevatebart:
We could create a playground-like environment to run VTU in a vite/ESM environment.
I think I would like to call the environment variable with another name than `__BROWSER__`. Maybe __ESM__ ? I called it this way because it is called this way in vue-next but I find it confusing. 
We use `__BROWSER__` in node and `!__BROWSER__` in vite with in-browser testing.
 
The good news is there is no emergency right now as I found an "unclean" workaround.

lmiller1990:
I will release this within 24h.

elevatebart:
thank you !!!

PR #521, opened 2021-04-05
fix: don't record native events that are included in emits option (#444)

### LINKED ISSUE #444: emitted Collect content that should not be collected
```js  
it.only('should does not have click', async () => {
    const Foo = defineComponent({
      name: 'Foo',
      emits: ['click', 'foo'],
      setup(_, ctx) {
        return () =>
          h(
            'div',
            {
              onClick: () => {
                ctx.emit('foo', 'bar')
              }
            },
            'hello world'
          )
      }
    })

    const wrapper = mount(Foo)
    console.log(wrapper.emitted()) //  { foo: [ [ 'bar' ] ], click: [ [ [MouseEvent] ] ] }
    expect(wrapper.emitted('foo')).toHaveLength(1)
  })
```
The example of the above

expect return **{foo:[['bar']]}** when call emitted()

but actual return  **{ foo: [ [ 'bar' ] ], click: [ [ [MouseEvent] ] ] }**

This may be related to the version of the vue, because do not have this problem when 3.0.4 version

version：
vue: v3.0.7
vtu: 2.0.0-rc.3

aethr:
fix #444 

I experimented with attaching the native event listener to the component's parentElement, hoping that vue's native event emitting behaviour would provide what we really want to collect, but this broke a couple of other tests.

Instead I opted to simply filter out any native event names that appear in `emits` while attaching our enumerated native event listeners. The added test passes and nothing else breaks so I think this should be adequate to fix #444.

### DIFF
--- a/src/vueWrapper.ts
+++ b/src/vueWrapper.ts
@@ -56,8 +56,14 @@ export class VueWrapper<T extends ComponentPublicInstance>
     const vm = this.vm
     if (!vm) return
 
+    const emits = vm.$options.emits || []
     const element = this.element
     for (let eventName of Object.keys(eventTypes)) {
+      // if a component includes events in 'emits' with the same name as native
+      // events, the native events with that name should be ignored
+      // @see https://github.com/vuejs/rfcs/blob/master/active-rfcs/0030-emits-option.md#fallthrough-control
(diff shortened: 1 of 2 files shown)


lmiller1990 (approved):
nicely done, thanks!

PR #522, opened 2021-04-06
refactor: avoid using import * in slots

elevatebart:
necessary for compat with vitejs

ESBuild hates `import *`. Since it looks at most files independently from each other (for speed). It can easily fail to understand what `import * as Vue from 'vue'` means.

I will create an issue on vitejs but in the meantime, this fixes the issues of @cypress/vite-dev-server with vue-demi.

### DIFF
--- a/src/utils/compileSlots.ts
+++ b/src/utils/compileSlots.ts
@@ -11,7 +11,9 @@ import type { SetupContext } from 'vue'
 
 export function processSlot(
   source = '',
-  Vue = {
+  // this parameter is only used for testing
+  // typing it does not matter much
+  Vue: unknown = {
     toDisplayString,
     createVNode,
     resolveComponent,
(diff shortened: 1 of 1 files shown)


lmiller1990 (commented):
Would jus doing `import Vue from 'vue'` not work? I feel like building the `Vue` object like this is error prone and going to lead to bugs, it really wants you to pass the entire Vue API.

lmiller1990:
Context is [this issue](https://github.com/vitejs/vite/issues/28890)

Although this fix just doesn't make sense - under this logic, we now need to patch every Vue library in existence to not using the valid `import * as <> from 'module` syntax. This seems like a Vue Demi (or Vite, but probably Vue Demi) problem. 

This code relies on having access to all of Vue, not just some parts of the render API, and will most likely introduce bugs/regressions.

Issue #321, opened 2021-01-29
listeners option on mount, better v-model support


mHooshdar:
Hello
I read the documenation but I couldn't find the listeners attribute in the mount option.
It was on v1 but we I couldn't find it for v2.
Do we have an alternative way to listen for the emitted events on the mounted component?

lmiller1990:
Hi! `$listeners` is [deprecated in Vue 3](https://v3.vuejs.org/guide/migration/listeners-removed.html#overview).

Generally you use [emitted](https://vue-test-utils.vuejs.org/v2/api/#emitted) to listen for emitted events. Does this work for your use case?

mHooshdar:
Thanks for your help
I wanted to test some side effects of the `update:modelValue` event.
I tried this and it fixed my problem:
`mount(Component, {
  props,
    attrs: {
      'onUpdate:modelValue' (value) {
        // something
       },
    },
});
`

afontcu:
Hi! Could you share your use case (component / test)?

lmiller1990:
Yes, please share your component/test, please! I normally do something like `expect(wrapper.emitted()['onUpdate:modelValue'][0]).toBe(...)` for example.

I think some better features to test v-model would be good. You could add your ideas to [this discussion](https://github.com/vuejs/vue-test-utils-next/discussions/279).

mHooshdar:
I have a component that opens a dropdown on the pill title click.
After clicking on one of the dropdown items, it will be selected and then the dropdown is closed.
I wanted to write a test for checking the call of the dropdown close function.
This is some part of my component code:
```javascript
export default defineComponent({
    props: {
        modelValue: {
            type: String,
            default: undefined,
        },
    },

    emits: ['update:modelValue'],

    setup (props, { emit }) {
        const dropdown = ref(null);
        const localValue = useVModel(props, 'modelValue', emit);

        watch(localValue, () => {
            // I wanted to test this function  when the localValue is changed 👇
            dropdown.value.close();
        });

        return {
            localValue,
            dropdown,
        };
    },
});
</script>
```

And this is some part of my test:
```javascript
test('should change pill title when click on a menu item', async () => {
    const wrapper = factory();
    wrapper.vm.$refs.dropdown.close = jest.fn();

    expect(pillElement.props('title')).toBe('title');
    expect(pillElement.props('active')).toBe(false);

    // open dropdown
    pillElement.trigger('click');
    await wrapper.vm.$nextTick();
[...]

PR #525, opened 2021-04-09
feat: vModel mounting option

lmiller1990:
Add a `vModel` mounting option.

I'd like lots of feedback. Please scrutinize - once we add this feature, we cannot go back or make any changes. We will need to maintain this for years to come. I am sure I am missing some tests and edge cases.

Also would like the opinion of @nekosaur and @kaelWD and @nandi95 - you are heavy users of test utils and have provided lots of useful feedback and contributions. Does this help with your `vModel` needs?

Component:
```vue
<template>
  <input @input="updateValue" />
  {{ color }}
</template>

<script>
export default {
  props: {
    color: String
  },
  methods: {
    handle($event) {
      this.$emit('update:color', $event.target.value)
    }
  }
}
</script>
```

Test:

```js
it('updates v-model automatically', async () => {
    vModel: {
      color: 'red'
    }
  })

  expect(wrapper.html()).toContain('red')

  // update the input value
  wrapper.find('input').element.value = 'blue'
  // trigger the input event
  await wrapper.find('input').trigger('input')

  // value is updated and reflected in the DOM
  expect(wrapper.html()).toContain('blue')
})
```

resolves #https://github.com/vuejs/vue-test-utils-next/issues/321 and https://github.com/vuejs/vue-test-utils-next/discussions/279.

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -16,6 +16,7 @@ interface MountingOptions<Props, Data = {}> {
   attrs?: Record<string, unknown>
   data?: () => {} extends Data ? any : Data extends object ? Partial<Data> : any
   props?: (RawProps & Props) | ({} extends Props ? null : never)
+  vModel?: (RawProps & Props) | ({} extends Props ? null : never)
   slots?: { [key: string]: Slot } & { default?: Slot }
   global?: GlobalMountOptions
   shallow?: boolean
@@ -198,7 +199,7 @@ test('data', () => {
 
 ### props
 
-Sets props on a component when mounted.
+Sets props on a component when mounted. Note that if you want to update a prop passed as `v-model`, [you should use `vModel`](/api/#vmodel).
 
 **Signature:**
 
@@ -241,6 +242,62 @@ test('props', () => {
   })
 
   expect(wrapper.html()).toContain('Count: 5')
+})
+```
+
+### vModel
+
+Similar to `props`, but values passed to `vModel` will be updated when using the [`update:modelValue` syntax](https://v3.vuejs.org/guide/component-custom-events.html#v-model-arguments).
+
+**Signature:**
+
+```ts
+vModel?: (RawProps & Props) | ({} extends Props ? null : never)
+```
+
+**Details:**
+
+`Component.vue`:
+
+```vue
+<template>
+  <input @input="updateValue" />
+  {{ color }}
+</template>
+
+<script>
+export default {
+  props: {
+    color: String
+  },
+  methods: {
+    handle($event) {
+      this.$emit('update:color', $event.target.value)
+    }
+  }
+}
+</script>
+```
+
+`Component.spec.js`:
+
+```js
+it('updates v-model automatically', async () => {
+  const wrapper = mount(Component, {
+    vModel: {
+      color: 'red'
+    }
+  })
+
+  expect(wrapper.html()).toContain('red')
+
+  // update the input value
+  wrapper.find('input').element.value = 'blue'
+  // trigger the input event
+  await wrapper.find('input').trigger('input')
+
+  // value is updated and reflected in the DOM
+  expect(wrapper.html()).toContain('blue')
 })
 ```
 
--- a/src/emit.ts
+++ b/src/emit.ts
@@ -27,20 +27,20 @@ export function emitted<T = unknown>(
   return vmEvents
 }
 
-export const attachEmitListener = () => {
+export const attachEmitListener = (vModelProps: string[]) => {
   events = {}
   // use devtools to capture this "emit"
-  setDevtoolsHook(createDevTools(events))
+  setDevtoolsHook(createDevTools(events, vModelProps))
 }
 
 // devtools hook only catches Vue component custom events
-function createDevTools(events: Events): any {
+function createDevTools(events: Events, vModelProps: string[]): any {
   return {
     emit(eventType, ...payload) {
       if (eventType !== DevtoolsHooks.COMPONENT_EMIT) return
 
       const [rootVM, componentVM, event, eventArgs] = payload
-      recordEvent(componentVM, event, eventArgs)
+      recordEvent(componentVM, event, eventArgs, vModelProps)
     }
   } as Partial<typeof devtools>
 }
@@ -48,7 +48,8 @@ function createDevTools(events: Events): any {
 export const recordEvent = (
   vm: ComponentInternalInstance,
   event: string,
-  args: unknown[]
+  args: unknown[],
+  vModelProps: string[]
 ): void => {
   // Functional component wrapper creates a parent component
   let wrapperVm = vm
@@ -64,4 +65,19 @@ export const recordEvent = (
 
   // Record the event message sent by the emit
   events[cid][event].push(args)
+
+  // see if we are emitting a v-model syntax sugar update:prop event
+  const match = event.match(/^(update:)(.*)/)
+
+  if (match && match[1] === 'update:' && vModelProps.includes(match[2])) {
+    if (args.length !== 1) {
+      throw new Error(
+        'Two-way bound properties have to emit a single value. ' +
+          args.length +
+          ' values given.'
+      )
+    }
+
+    vm.props[event.slice('update:'.length)] = args[0]
+  }
 }
--- a/src/types.ts
+++ b/src/types.ts
@@ -54,9 +54,14 @@ export interface MountingOptions<Props, Data = {}> {
    */
   props?: (RawProps & Props) | ({} extends Props ? null : never)
   /**
-   * @deprecated use `data` instead.
+   * @deprecated use `props` instead.
    */
   propsData?: Props
+  /**
+   * Sets props to be updated with v-model `update` syntax sugar.
+   * @see https://next.vue-test-utils.vuejs.org/api/#vModel
+   */
+  vModel?: Props
   /**
    * Sets component attributes when mounted.
    * @see https://next.vue-test-utils.vuejs.org/api/#attrs
--- a/tests/mountingOptions/vModel.spec.ts
+++ b/tests/mountingOptions/vModel.spec.ts
@@ -12,7 +12,7 @@ describe('vModel', () => {
         {{ foo }}
       `,
       methods: {
-      handle($event: KeyboardEvent) {
+        handle($event: KeyboardEvent) {
           this.$emit('update:foo', ($event.target as HTMLInputElement).value)
         }
       }
@@ -30,4 +30,3 @@ describe('vModel', () => {
     expect(wrapper.html()).toContain('bar')
   })
 })
-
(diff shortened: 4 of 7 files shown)


docs/api/index.md
```diff
@@ -16,6 +16,7 @@ interface MountingOptions<Props, Data = {}> {
   attrs?: Record<string, unknown>
   data?: () => {} extends Data ? any : Data extends object ? Partial<Data> : any
   props?: (RawProps & Props) | ({} extends Props ? null : never)
+  vModel?: (RawProps & Props) | ({} extends Props ? null : never)
```
pikax, line 19: In terms of typing, should we just extract the props that the user has defined? eg: `update:${Prop}`

This is because vue-3 already gives a warning when using an non declared event.

docs/api/index.md
```diff
@@ -244,6 +245,62 @@ test('props', () => {
...
+    }
+  }
+}
+</script>
+```
+
+`Component.spec.js`:
+
+```js
+it('updates v-model automatically', async () => {
```
afontcu, line 285: Not that important, but examples in docs use `test` instead of `it` :) #nitpicking

```suggestion
test('updates v-model automatically', async () => {
```

docs/api/index.md
```diff
@@ -244,6 +245,62 @@ test('props', () => {
...
+<template>
+  <input @input="updateValue" />
+  {{ color }}
+</template>
+
+<script>
+export default {
+  props: {
+    color: String
+  },
```
nandi95, line 272: Perhaps use modelValue as example because I imagine `v-model=""` will be the most frequent use-case.
```suggestion
  },
  emits: ['update:modelValue'],
```

docs/api/index.md
```diff
@@ -244,6 +245,62 @@ test('props', () => {
...
+      color: 'red'
+    }
+  })
+
+  expect(wrapper.html()).toContain('red')
+
+  // update the input value
+  wrapper.find('input').element.value = 'blue'
+  // trigger the input event
+  await wrapper.find('input').trigger('input')
```
afontcu, line 297: I might be missing something here, but I would've expected something along the lines of:

```suggestion
  await wrapper.find('input').setValue('blue')
```

src/types.ts
```diff
@@ -54,9 +54,14 @@ export interface MountingOptions<Props, Data = {}> {
    */
   props?: (RawProps & Props) | ({} extends Props ? null : never)
   /**
-   * @deprecated use `data` instead.
+   * @deprecated use `props` instead.
```
afontcu, line 57: nice catch 👍

tests/mountingOptions/vModel.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+import { defineComponent, h } from 'vue'
+
+describe('vModel', () => {
+  it('updates v-model automatically', async () => {
+    const Comp = defineComponent({
+      props: ['foo'],
+      template: `
+        <input @input="handle" />
+        {{ foo }}
+      `,
```
nandi95, line 13: missing `emits: ['update:foo']` here, I believe

src/emit.ts
```diff
@@ -64,4 +65,19 @@ export const recordEvent = (
 
   // Record the event message sent by the emit
   events[cid][event].push(args)
+
+  // see if we are emitting a v-model syntax sugar update:prop event
+  const match = event.match(/^(update:)(.*)/)
+
+  if (match && match[1] === 'update:' && vModelProps.includes(match[2])) {
+    if (args.length !== 1) {
+      throw new Error(
```
afontcu, line 74: Missing a test for this scenario? Not that common, but it would be great to know we don't run into some regression

tests/mountingOptions/vModel.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+        foo: 'foo'
+      }
+    })
+
+    expect(wrapper.html()).toContain('foo')
+    wrapper.find('input').element.value = 'bar'
+    await wrapper.find('input').trigger('input')
+    expect(wrapper.html()).toContain('bar')
+  })
+})
```
nandi95, line 32: Added missing test as pointed out by @afontcu 😂 
```suggestion
import { mount } from '../../src'
import { defineComponent, h } from 'vue'

describe('vModel', () => {
  const Comp = defineComponent({
    props: ['foo'],
    template: `
        <input @input="handle" />
        {{ foo }}
      `,
    emits: ['update:foo'],
    methods: {
      handle($event: KeyboardEvent) {
        this.$emit('update:foo', ($event.target as HTMLInputElement).value)
      }
    }
  })

  it('updates v-model automatically', async () => {
    const wrapper = mount(Comp, {
      vModel: {
        foo: 'foo'
      }
    })

    expect(wrapper.html()).toContain('foo')
    wrapper.find('input').element.value = 'bar'
    await wrapper.find('input').trigger('input')
    expect(wrapper.html()).toContain('bar')
  })

  it('should throw an error if payload is not 1 item', () => {
    const wrapper = mount(Comp, {
      vModel: {
        foo: 'foo'
      }
    })

    expect(() => wrapper.vm.$emit('update:foo')).toThrow(
      new Error('Two-way bound properties have to emit a single value. 0 values given.')
    )
    expect(() => wrapper.vm.$emit('update:foo', 1, 2)).toThrow(
      new Error('Two-way bound properties have to emit a single value. 2 values given.')
    )
  })
})
```

docs/api/index.md
```diff
@@ -244,6 +245,62 @@ test('props', () => {
...
+vModel?: (RawProps & Props) | ({} extends Props ? null : never)
+```
+
+**Details:**
+
+`Component.vue`:
+
+```vue
+<template>
+  <input @input="updateValue" />
```
cexbrayat, line 264: Should that call the `handle` method?

src/emit.ts
```diff
@@ -64,4 +65,19 @@ export const recordEvent = (
 
   // Record the event message sent by the emit
   events[cid][event].push(args)
+
+  // see if we are emitting a v-model syntax sugar update:prop event
```
cexbrayat, line 69: nit: maybe add a comment to explain that, if that's the case, we want to update props automatically.

cexbrayat (commented):
Code looks good, just a few nits.

I'm not sure I like the idea to add another option to the API just to save a bit of boilerplate, but if the team likes it, I don't mind.

nandi95:
I could get used to having a `vModel` option. I think it saves a lot of writing if I don't have to write a wrapping component in the `mount` method [like this](https://github.com/vuejs/vue-test-utils-next/discussions/279#discussioncomment-432686)
I can't think of edge cases but regardless there is still time to find any between the release of this and a stable major release I would have thought.

nekosaur:
@nandi95 that is one way of doing it, but a shorter one (with the exception of this PR) is simply

```js
  props: {
    modelValue: 5,
    'onUpdate:modelValue': modelValue => wrapper.setProps({ modelValue }),
  },
})
```

I'll try to find some time after work to have a look at this 👍

pikax:
Other API could provide an vmodel wrapper:
```ts
import { vmodel, mount } from 'vue-test-utils'

mount(Component, {
 props: {
   modelValue: vmodel(5)
  }
}}
 
```

lmiller1990:
I am also not 100% sold on having `vModel`. I don't really have a super strong feeling about it, though. This problem does come up often enough to justify having *something* (whether it's just better documentation, the `vModel` option, etc).

I'll go over the feedback closely and think about it more. No rush on merging this, happen to have alternative implementations explored!

nandi95:
> @nandi95 that is one way of doing it, but a shorter one (with the exception of this PR) is simply
> 
> ```js
> const wrapper = mount(Component, {
>   props: {
>     modelValue: 5,
>     'onUpdate:modelValue': modelValue => wrapper.setProps({ modelValue }),
>   },
> })
> ```
> 
> I'll try to find some time after work to have a look at this 👍

@nekosaur I have found an issue with this approach.
I've got a component that has a v-model which could be `null`, `Record<string, unknown>`, `Record<string, unknown>[]`
This component has a prop called `multi`, which is a boolean indicating that multiple things can be selected resulting in `Record<string, unknown>[]`
The component has a watcher that runs immediately and if `multi` is true, emit the `modelValue` to be `[]` if the `modelValue` is `null` or wrap it in an array if not an array already.

When setting a listener in the props, it will not update the props from the immediate watcher.

lmiller1990:
Good point @nandi95. Looking over VTU v1 there is many issues relating to watch and `setData` and `setProps`, for example: https://github.com/vuejs/vue-test-utils/issues/1756

I think this is one of the many complexities we should be aware of and why I'm very cautious about this new features.

@nandi95 does the wrapping method work in your use case?

nandi95:
@lmiller1990 The wrapping the component works as expected.
However I'm against this approach as I have might voiced before because it requires a lot more writing which adds up when writing a lot of components.
Here's how it was possible to do in rc-4 vs rc-5(with wrapping):
Screenshot 2021-04-21 at 09 51 48

Note: that `nextTick` was needed for some reason to wait for the `watch` to update the `modelValue` before clicking.

nekosaur:
@nandi95 @lmiller1990 The `vModel` option added in this PR does seem to support `watch` with `immediate: true`, since it will manually set the new prop value directly on `vm`.

It would be nice to be able to support `immediate: true` when doing it manually, but it looks like a catch 22

```ts
const wrapper = mount(Component as any, {
  props: {
    modelValue: null,
    'onUpdate:modelValue': (modelValue: any) => {
	  // With `immediate: true` this gets called before mounting is complete, which means wrapper is not yet assigned
      wrapper.setProps({ modelValue })
    }
  }
})
```

lmiller1990:
I think passing a `onUpdate:modelValue` prop with a `setProps` combo is a little weird. I personally don't want to write that every time I write a test. I feel like a wrapper is a bit better (`setProps` feels like a bit of a hack to me, as opposed to wrapping with another component).

I think we should add a bunch more tests for complex cases to at least explore `vModel` as a mounting option. Things like `immediate: true` with `watch`, `watchEffect`, etc, and see if there are any edge cases before we serious consider actually merging this. 

Does anyone feel strongly about the `vModel` mounting option? I personally don't - I have a similar problem in React, I just write a wrapper, the extra code doesn't bother me all that much.

lmiller1990:
This is still on the table, but not pre 2.0.0. I'm going to close it. We can revisit this once we hit a 2.0.0 release and Vue 3 is the main version on npm.

nandi95:
For the time being if someone's willing, docs can be added which outlines how it can currently be tested
 - Passing in the listener suggestion by @nekosaur which works to some extent
 - Wrapping the component suggestion by @KaelWD

This later can be updated along the lines of the v-model PRs

lmiller1990:
I will try to document something in the near future.

PR #531, opened 2021-04-10
feat: `vmodel` helper for two way binding

pikax:
Proposal for having a `vmodel` function to explicitly wrap `vmodel` value - @nekosaur, @KaelWD, @nandi95 and @cexbrayat 

```ts
const wrapper = mount(component, {
  props: {
    modelValue: vmodel(1, jest.fn()) // now modelValue will be reactive
  }
})
```
## vmodel

`vmodel` is a new function which will return a new `vmodel` object

```ts
export interface VModel<T> {
  value: Ref<UnwrapRef<T>>
  onChange?(v: T): void
}
```

Test:
```ts
  it('should return the updated props on 2 way binding', async () => {
    const component = defineComponent({
      template: '<button @click="increment"></button>',
      props: {
        modelValue: {
          type: Number,
          required: true
        }
      },
      emits: ['update:modelValue'],
      setup(props, ctx) {
        return {
          increment: () => ctx.emit('update:modelValue', props.modelValue + 1)
        }
      }
    })

    // @ts-expect-error TODO fix the type
    const wrapper = mount(component, {
      props: {
        modelValue: vmodel(1)
        // Not needed 
        // 'onUpdate:modelValue': async (modelValue: number) =>
        //   wrapper.setProps({ modelValue })
      }
    })

    expect(wrapper.props('modelValue')).toBe(1)
    expect(wrapper.props('modelValue')).toBe(2)
  })
``` 

### Other options

Implementation with a new `vmodel` option : https://github.com/vuejs/vue-test-utils-next/pull/525 

### Note

Types are not working with this, is this API is accepted I can fix that.

### DIFF
--- a/src/emit.ts
+++ b/src/emit.ts
@@ -4,6 +4,7 @@ import {
   ComponentPublicInstance,
   ComponentInternalInstance
 } from 'vue'
+import { VModel } from './vmodel'
 
 type Events<T = unknown> = Record<number, Record<string, T[]>>
 
@@ -27,20 +28,23 @@ export function emitted<T = unknown>(
   return vmEvents
 }
 
-export const attachEmitListener = () => {
+export const attachEmitListener = (vmodels: Map<string, VModel<any>>) => {
   events = {}
   // use devtools to capture this "emit"
-  setDevtoolsHook(createDevTools(events))
+  setDevtoolsHook(createDevTools(events, vmodels))
 }
 
 // devtools hook only catches Vue component custom events
-function createDevTools(events: Events): any {
+function createDevTools(
+  events: Events,
+  vmodels: Map<string, VModel<any>>
+): any {
   return {
     emit(eventType, ...payload) {
       if (eventType !== DevtoolsHooks.COMPONENT_EMIT) return
 
       const [rootVM, componentVM, event, eventArgs] = payload
-      recordEvent(componentVM, event, eventArgs)
+      recordEvent(componentVM, event, eventArgs, vmodels)
     }
   } as Partial<typeof devtools>
 }
@@ -48,7 +52,8 @@ function createDevTools(events: Events): any {
 export const recordEvent = (
   vm: ComponentInternalInstance,
   event: string,
-  args: unknown[]
+  args: unknown[],
+  vmodels?: Map<string, VModel<any>>
 ): void => {
   // Functional component wrapper creates a parent component
   let wrapperVm = vm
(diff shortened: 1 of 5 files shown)


src/emit.ts
```diff
@@ -62,6 +67,26 @@ export const recordEvent = (
...
+          'Two-way bound properties have to emit a single value. ' +
+            args.length +
+            ' values given.'
+        )
+      }
+      const v = args[0]
+      vmodel.value.value = v
+      vmodel.onChange?.(v)
+    } else {
+      // should we warn here?
```
nandi95, line 86: I would say no, because you could have props that have 2 way binding but not required.
In that case just I would just do:
```ts
if (!vmodel) {
    return;
}

if (args.length !== 1) {... etc
```
```

nandi95:
This to me seems more complicated then having an option `vModel` on the `mount` method.
Is there a advantage to this I might have missed?

pikax:
> Is there a advantage to this I might have missed?

Arguable (since @lmiller1990 does not like the extra option 😄 ) cleaner API, since you need to import `vmodel` method to have this functionality, also this should cover some more edge cases:
- Passing a `ref` should provide a two way binding to that `ref`, making the test possibly "cleaner"
- Opt-in approach without breaking or changing the current API.
- You can pass a `computed` and prevent property changes based on the `computed.set`
- simple `callback` to test if the update is being called without checking for the emitted. `vmodel(1, jest.fn())`

lmiller1990:
I think both are fine - either way we add another option (be it a method or mounting option).

I think `vModel` seems a (bit) cleaner, but if this covers more edge cases and solves the problem in a cleaner way, it might be better.

Maybe we go ahead with another RC while we think about this a bit more. 

The advantage of a `vModel` mounting option is it's more consistent with the existing Test Utils API.

pikax:
> The advantage of a `vModel` mounting option is it's more consistent with the existing Test Utils API. Do you have a strong opinion @pikax or just doing some exploration?

I don't have a strong opinion, either way is good.



> I think `vModel` seems a (bit) cleaner, but if this covers more edge cases and solves the problem in a cleaner way, it might be better.

This is pretty much a matter of preference, the only edge case this might cover and the `vModel` does not (at least atm) is the `vmodel(1, jest.fn())`, which could be added to the API

lmiller1990:
This is still on the table, but not pre 2.0.0. I'm going to close it. We can revisit this once we hit a 2.0.0 release and Vue 3 is the main version on npm.

PR #540, opened 2021-04-19
Fix set data reactive

### LINKED ISSUE #538: setData doesn't work
`wrapper.setData` doesn't set data in component. It sets data only in `wrapper.vm` object.

I created a simple components:

```
<template>
    nothing here
</template>

<script>
export default {
    data() {
        return {
            field: null
        };
    },
    computed: {
        isFieldNull() {
            return this.field == null;
        }
    }
}
</script>
```

And a test:

```
import { shallowMount } from '@vue/test-utils'
import testSetData from '../../src/cart/testSetData';
import { nextTick } from 'vue';


describe("test set data component", () => {
    it ("should set data", async () => {
        let wrapper = shallowMount(testSetData, {
            data () {
                return {
                    field: null
                }
            }
        });

        expect(wrapper.vm.isFieldNull).toBe(true);

        await wrapper.setData({ field: 10 });
        await nextTick();


        expect(wrapper.vm.field).toEqual(10);
        expect(wrapper.vm.isFieldNull).toBe(false);
    })
})
```

And the test fails on `expect(wrapper.vm.isFieldNull).toBe(false);` because `field` is still null.
Am I doing something wrong?

lmiller1990:
Make data via mounting option reactive.

Closes #538

### DIFF
--- a/tests/setData.spec.ts
+++ b/tests/setData.spec.ts
@@ -95,7 +95,7 @@ describe('setData', () => {
   })
 
   it('does not set new properties', async () => {
-    jest.spyOn(console, 'warn').mockImplementationOnce(() => { })
+    jest.spyOn(console, 'warn').mockImplementationOnce(() => {})
 
     const Component = {
       template: `<div>{{ foo || 'fallback' }}</div>`
@@ -125,7 +125,12 @@ describe('setData', () => {
   })
 
   it('https://github.com/vuejs/vue-test-utils-next/issues/538', async () => {
-    const Comp = defineComponent<{}, {}, { field: number | null }, { isFieldNull: any }>({
+    const Comp = defineComponent<
+      {},
+      {},
+      { field: number | null },
+      { isFieldNull: any }
+    >({
       template: `
         <div>{{ isFieldNull ? 'It is null' : 'It is not null' }}</div> 
       `,
--- a/tests/setData.spec.ts
+++ b/tests/setData.spec.ts
@@ -95,7 +95,7 @@ describe('setData', () => {
   })
 
   it('does not set new properties', async () => {
-    jest.spyOn(console, 'warn').mockImplementationOnce(() => {})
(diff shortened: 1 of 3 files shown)


tests/setData.spec.ts
```diff
@@ -123,4 +123,43 @@ describe('setData', () => {
       'Cannot add property count'
     )
   })
+
+  it('https://github.com/vuejs/vue-test-utils-next/issues/538', async () => {
+    const Comp = defineComponent<
+      {},
+      {},
+      { field: number | null },
+      { isFieldNull: any }
```
lmiller1990, line 132: Unfortunately the type for this is not exported. I thought we had some way to type `computed` properties and `vm` 🤔

cexbrayat: I don't use the Options API, so I don't know... Shouldn't it be `boolean` though?

tests/setData.spec.ts
```diff
@@ -123,4 +123,45 @@ describe('setData', () => {
       'Cannot add property count'
     )
   })
+
+  it('https://github.com/vuejs/vue-test-utils-next/issues/538', async () => {
```
cexbrayat, line 127: super nit: I prefer when we have a readable test name instead of just a link to a Github issue (easier to understand when the test is failing)

cexbrayat (approved):

tests/setData.spec.ts
lmiller1990: According to VS Code it should be `Computed<boolean>` or something like that, but it is not exported from Vue. I shouldn't actually need to specify this here, but if I don't TS complains. 

I have not used TS with Options API before so I have no idea 🤷‍♂️

tests/setData.spec.ts
lmiller1990: Sure, I'll update

PR #551, opened 2021-04-27
feat: added typehint to the trigger method

### LINKED ISSUE #415: Add event union type to trigger method
This will give us nice, typehints of the possible values.

The dom-event-type already includes the keys that can be triggered.
would `as const` and `keyof` work to extract the keys from that package?

Are some of it needs to be filtered out?

nandi95:
closes #415

### DIFF
--- a/src/baseWrapper.ts
+++ b/src/baseWrapper.ts
@@ -2,6 +2,7 @@ import { textContent } from './utils'
 import type { TriggerOptions } from './createDomEvent'
 import { nextTick } from 'vue'
 import { createDOMEvent } from './createDomEvent'
+import { DomEventName } from './constants/dom-event-types'
 
 export default abstract class BaseWrapper<ElementType extends Element> {
   private readonly wrapperElement: ElementType
@@ -64,12 +65,17 @@ export default abstract class BaseWrapper<ElementType extends Element> {
     return hasDisabledAttribute && elementCanBeDisabled
   }
 
+  async trigger(
+    eventString: DomEventName,
+    options?: TriggerOptions
+  ): Promise<void>
+  async trigger(eventString: string, options?: TriggerOptions): Promise<void>
   async trigger(eventString: string, options?: TriggerOptions) {
     if (options && options['target']) {
       throw Error(
         `[vue-test-utils]: you cannot set the target value of an event. See the notes section ` +
-        `of the docs for more details—` +
-        `https://vue-test-utils.vuejs.org/api/wrapper/trigger.html`
+          `of the docs for more details—` +
+          `https://vue-test-utils.vuejs.org/api/wrapper/trigger.html`
       )
     }
 
--- a/src/baseWrapper.ts
+++ b/src/baseWrapper.ts
@@ -1,6 +1,7 @@
 import { textContent } from './utils'
 import type { TriggerOptions } from './createDomEvent'
 import { nextTick } from 'vue'
+import { createDOMEvent } from './createDomEvent'
 
 export default abstract class BaseWrapper<ElementType extends Element> {
   private readonly wrapperElement: ElementType
@@ -63,5 +64,20 @@ export default abstract class BaseWrapper<ElementType extends Element> {
     return hasDisabledAttribute && elementCanBeDisabled
   }
 
-  abstract trigger(eventString: string, options?: TriggerOptions): Promise<void>
+  async trigger(eventString: string, options?: TriggerOptions) {
+    if (options && options['target']) {
+      throw Error(
+        `[vue-test-utils]: you cannot set the target value of an event. See the notes section ` +
+        `of the docs for more details—` +
+        `https://vue-test-utils.vuejs.org/api/wrapper/trigger.html`
+      )
+    }
+
+    if (this.element && !this.isDisabled()) {
+      const event = createDOMEvent(eventString, options)
+      this.element.dispatchEvent(event)
+    }
+
+    return nextTick()
+  }
 }
--- a/src/baseWrapper.ts
+++ b/src/baseWrapper.ts
@@ -4,7 +4,7 @@ import { nextTick } from 'vue'
 import { createDOMEvent } from './createDomEvent'
 import { DomEventName } from './constants/dom-event-types'
 
-export default abstract class BaseWrapper<ElementType extends Element> {
+export default class BaseWrapper<ElementType extends Element> {
   private readonly wrapperElement: ElementType
 
   get element() {
--- a/src/baseWrapper.ts
+++ b/src/baseWrapper.ts
@@ -2,6 +2,7 @@ import { textContent } from './utils'
 import { createDOMEvent } from './createDomEvent'
 import type { TriggerOptions } from './createDomEvent'
 import { nextTick } from 'vue'
+import { DomEventName } from './constants/dom-event-types'
 
 export default class BaseWrapper<ElementType extends Element> {
   private readonly wrapperElement: ElementType
@@ -44,7 +45,7 @@ export default class BaseWrapper<ElementType extends Element> {
     return true
   }
 
-  async trigger(eventString: string, options?: TriggerOptions) {
+  async trigger(eventString: DomEventName | string, options?: TriggerOptions) {
     if (options && options['target']) {
       throw Error(
         `[vue-test-utils]: you cannot set the target value of an event. See the notes section ` +
--- a/src/baseWrapper.ts
+++ b/src/baseWrapper.ts
@@ -1,10 +1,8 @@
 import { textContent } from './utils'
-import { createDOMEvent } from './createDomEvent'
 import type { TriggerOptions } from './createDomEvent'
 import { nextTick } from 'vue'
-import { DomEventName } from './constants/dom-event-types'
 
-export default class BaseWrapper<ElementType extends Element> {
+export default abstract class BaseWrapper<ElementType extends Element> {
   private readonly wrapperElement: ElementType
 
   get element() {
@@ -45,40 +43,28 @@ export default class BaseWrapper<ElementType extends Element> {
     return true
   }
 
-  async trigger(eventString: DomEventName | string, options?: TriggerOptions) {
-    if (options && options['target']) {
-      throw Error(
-        `[vue-test-utils]: you cannot set the target value of an event. See the notes section ` +
-          `of the docs for more details—` +
-          `https://vue-test-utils.vuejs.org/api/wrapper/trigger.html`
-      )
-    }
+  protected isDisabled = () => {
+    const validTagsToBeDisabled = [
+      'BUTTON',
+      'COMMAND',
+      'FIELDSET',
+      'KEYGEN',
+      'OPTGROUP',
+      'OPTION',
+      'SELECT',
+      'TEXTAREA',
+      'INPUT'
+    ]
+    const hasDisabledAttribute = this.attributes().disabled !== undefined
+    const elementCanBeDisabled = validTagsToBeDisabled.includes(
+      this.element.tagName
+    )
 
-    const isDisabled = () => {
-      const validTagsToBeDisabled = [
-        'BUTTON',
-        'COMMAND',
-        'FIELDSET',
-        'KEYGEN',
-        'OPTGROUP',
-        'OPTION',
-        'SELECT',
-        'TEXTAREA',
-        'INPUT'
-      ]
-      const hasDisabledAttribute = this.attributes().disabled !== undefined
-      const elementCanBeDisabled = validTagsToBeDisabled.includes(
-        this.element.tagName
-      )
-
-      return hasDisabledAttribute && elementCanBeDisabled
-    }
-
-    if (this.element && !isDisabled()) {
-      const event = createDOMEvent(eventString, options)
-      this.element.dispatchEvent(event)
-    }
-
-    return nextTick()
+    return hasDisabledAttribute && elementCanBeDisabled
   }
+
+  abstract trigger(
+    eventString: string,
+    options?: TriggerOptions
+  ): ReturnType<typeof nextTick>
 }
--- a/src/baseWrapper.ts
+++ b/src/baseWrapper.ts
@@ -63,8 +63,5 @@ export default abstract class BaseWrapper<ElementType extends Element> {
     return hasDisabledAttribute && elementCanBeDisabled
   }
 
-  abstract trigger(
-    eventString: string,
-    options?: TriggerOptions
-  ): ReturnType<typeof nextTick>
+  abstract trigger(eventString: string, options?: TriggerOptions): Promise<void>
 }
--- a/src/constants/dom-event-types.ts
+++ b/src/constants/dom-event-types.ts
@@ -1,0 +1,935 @@
+export type EventInterface =
+'AnimationEvent' |
+'AudioProcessingEvent' |
+'BeforeInputEvent' |
+'BeforeUnloadEvent' |
+'BlobEvent' |
+'CSSFontFaceLoadEvent' |
+'ClipboardEvent' |
+'CloseEvent' |
+'CompositionEvent' |
+'CustomEvent' |
+'DOMTransactionEvent' |
+'DeviceLightEvent' |
+'DeviceMotionEvent' |
+'DeviceOrientationEvent' |
+'DeviceProximityEvent' |
+'DragEvent' |
+'EditingBeforeInputEvent' |
+'ErrorEvent' |
+'Event' |
+'FetchEvent' |
+'FocusEvent' |
+'GamepadEvent' |
+'HashChangeEvent' |
+'IDBVersionChangeEvent' |
+'InputEvent' |
+'KeyboardEvent' |
+'MediaStreamEvent' |
+'MessageEvent' |
+'MouseEvent' |
+'MutationEvent' |
+'OfflineAudioCompletionEvent' |
+'OverconstrainedError' |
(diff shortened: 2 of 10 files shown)


src/createDomEvent.ts
```diff
@@ -146,7 +144,8 @@ function createEvent(eventParams: EventParams) {
   const { eventProperties, meta, eventType } = getEventProperties(eventParams)
 
   // user defined eventInterface
-  const metaEventInterface = window[meta.eventInterface]
+  // @ts-expect-error
+  const metaEventInterface: Event = window[meta.eventInterface]
```
nandi95, line 148: it says window is expecting a number index....

src/constants/dom-event-types.ts
```diff
@@ -0,0 +1,935 @@
...
+    cancelable: false
+  },
+  appinstalled: {
+    eventInterface: 'Event',
+    bubbles: false,
+    cancelable: false
+  },
+  /**
+   * @deprecated
+   */
```
nandi95, line 99: There's probably more I just added the missing properties and saw that it was depricated.

src/baseWrapper.ts
```diff
@@ -44,7 +45,7 @@ export default class BaseWrapper<ElementType extends Element> {
     return true
   }
 
-  async trigger(eventString: string, options?: TriggerOptions) {
+  async trigger(eventString: DomEventName | string, options?: TriggerOptions) {
```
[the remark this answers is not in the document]
cexbrayat: Or maybe an intermediate step would be to have 2 signatures: one with `DomEventName` and another deprecated one with `string`. People who are using a bad event name will have a deprecation warning and catch their mistake, without us breaking the compilation.

cexbrayat (commented):
I like the idea 👍

src/createDomEvent.ts
nandi95: The type from `getEventProperties` is correctly inferred, when I check it is the expected type, it's the `window` that can't be indexed. it has the type [`declare var window: Window & typeof globalThis;`](https://github.com/microsoft/TypeScript/blob/master/lib/lib.dom.d.ts#L19683)
We can maybe add to that window object...?

cexbrayat: Hmm strange. maybe if you cast `meta.eventInterface` as `keyof Window`?

src/baseWrapper.ts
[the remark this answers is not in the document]
nandi95: What about user defined events like it being checked [here](https://github.com/vuejs/vue-test-utils-next/blob/master/src/createDomEvent.ts#L151)?
I tried to deprecate only one signature but it doesn't seem to work. See this issue on ts: https://github.com/microsoft/TypeScript/issues/407#issuecomment-650266584

src/createDomEvent.ts
nandi95: Unfortunately that doesn't work either.
Screenshot 2021-04-27 at 12 33 56

What's interesting that it reckons `meta.eventInterface` is a `string` as opposed to `string | EventInterface` like in the `DomEvent` interface

src/baseWrapper.ts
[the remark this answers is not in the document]
cexbrayat: TBH I've never used user defined events. What is the use-case? Do we have tests for this? 

Yeah the deprecation needs to be in the proper order, I've fought with that in the past.

src/createDomEvent.ts
cexbrayat: What about `window[meta.eventInterface as keyof Window]`? I think that should work. If it does not I'll take a look myself.

src/baseWrapper.ts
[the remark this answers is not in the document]
nandi95: > TBH I've never used user defined events

Me neither, but I didn't think it was my place to remove the option.

> What is the use-case?  Do we have tests for this?

No clue, but there is a [test for it](https://github.com/vuejs/vue-test-utils-next/blob/master/tests/trigger.spec.ts#L323) that demonstrates.

> Yeah the deprecation needs to be in the proper order, I've fought with that in the past.

I think I tried every permuation where I can place the jsdoc comment :D
If you sure can be done, can you maybe take a look?

src/createDomEvent.ts
nandi95: That does! I think it's only marginally better than `// @ts-expect-error` though

src/baseWrapper.ts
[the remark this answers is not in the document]
cexbrayat: Hmm interesting. Let's see what other reviewers think about keeping it or removing it.
If necessary, I'll take a look on the deprecation thing 👍

src/createDomEvent.ts
cexbrayat: Cool! `ts-expect-error` will blindly swallow any error, so it should only be used as a last resort.

src/baseWrapper.ts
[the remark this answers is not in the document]
lmiller1990: I also have no idea about user defined events. This code is copy-pasted from VTU v1, it's been around for a long time.

lmiller1990:
Do we need tsd tests? This is a pretty nice PR. I'm looking forward to having type hints for events.

nandi95:
> Do we need tsd tests? This is a pretty nice PR. I'm looking forward to having type hints for events.

What is there to test? The return value doesn't change depending on the argument.

cexbrayat:
I think there are 3 options:
- we do not care about user defined events, remove this support, and change the signature to accept only DomEventName (and we can a tsd test to check that a simple string is not accepted)
- we want to keep the support but improve the typechecking, and we might want to split the signature to have two overloads (and maybe deprecate the one with string only?)
- we want to keep the support and not change much to the typechecking, and we merge as is (as `DomEventName | string` is accepted, TS can't really help).

cexbrayat:
I just realized this is the signature in the base wrapper!
So maybe user defined events are useful when you want to trigger a custom event of a child component (VueWrapper), but not on a DOM event (DomWrapper).
Maybe we should have different trigger signatures in both wrappers then:
- one which accepts only DomEventName in DomWrapper
- one which accepts strings (or even better, only the events defined in emits by the component) on the VueWrapper

nandi95:
That would be great actually (maybe stick an abstract signature into the base wrapper too).
However, It's a valid thing to do to emit your own event so dom wrapper should probably still be `DomeEventName | string` no?

As for the vueWrapper I'm not entirely sure how to get the `emits: []` type information into the trigger method.

cexbrayat:
I still think t would be better with overloads:
```
async trigger(eventString: DomEventName, options?: TriggerOptions): Promise<void>
async trigger(eventString: string, options?: TriggerOptions): Promise<void>
async trigger(eventString: DomEventName | string, options?: TriggerOptions): Promise<void> {
```
because then TS will pick the first signature when typing `trigger('cli` and offer `click` as the autocomplete.
But it will still allow a custom event as the signature falls back to string.
So that would be great for the DOMWrapper (and maybe for both)

For the VueWrapper, I think it's doable to get the emits and fall back to string as well, but I don't know exactly how of the top of my head. It's one of the generic parameters of `ComponentPublicInstance`. It's not a big deal if we don't do this, I just thought it would be ideal 😉

lmiller1990:
I personally don't mind what we go for, I have never had a problem with the existing typings, but always good to have better IDE hints. This sounds pretty good to me:

> one which accepts only DomEventName in DomWrapper
> one which accepts strings (or even better, only the events defined in emits by the component) on the VueWrapper

Are we planning to make any of these changes here, or shall we merge this and think about it more?

nandi95:
Please carefully consider the latest commit https://github.com/vuejs/vue-test-utils-next/pull/551/commits/9bba3822e3ad334a5ae08b9a625852407f081a62
To get the vue triggers to only accept the `emits: []` events I had to update couple of tests so it uses the dom wrapper
I imagine this could also be the case in the wild. If I were to add `| string` to the vue wrapper it seems my ide cannot type hint anymore.
Also The vue wrapper typehinting only seems to work when using the `defineComponent` otherwise the trigger method accepts anything

Currently in summary:
DOMWrapper takes `DomEventName | string`
VueWrapper takes the event defined in the emits only if `defineComponent` is used otherwise anything

src/domWrapper.ts
```diff
@@ -137,4 +140,21 @@ export class DOMWrapper<ElementType extends Element>
 
     return new DOMWrapper(parentElement).trigger('change')
   }
+
+  async trigger(eventString: DomEventName | string, options?: TriggerOptions) {
```
cexbrayat, line 144: I still think we need overrides to make the inference better:
```
async trigger(eventString: DomEventName, options?: TriggerOptions)
async trigger(eventString: string, options?: TriggerOptions)
```

src/vueWrapper.ts
```diff
@@ -257,6 +258,26 @@ export class VueWrapper<T extends ComponentPublicInstance>
 
     this.__app.unmount()
   }
+
+  async trigger(
+    eventString: Parameters<T['$emit']>[0],
```
cexbrayat, line 263: I don't think this works in all cases. This is going to accept only the parameters of the first `$emit` function, whereas there might be several signatures possible no?

`emits:` accepts an array with several events, or an object with several keys. It would be great to have tests showing that all these cases work (and I'm suspecting this is not the case).

tests/emit.spec.ts
```diff
@@ -316,7 +316,7 @@ describe('emitted', () => {
     })
 
     const wrapper = mount(Foo)
-    await wrapper.trigger('click')
+    await wrapper.get('div').trigger('click')
```
cexbrayat, line 319: This looks wrong, we should be able to keep the previous behavior.

src/baseWrapper.ts
```diff
@@ -45,40 +43,28 @@ export default class BaseWrapper<ElementType extends Element> {
...
-
-    return nextTick()
+    return hasDisabledAttribute && elementCanBeDisabled
   }
+
+  abstract trigger(
+    eventString: string,
+    options?: TriggerOptions
+  ): ReturnType<typeof nextTick>
 }
```
cexbrayat, line 70: Maybe I'm missing something, but the methods look very similar in both cases. Is it needed to have an abstract base method?

nit `typeof nextTick` can be written `Promise<void>` maybe?

cexbrayat (commented):
Thank you for giving this a try. I'm sadly not sure this works in every cases, and not convinced by having to update tests to get the DomWrapper and trigger a DomEvent.

Two possibilities IMO:
- you find a way to handle all emits cases, and remove the need to get the DomWrapper
- or we keep your previous version for now

src/vueWrapper.ts
nandi95: Yes, this tripped me up as well, this would ideally be `Parameters<T['$emit']>[number]` but then I don't get the typehint.

> It would be great to have tests showing that all these cases work

I'm not aware how you could test the arguments of the trigger method in type testing.

tests/emit.spec.ts
nandi95: Agreed

src/vueWrapper.ts
cexbrayat: I would add a few calls to trigger in `wrapper.d-test.ts`, like:
```ts
const CompWithEmitsAsObject = defineComponent({ emits: { event1: () => true, event2: () => true }});
const wrapper = mount(CompWithEmitsAsObject);
wraper.emit('event1');
wraper.emit('event2');
// @ts-expect-error 'other' is not defined as an emitted event
wrapper.emit('other');
```
Same thing for events defined in an array

lmiller1990:
I have not kept up with this thread - I was away for a few days.

I will leave the review/decisions to @nandi95 and @cexbrayat 🙏

src/vueWrapper.ts
nandi95: Ah, I got it now.
vueWrapper now also accepts dom event name and string so this is not applicable anymore

src/domWrapper.ts
nandi95: Hmm, didn't realise it would make it better but it did.

cexbrayat (commented):
This is starting to look good now 👍

One last thing: as both trigger methods are the same, can't we move it back to the BaseWrapper?
Once that's solved, I think we should merge and cut a release to give it a try.

src/vueWrapper.ts
cexbrayat: @pikax Maybe you can help here? We'd like to type `eventString` to accept the possible emitted events by the component. As I explained above, I'm afraid `Parameters<T['$emit']>[0]`, but you probably know best.

src/domWrapper.ts
```diff
@@ -137,4 +140,25 @@ export class DOMWrapper<ElementType extends Element>
 
     return new DOMWrapper(parentElement).trigger('change')
   }
+
+  async trigger(
+    eventString: DomEventName,
```
cexbrayat, line 145: While reviewing I realized that with TS 4.1+, we could even type and have autocompletion for things like `trigger('click.prevent')` with template literal types. But let's keep this for a followup PR 😅

src/vueWrapper.ts
pikax: @cexbrayat  I would extract the `Emits` like this:

```ts
const x = defineComponent({
  emits: {
    click: (n: number) => {},
    input: (b: string) => b.length > 1
  }
})
declare const t: typeof x extends DefineComponent<
  any,
  any,
  any,
  any,
  any,
  any,
  any,
  infer E
>
  ? E
  : undefined

t.click(1)
t.input('1')

//@ts-expect-error
t.click('2')
//@ts-expect-error
t.input(1)
```

nandi95:
> This is starting to look good now 👍
> 
> One last thing: as both trigger methods are the same, can't we move it back to the BaseWrapper?
> Once that's solved, I think we should merge and cut a release to give it a try.

I could move the one from dom wrapper, but not the vue wrapper one as the base wrapper doesn't have access to `T extends ComponentPublicInstance` generic type

cexbrayat:
@nandi95 for the signatures, I agree. but I'm talking about the body of the methods. You can keep the overloaded signatures in both wrappers and simply do `super.trigger(eventString, options);` with all the code in the base trigger method no?

nandi95:
Ah! didn't think of that 😅

Once this is merged, I'll create an issue for the modifiers

src/vueWrapper.ts
nandi95: I don't see the issue with `Parameters<T['$emit']>[0]` it's always going to be the first parameter of the emit function and it is correctly inferring both:
```ts
emits: {
    click: (n: number) => {},
    input: (b: string) => b.length > 1
  }
```
and
`emits: ['some', thing']`

when writing `wrapper.trigger('the event')`

Still cannot be tested as it accepts `string` too here.

nandi95:
> 👍 Did you try pikax suggestion for emits? https://github.com/vuejs/vue-test-utils-next/pull/551/files#r626442106

It was resolved, responded now

src/vueWrapper.ts
cexbrayat: Oh, I missed the edit of your comment.

Yeah I suspected it was not enough. I guess you can take a look into how the `emit` function of Vue itself is typed, but this is probably quite complex to handle properly.

Or maybe @pikax can give us a hand?

If this is too complex we can simply remove it and just use string. In that case, I guess all the overrides in the base class will be enough, and the code in both wrappers can be removed.

pikax: > I don't see the issue with `Parameters<T['$emit']>[0]` it's always going to be the first parameter of the emit function and it is correctly inferring both:

For events we need both event name and argument typed.

> I was wrong it was inferring the object syntax because it was click and input which is also a dom event name.

If the user declared `click` and `input` that component will not trigger the [DOM event](https://v3.vuejs.org/api/options-data.html#emits). - "Events listed in the emits option will not be inherited by the root element of the component and also will be excluded from the $attrs property." 






> For reference `E extends EmitsOptions` which is:
> `export declare type EmitsOptions = ObjectEmitsOptions | string[];` and `ObjectEmitsOptions` is a object with functions as you might have guessed.
> Have you got some guidance for correcting this type extraction?


The `emit` function type used in the setup context is [EmitFn<E>](https://github.com/vuejs/vue-next/blob/24e9efcc21dd517fddf8251873a06d71830e46eb/packages/runtime-core/src/componentEmits.ts#L24-L35) where's `E` is the `emit` object

lmiller1990:
Anything else needed here, or we can merge once CI is green?

Also no breaking change - right - just types (from what I can see, just checking).

lmiller1990:
Great, thanks for the update @cexbrayat!

src/domWrapper.ts
nandi95: agreed, I'll create a PR for this once, this is merged.

nandi95:
After discussion with @pikax I'm reverting the changes concerning the vue and dom wrapper and keep everthing in the base wrapper for now.

I think a discussion or issue can be raised if this comes up again whether should we typehing the event names from the components and its argument

My reasoning for this is:
I think it's hardly the case that in tests you manually emit a component event as its implementation of a behaviour not an organic user event (perhaps if you writing a component library with an odd design and the developer have to trigger an event?)

src/baseWrapper.ts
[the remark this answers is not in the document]
nandi95: Resolving as pr my https://github.com/vuejs/vue-test-utils-next/pull/551#issuecomment-838256534

PR #574, opened 2021-05-06
fix: use flushPromises from @vue/test-utils in the docs

alexandrecoin:
This PR removes the use of the `flushPromises` package in the code examples in favor of the `flushPromises` function from the `@vue/test-utils` package as it is already the case in the other examples found in the documentation.


cexbrayat (approved):
Good catch, thanks @alexandrecoin

PR #590, opened 2021-05-11
feat: Add event modifiers

nandi95:
linked #551


cexbrayat:
I think if we handle what's currently in the code base, that would be a nice start https://github.com/vuejs/vue-test-utils-next/blob/7a43f0756e72b3f10c98ebe0135347f5b53a5023/src/createDomEvent.ts#L7-L28

lmiller1990:
If we don't have any intention of doing this, I'd like to close it. Any update?

nandi95:
I'll continue this on the coming weekend

lmiller1990:
No rush, just determining if it's ongoing work or not. Sounds like it is, happy to leave it open 👍

src/constants/dom-events.ts
```diff
@@ -62,6 +62,41 @@ export interface DomEvent {
...
+  | typeof systemKeyModifiers[number]
+  | typeof mouseKeyModifiers[number]
+
+export type DomEventNameWithModifier =
+  | DomEventName
+  | `${DomEventName}.${typeof systemKeyModifiers[number]}`
+  | `click.${typeof mouseKeyModifiers[number]}`
+  | `click.${typeof systemKeyModifiers[number]}.${typeof mouseKeyModifiers[number]}`
+  | `${'keydown' | 'keyup'}.${keyof typeof keyCodesByKeyName}`
+  | `${'keydown' | 'keyup'}.${typeof systemKeyModifiers[number]}.${keyof typeof keyCodesByKeyName}`
```
nandi95, line 98: these 3 lines can have even more modifiers no?
Like: `click.ctrl.shift.middle`?

src/createDomEvent.ts
```diff
@@ -40,7 +18,7 @@ interface TriggerOptions {
 }
 
 interface EventParams {
-  eventType: DomEventName | string
+  eventType: DomEventName
```
nandi95, line 21: why was this also a string?

src/constants/dom-events.ts
lmiller1990: I have not actually combined them like that before, but I guess you can? Would need to verify. 

There is also a bunch of other ones apparently:

```
<!-- Vue 3 Key Modifier on v-on -->
<input @keyup.page-down="nextPage">

<!-- Matches both q and Q -->
<input @keypress.q="quit">
```

nandi95: I saw that there's `page-down` https://v3.vuejs.org/guide/events.html#key-modifiers (second example) therefore `page-up` must be a thing too. That has to be an alternative to `pagedown`, `pageup` which is already a thing in vtu
I just figured I won't bother as it's out of scope of this PR but if you can link to a list of modifiers I could add them.

> I have not actually combined them like that before, but I guess you can? Would need to verify.

I imagine it's a thing and probably you could chain as many you want.
This might be where it's actually processed: https://github.com/vuejs/vue-next/blob/8ed3ed6c27/packages/runtime-dom/src/directives/vOn.ts#L37

I would say so long we cover the most common ones for everything else `string` is still accepted

What do you think?

cexbrayat: @nandi95 Yeah I think this would be enough: we'll never be able to cover all cases, but the point is to just have a bit of help for the most common ones, and fall back to string for the rest.

lmiller1990: agreed, good enough is good enough sometimes,

lmiller1990:
Oh neat, it's green!

Anything specific you think is worth manually testing, or we good to merge?

nandi95:
> Oh neat, it's green!
> 
> Anything specific you think is worth manually testing, or we good to merge?

I don't think this cause any side effects so we good to merge.
I just tested it by checking the IDE makes the suggestions I expect.

Only thing could you resolve that comment about the `interface EventParams`?

src/createDomEvent.ts
lmiller1990: Not sure 🤔

lmiller1990:
Not sure on that `EventParams` type.

lmiller1990:
Will release sometime in the next few days, see if we can clean up a few other issues first. Thanks for this!!

Issue #245, opened 2020-11-12
Yarn 2 issues, @vue/test-utils tried to access @vue/compiler-dom


miii:
`@vue/compiler-dom` is not listed as a dependency which causes Yarn 2 to break.
```
@vue/test-utils tried to access @vue/compiler-dom, but it isn't declared in its dependencies;
This makes the require call ambiguous and unsound.
```
https://github.com/vuejs/vue-test-utils-next/blob/8a1f348969ffadfe1b35de746f0c27c020ee074c/src/utils/compileSlots.ts#L1

Temporary solution:
```yaml
# <project root>/.yarnrc.yml
packageExtensions:
  "@vue/test-utils@2.*":
    dependencies:
      "@vue/compiler-dom": "*"
```

lmiller1990:
Hi! I have not tried out yarn 2 yet.

Can we just add `@vue/compiler-dom` as a dependency to fix this? 

If you want to send a PR, I will do a release this weekend so you can use VTU with yarn 2.

sceee:
I am seeing the same issue with npm (I am in a vue2 project):
```
 ERROR  Failed to compile with 1 error

 error  in ./node_modules/@vue/test-utils/dist/vue-test-utils.esm-bundler.js

Module not found: Error: Can't resolve '@vue/compiler-dom' in 'C:\proj\node_modules\@vue\test-utils\dist'
```
So I guess @vue/compiler-dom should be added as dependency.

Installing it using `npm i -D @vue/compiler-dom` seems to solve the issue but I can unfortunately not be 100% sure as I now did run in the next issue (unrelated to this).

@lmiller1990 if you think it's correct to add this, I can submit a PR for this.

lmiller1990:
Vue 2 was not a monorepo and does not have a @vue/compiler-dom - have you got the right version of test utils? test-utils v1 is for Vue 2, test-utils v2 is for Vue 3 (confusing, yes - Vue 1 never had any testing utils, so that's why).

Either way if it is as simple as adding the dependency here, please go for it! yarn 2 compat would be great.

sceee:
Ok, I think I was getting confused with the versioning. My fault, sorry.
I am now on Vue2, @vue/test-utils 1 and webpack 4. So far so good, everything works (except I ideally would like to upgrade to webpack 5) but when doing so, I face quite some issues.

Anyway, as I'm on Vue 2, it definitely does not relate to here.
[...]

PR #591, opened 2021-05-11
Add  @vue/compiler-dom dep

shayneo:
Should resolve this issue: https://github.com/vuejs/vue-test-utils-next/issues/245


lmiller1990:
Cool, thanks. Have you tried this with Yarn 2? It would be good to know if this does actually solve the problem.

PR #598, opened 2021-05-13
docs: Port migration page from old repo

### LINKED ISSUE #593: Update migration guide
We have this: https://github.com/vuejs/vue-test-utils-next-docs/blob/master/src/guide/migration.md

It should be in the docs - I guess it got missed when moving the docs to this repo.

afontcu:
Closes #593

Looks like I missed it when migrating from repos, so here it is again :) It is a simple copy&paste from old file (plus some tweaks), we might be missing some additional content that we can add afterwards

### DIFF
--- a/docs/migration/index.md
+++ b/docs/migration/index.md
@@ -1,6 +1,6 @@
-# Migrating from Vue Test Utils 1
+# Migrating from Vue Test Utils v1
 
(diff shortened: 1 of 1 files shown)


lmiller1990:
Thanks!

Issue #482, opened 2021-03-27
Slotting behaviour


nandi95:
Hi,

I would like to do the two following things:
- Slot multiple components into the default or named slots. 
- Give the slotted components props (perhaps give the slot a `mount(Comp)`)

This isn't necessarily a code smell that the testing scope is too wide. My use case is I have a component that has a default slot into which you pass multiple components of one type.

```vue
<AvatarGroup>
    <Avatar>
<AvatarGroup />
```

The second would also be very useful when the user trying to test some unexpected patterns such as accessing children or parent.

Are these already possible I just missed something and perhaps the documentation can use some improvement?
Without delving into the source is this possible to implement? Is it desired?

nandi95:
In the linked test, it only tests with a [single html element](https://github.com/vuejs/vue-test-utils-next/blob/a3cb7f33deb8336a2294ae927593bc33591a9310/tests/mountingOptions/slots.spec.ts#L24)
And if this html string includes a component you would get the warning
> Failed to resolve component: Blah

This would work if to the mount the first argument is a wrapping component in which would have the `template` and the `components` both including the desired components. Something like
```vue
const wrapper = mount({
    template: `
        <ParentComp prop="something">
          <ChildComp prop="foo" />
          <ChildComp prop="bar" />
        </ParentComp>`,
    components: { ParentComp, ChildComp }
});
```
The `h(Foo)` is excellent, probably can be included in the docs somewhere? (more explicit docs or a cookbook section?)
 
The array to slots is probably a good idea.

lmiller1990:
We should include `h(Foo)` in the docs. Supporting an array should be pretty easy - if it's an array, just return a `h` function with an array. 

Another alternative might be supporting `h` as slots (no idea if this works). Something like

```ts
mount(Foo {
  slots: {
    default: h(Bar, { someProp: 'val' })
  }
})
```

That solves the props problem too. 🤔

cowills:
Running into this same issue being unable to pass an array into a slot, would be happy to try to implement that in VTU if that work hasn't been started and you can point me in the right direction. Not sure how to work around it currently without wrapping the array contents in a div.

lmiller1990:
Sure @cowills. Looks like the slot stuff [is here](https://github.com/vuejs/vue-test-utils-next/blob/ef69af17c9b29a952154679994475300450cd613/src/mount.ts#L262), I guess you can just check if it's an array an iterate over it. 

[...]

PR #573, opened 2021-05-06
work-in-progress array slot support

cowills:
Attempts to address https://github.com/vuejs/vue-test-utils-next/issues/482

Not currently working and not sure how to proceed

PR #599, opened 2021-05-13
feat: array of slots

lmiller1990:
Continue from: https://github.com/vuejs/vue-test-utils-next/pull/573

resolves https://github.com/vuejs/vue-test-utils-next/issues/482

Now you may pass an array of slots. There is some TS weirdness, I have no idea if the types are wrong or my code is wrong, but everything works as expected.

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -264,17 +264,17 @@ export function mount(
     to.appendChild(el)
   }
 
-  function slotToFunction(slot: Slot): Function {
-    if (typeof slot === 'object' && 'render' in slot && slot.render) {
-      return slot.render
+  function slotToFunction(slot: Slot) {
+    if (typeof slot === 'object') {
+      if ('render' in slot && slot.render) {
+        return slot.render
+      }
+
+      return () => slot
     }
 
     if (typeof slot === 'function') {
       return slot
-    }
-
-    if (typeof slot === 'object') {
-      return () => slot
     }
 
     if (typeof slot === 'string') {
@@ -302,12 +302,12 @@ export function mount(
         if (Array.isArray(slot)) {
           const normalized = slot.reduce<Array<Function | VNode>>(
             (acc, curr) => {
-              const toF = slotToFunction(curr)
+              const slotAsFn = slotToFunction(curr)
               if (isObject(curr) && 'render' in curr) {
-                const rendered = h(toF as any)
+                const rendered = h(slotAsFn as any)
                 return acc.concat(rendered)
               }
-              return acc.concat(toF())
+              return acc.concat(slotAsFn())
             },
             []
           )
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -264,7 +264,6 @@ export function mount(
     to.appendChild(el)
   }
 
-
   function slotToFunction(slot: Slot): Function {
     if (typeof slot === 'object' && 'render' in slot && slot.render) {
       return slot.render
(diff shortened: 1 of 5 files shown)


src/mount.ts
```diff
@@ -269,33 +299,24 @@ export function mount(
...
-        if (typeof slot === 'object' && 'render' in slot && slot.render) {
-          acc[name] = slot.render
-          return acc
-        }
+        if (Array.isArray(slot)) {
+          const normalized = slot.reduce<Array<Function | VNode>>(
+            (acc, curr) => {
+              const toF = slotToFunction(curr)
+              if (isObject(curr) && 'render' in curr) {
+                const rendered = h(toF as any)
```
lmiller1990, line 307: Not entirely clear why I need to pass `toF` (short for `toFunction`) into `h` here... but you need to.

src/mount.ts
```diff
@@ -261,6 +264,33 @@ export function mount(
...
+      return slot.render
+    }
+
+    if (typeof slot === 'function') {
+      return slot
+    }
+
+    if (typeof slot === 'object') {
+      return () => slot
+    }
```
cexbrayat, line 278: nit: I would write a single ` if (typeof slot === 'object')` and then deal with two cases inside.

src/mount.ts
```diff
@@ -261,6 +264,33 @@ export function mount(
     to.appendChild(el)
   }
 
+  function slotToFunction(slot: Slot): Function {
```
cexbrayat, line 267: nit: Function is rarely used as a type. We usually prefer `() => string` for example.

src/mount.ts
```diff
@@ -269,33 +299,24 @@ export function mount(
...
-        // case of an SFC getting passed
-        if (typeof slot === 'object' && 'render' in slot && slot.render) {
-          acc[name] = slot.render
-          return acc
-        }
+        if (Array.isArray(slot)) {
+          const normalized = slot.reduce<Array<Function | VNode>>(
+            (acc, curr) => {
+              const toF = slotToFunction(curr)
+              if (isObject(curr) && 'render' in curr) {
```
cexbrayat, line 306: I have a hard time understanding the code. Maybe some comments would help. `toF` could probably be renamed to something more meaningful (`slotAsFunction`?).

It feels like this check is somewhat redundant with what is done in `slotToFunction`. I don't know exactly how, but it looks like this should be refactor/simplified.

src/mount.ts
cexbrayat: This is indeed weird. I don't know enough about slot internal behavior to help sadly.

src/mount.ts
lmiller1990: Naming definitely can be improved.

I also have a hard time understanding this, or why it works - the amount of ways you can create a Vue component (and as such, a lot) is a bit overwhelming. I can think of at least 5 different ways:

- class component
- function component
- SFC
- setup returning render function
- string (`i am a string`)
- template (`<div>template</div>`)

Throw in scoped slots, and you end up with this.

src/mount.ts
lmiller1990: Ok, I will change this. I suppose it'll be like `type SlotLike = () => string | (() => string) | { render: () => string }` or something to that meaning.

lmiller1990: The type here is *very* complicated, it wants me to specify some very large union of public component instance, functional components... I just removed the entire return type and will let TS figure it out.

src/mount.ts
lmiller1990: DOne

Issue #606, opened 2021-05-19
Using renderStubDefaultSlot disable shallow option.


YutamaKotaro:
## What's happened?

I ran into strange behavior using shallow and renderStubDefaultSlot.
It seems to be a bug, I guess.

## Code

```js
  const Child = defineComponent({
    template: '<div><p>child</p><slot /></div>'
  })
  const Component = defineComponent({
    template: '<div><Child><div>hello</div></Child></div>',
    components: {
      Child
    }
  })
  const wrapper = mount(Component, {
    shallow: true,
    global: {
      renderStubDefaultSlot: true
    }
  })
  console.log(wrapper.html())
```

## actual

Shallow option is disabled unexpectedly.

```
    <div>
      <div>
        <p>child</p>
        <div>hello</div>
      </div>
    </div>
```

## expected

```
<div>
    <child-stub><div>hello</div></ child-stub>
</div>
```

## Workaround
As workaround... there is a way using config object, but I wonder this is not expected behavior for global option of mount.

PR: #607

PR #607, opened 2021-05-19
fix: shallow config issue

YutamaKotaro:
## What for ??
I'm sorry to bother you guys 🙇🏻 
This pull request is for #606 

## Why the issue happened.
This caused that issue.

```js
stubComponents(global.stubs, global.renderStubDefaultSlot ? false : options?.shallow )
```

If global.renderStubDefaultSlot is true, options.shallow is ignored.
Hence, the issue caused.

### DIFF
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -12,7 +12,6 @@ import {
 } from 'vue'
 import { hyphenate } from './utils/vueShared'
 import { MOUNT_COMPONENT_REF, MOUNT_PARENT_NAME } from './constants'
-import { config } from './config'
 import { matchName } from './utils/matchName'
 import { ComponentInternalInstance } from '@vue/runtime-core'
 
@@ -33,7 +32,7 @@ export const createStub = ({
   const tag = name ? `${hyphenate(name)}-stub` : anonName
 
   const render = (ctx: ComponentPublicInstance) => {
-    return h(tag, props, (renderStubDefaultSlot || config.renderStubDefaultSlot) ? ctx.$slots : undefined)
+    return h(tag, props, renderStubDefaultSlot ? ctx.$slots : undefined)
   }
 
   return defineComponent({
@@ -160,7 +159,7 @@ export function stubComponents(
 
       // No name found?
       if (!registeredName && !componentName) {
-        return shallow ? ['stub'] : args
+        return (renderStubDefaultSlot || !shallow) ? args : ['stub']
       }
 
       let stub = null
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -20,22 +20,20 @@ interface StubOptions {
   name: string
   props?: any
   propsDeclaration?: any
-}
-
-function getSlots(ctx: ComponentPublicInstance): Slots | undefined {
-  return !config.renderStubDefaultSlot ? undefined : ctx.$slots
+  renderStubDefaultSlot?: boolean
 }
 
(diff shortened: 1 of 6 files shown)


src/utils.ts
```diff
@@ -31,9 +36,7 @@ export function mergeGlobalProperties(
     mocks: { ...configGlobal.mocks, ...mountGlobal.mocks },
     config: { ...configGlobal.config, ...mountGlobal.config },
     directives: { ...configGlobal.directives, ...mountGlobal.directives },
-    renderStubDefaultSlot: !!(mountGlobal.renderStubDefaultSlot !== undefined
-      ? mountGlobal.renderStubDefaultSlot
-      : configGlobal.renderStubDefaultSlot)
```
YutamaKotaro, line 36: This library has three point where we can write renderStubDefaultSlot option.

- mount function
- config.global object
- config object

The last one,  I wasn't able to find  in document, but it exists in GlobalConfigOptions.
So, I fixed code to use one of them (priority: mount -> config -> config.global).

YutamaKotaro: Or if it's for internal, I'll fix my code 🙇🏻

src/stubs.ts
```diff
@@ -12,30 +12,27 @@ import {
...
 
 interface StubOptions {
   name: string
   props?: any
   propsDeclaration?: any
-}
-
-function getSlots(ctx: ComponentPublicInstance): Slots | undefined {
-  return !config.renderStubDefaultSlot ? undefined : ctx.$slots
+  renderStubDefaultSlot?: boolean
```
YutamaKotaro, line 22: In mount.ts, mergedConfig is used.

```
  const global = mergeGlobalProperties(config.global, options?.global)
```

So, I unified using mergetConfig.

tests/mountingOptions/stubs.global.spec.ts
```diff
@@ -607,7 +607,7 @@ describe('mounting options: stubs', () => {
       )
     })
 
-    it('renders the default slot of deeply nested stubs when renderStubDefaultSlot=true', () => {
+    it('renders the default slot of deeply nested stubs when renderStubDefaultSlot=true', async () => {
```
lmiller1990, line 610: Not needed.

src/utils.ts
lmiller1990: This makes sense, nice job figuring this out.

lmiller1990 (commented):
Just one comment - can you removed the `async` that you added (which is not needed).

Great job figuring out how this works.

src/utils.ts
YutamaKotaro: Thanks a lot !!
It's an honor to hear you say that. 🙇🏻

tests/mountingOptions/stubs.global.spec.ts
YutamaKotaro: oh...my gosh... Sorry for my fault. I'll remove this.

lmiller1990: You apologized twice - once in your PR description and once here. No need to do that! You are doing the world a favor with your PR, if anything everyone should be thanking *you*. 🎉 Thanks a lot!

lmiller1990 (approved):

lmiller1990:
LGTM, I will do a release in the next few days and this will be included.

Issue #530, opened 2021-04-09
shallowMount weird behaviour with repeating component stubs in snapshot


LorinIonita:
I haven't seen an issue like this and I'm not sure if this is an issue on my end but there's a weird behaviour when trying to shallow mount a component with multiple repeating components. `ai-btn` and `ai-icon` are added as global components in the mount.

This is the component:

```
<template lang="pug">
  .ai-simple-pagination
    ai-btn.ai-simple-pagination__arrow-button.left(
      variant="label",
      @click="goBack",
      :disabled="!hasPrev"
    )
      ai-icon.ai-simple-pagination__icon(
        icon="arrow-small-left",
        width="30",
        height="30"
      )
    ai-btn.ai-simple-pagination__arrow-button.right(
      variant="label",
      @click="goForward",
      :disabled="!hasNext"
    )
      ai-icon.ai-simple-pagination__icon(
        icon="arrow-small-right",
        width="30",
        height="30"
      )
</template>

<script lang="ts">
import {
  ref,
  PropType,
  defineComponent,
} from 'vue';

export default defineComponent({
  name: 'AiSimplePagination',
  props: {
    value: {
      type: Number,
      default: 1,
    },
    onChange: {
      type: Function as PropType<(val: number) => void>,
      required: true,
    },
    hasPrev: {
      type: Boolean,
      required: true,
    },
    hasNext: {
      type: Boolean,
      required: true,
    },
  },
  setup(props) {
    const model = ref<number>(props.value);

    const goForward = () => {
      model.value += 1;
      props.onChange(model.value);
    };

    const goBack = () => {
      model.value -= 1;
      props.onChange(model.value);
    };

    return {
      model,
      props,
      goForward,
      goBack,
    };
  },
});
</script>
```

When mounting it with `shallowMount` and expecting it to match the snapshot, the snapshot looks like this:

```
<div
  class="ai-simple-pagination"
>
  <ai-btn-stub
    class="ai-simple-pagination__arrow-button left"
    disabled="true"
    variant="label"
  >
    <ai-icon-stub
      class="ai-simple-pagination__icon"
      height="30"
      icon="arrow-small-left"
      width="30"
    />
  </ai-btn-stub>
  <ai-btn-stub
    class="ai-simple-pagination__arrow-button left ai-simple-pagination__arrow-button right"
    disabled="false"
    variant="label"
  >
    <ai-icon-stub
      class="ai-simple-pagination__icon"
      height="30"
[...]

PR #610, opened 2021-05-22
fix: Problem with stub props

YutamaKotaro:
# For what ??
This pull request is for #530 

# why the issue happened ?
CreateStub was using props of transformVNodeArgs on renderFunction.

```js
return h(tag, props, renderStubDefaultSlot ? ctx.$slots : undefined)<br class="Apple-interchange-newline">
```
This code means that the firstProps is always used for render function.
Hence, if we use shallow or stub this, problem happened.


tests/mountingOptions/stubs.global.spec.ts
```diff
@@ -119,7 +119,7 @@ describe('mounting options: stubs', () => {
 
     expect(wrapper.html()).toEqual(
       '<div>\n' +
-        '  <foo-stub class="bar" test-id="foo" dynamic="[object Object]"></foo-stub>\n' +
+        '  <foo-stub dynamic="[object Object]" class="bar" test-id="foo"></foo-stub>\n' +
```
YutamaKotaro, line 122: Instead of fixing this issue... I wasn't able order of props and attributes.
If we need to keep this order... I'll try it !!

lmiller1990: The order of attributes has no specific meaning, so it's fine if it changes a bit. This is one of the things I dislike about snapshots; they often change, even if the component works the same.

lmiller1990 (commented):
Code seems fine - `shallowMount` is so complicated, lol.

You mention a problem - is that the order of the attributes? I don't think this is really a problem - the order isn't important for the actual behavior. Some people might need to update their snapshots. That's fine - the order of the props is not guaranteed to be deterministic, as far as I know.

src/stubs.ts
```diff
@@ -24,15 +24,14 @@ interface StubOptions {
...
-  props,
   propsDeclaration,
   renderStubDefaultSlot
 }: StubOptions): ComponentOptions => {
   const anonName = 'anonymous-stub'
   const tag = name ? `${hyphenate(name)}-stub` : anonName
 
   const render = (ctx: ComponentPublicInstance) => {
-    return h(tag, props, renderStubDefaultSlot ? ctx.$slots : undefined)
+    return h(tag, ctx.$props, renderStubDefaultSlot ? ctx.$slots : undefined)
```
lmiller1990, line 34: Ah, nice.

src/stubs.ts
```diff
@@ -112,6 +111,7 @@ export function stubComponents(
   shallow: boolean = false,
   renderStubDefaultSlot: boolean = false
 ) {
+  const component: { [key: string]: ComponentOptions } = {}
```
lmiller1990, line 114: You could write `Record<string, ComponentOptions>`.

YutamaKotaro:
> Code seems fine - shallowMount is so complicated, lol.

Yep, I think so too. It looks me a little time to find problems lol.

> You mention a problem - is that the order of the attributes? I don't think this is really a problem - the order isn't important for the actual behavior. Some people might need to update their snapshots. That's fine - the order of the props is not guaranteed to be deterministic, as far as I know.

That's right. What I meant was the order of the attributes.
Thanks a lot !! I'm relived to hear of that.

tests/mountingOptions/stubs.global.spec.ts
YutamaKotaro: I got it !!
Thanks a lot !!

src/stubs.ts
YutamaKotaro: I got it !! I'll fix it soon !!

YutamaKotaro:
I corrected the points you pointed out !!

lmiller1990 (approved):

PR #636, opened 2021-05-31
Use null instead of false for disabled attribute.

connorshea:
In Vue 3, passing `false` won't do the same thing as it did in Vue 2, and this test will fail.

https://v3.vuejs.org/guide/migration/attribute-coercion.html


lmiller1990:
Wow, I did not know that. That is really unintuitive, what a weird change.

Thanks!

connorshea:
I learned it the hard way while moving to Vue 3, so I noticed it immediately while browsing these docs  😅

Issue #638, opened 2021-06-01
2.0.0-rc.6 Poor error handling with promises & empty wrapper 


Giwayume:
### Subject of the issue

If I resolve an empty DOMWrapper as the return value of a promise, which is then awaited by an async function, get this error:

```
Cannot call then on an empty DOMWrapper.
```

### Steps to reproduce

```
import { defineComponent, nextTick } from 'vue';
import { mount } from '@vue/test-utils';

async function findAfterNextTick(wrapper, selector) {
    await nextTick();
    return wrapper.find(selector);
}

it('some test', async () => {
    const wrapper = mount(
        defineComponent({
            template: `<div>My component</div>`
        })
    );
    const foundElement = await findAfterNextTick(wrapper, '.something-that-does-not-exist');
    expect(foundElement.exists()).toBeFalsy();
});
```

### Expected behaviour

The test above should pass.

### Actual behaviour

Getting this error.

```
Cannot call then on an empty DOMWrapper.
```

### Possible Solution

This is way too generic.
https://github.com/vuejs/vue-test-utils-next/blob/ecf7905fd36a707174421f5c58b57e000faaf3e0/src/errorWrapper.ts#L11

I'm assuming nodejs is testing if the returned object is thenable, which causes this error. Fix so it's only checking for wrapper properties, [just like V1 of @vue/test-utils](https://github.com/vuejs/vue-test-utils/blob/dev/packages/test-utils/src/error-wrapper.js). Denying all properties is going to lead to weird bugs like this. It may be nice for brevity in your codebase, but it wasn't a smart change.

lmiller1990:
How about this? https://github.com/vuejs/vue-test-utils-next/pull/641. I added your test case.

Giwayume:
That works too, I just worry there's always something you're going to be missing by piecemealing this way.

lmiller1990:
I'm not sure what "piecemealing" means, but this seems okay. If we run into more problems, we can consider something similar to VTU v1.

Giwayume:
Piecemeal is to hand-pick, as in add excluded properties as problems are found. I prefer to write code that is safe under unknown conditions, which is the way v1 worked.

lmiller1990:
Thanks for clarifying that. I see your perspective too; if anything else comes up around this, let's revert to the more boilerplate heavy but reliable style of writing out each method.

I will release this patch soon.

PR #641, opened 2021-06-02
fix: better errors for async find

lmiller1990:
resolves https://github.com/vuejs/vue-test-utils-next/issues/638

Eh, a bit "smart" but avoids having to duplicate [everything like in Test Utils v1](https://github.com/vuejs/vue-test-utils/blob/dev/packages/test-utils/src/error-wrapper.js), which over the years has become tough to maintain.

### DIFF
--- a/src/errorWrapper.ts
+++ b/src/errorWrapper.ts
@@ -4,6 +4,10 @@ export function createWrapperError<T extends object>(
   return new Proxy<T>(Object.create(null), {
     get(obj, prop) {
       switch (prop) {
+        case 'then':
+          // allows for better errors when wrapping `find` in `await`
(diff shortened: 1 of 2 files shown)


src/errorWrapper.ts
```diff
@@ -4,6 +4,10 @@ export function createWrapperError<T extends object>(
   return new Proxy<T>(Object.create(null), {
     get(obj, prop) {
       switch (prop) {
+        case 'then':
+          // allows for better errors when wrapping `find` in `await`
+          // https://github.com/vuejs/vue-test-utils-next/issues/638
+          return
```
lmiller1990, line 10: I wonder if we need `catch`, too. Not sure if there is a use case.

cexbrayat (approved):

PR #645, opened 2021-06-03
chore: add volar check

cexbrayat:
This commit adds a check with volar to see if everything compiles fine.
This will allow to catch regressions when bumping Vue.

Currently, the ClassComponent test does not compile with Volar, so I'll open an issue to see if this is normal. 
The component is excluded from the check for now. 
(edit: this is not an issue see https://github.com/johnsoncodehk/volar/issues/245 I fixed the component in a second commit)

This also fixes a bunch of tiny things to make the PR green.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,3 +32,4 @@ jobs:
         CI: true
     - run: yarn test:build
     - run: yarn tsd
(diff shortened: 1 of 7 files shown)


lmiller1990 (approved):
LGTM!

PR #660, opened 2021-06-11
fix: Gracefully handle empty root node for nested component

LinusBorg:
This PR covers a kind of edge case.

Conditions: 

1. The component to test is nested in an element (so the parent element is not the mount wrapper)
2. the component doesn't render anything (so `$el` is a comment node)
3. The component is extracted with `wrapper.findComponent(Component)`
4. The test tries to use `extractedWrapper.find()` or `ExtractedWrapper.findAll()`

This will fail with an error being thrown:
> `TypeError: this.element.querySelector(All) is not a function`

This PR handles this case in a graceful way.

### DIFF
--- a/src/vueWrapper.ts
+++ b/src/vueWrapper.ts
@@ -123,7 +123,7 @@ export class VueWrapper<T extends ComponentPublicInstance>
     const result = this.parentElement['__vue_app__']
       ? // force using the parentElement to allow finding the root element
         this.parentElement.querySelector(selector)
-      : this.element.querySelector(selector)
+      : this.element.querySelector && this.element.querySelector(selector)
 
     if (result) {
       return new DOMWrapper(result)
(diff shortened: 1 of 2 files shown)


LinusBorg:
Huh, tests failed to run because of styles issues. Do I need to run prettier manually?

lmiller1990 (approved):
LGTM!

Yep, I just run prettier manually.

src/vueWrapper.ts
```diff
@@ -223,7 +223,9 @@ export class VueWrapper<T extends ComponentPublicInstance>
   findAll(selector: string): DOMWrapper<Element>[] {
     const results = this.parentElement['__vue_app__']
       ? this.parentElement.querySelectorAll(selector)
-      : this.element.querySelectorAll(selector)
+      : this.element.querySelectorAll
+      ? this.element.querySelectorAll(selector)
+      : ([] as unknown as NodeListOf<Element>)
```
cexbrayat, line 228: nit: is this cast really necessary here? I would have thought that returning a simple array would be enough.

cexbrayat: I gave it a quick try, and it looks like it's unnecessary. I'll make a PR to remove it.

cexbrayat: My bad: it _was_ necessary!

LinusBorg: Yeah, tsd doesn't like it :-P

PR #696, opened 2021-06-27
fix: Allow finding component by definition when using shallow mount

xanf:
ATM finding stubbed component is extremely inconsistent. It depends on "register name" in `components: {}` section, if component has `name` itself and so on, making it hard-to-guess if it will work and creating a regression,  compared to @vue/test-utils 1.x

This PR introduces a cache (using `WeakMap`, so we do not have to care about cleanup) and makes `matches` aware of stub

### DIFF
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -13,6 +13,7 @@ import {
 import { hyphenate } from './utils/vueShared'
 import { MOUNT_COMPONENT_REF, MOUNT_PARENT_NAME } from './constants'
 import { matchName } from './utils/matchName'
+import { isComponent, isFunctionalComponent } from './utils'
 import { ComponentInternalInstance } from '@vue/runtime-core'
 
 interface StubOptions {
@@ -21,6 +22,12 @@ interface StubOptions {
   propsDeclaration?: any
   renderStubDefaultSlot?: boolean
 }
+
+const stubsMap: WeakMap<ComponentOptions, VNodeTypes> = new WeakMap()
+
+export const getOriginalVNodeTypeFromStub = (
+  type: ComponentOptions
+): VNodeTypes | undefined => stubsMap.get(type)
 
 export const createStub = ({
   name,
@@ -99,12 +106,6 @@ const isMountedComponent = (
   type: VNodeTypes,
   props: ({ [key: string]: unknown } & VNodeProps) | null | undefined
 ) => isComponent(type) && props && props['ref'] === MOUNT_COMPONENT_REF
-
-const isComponent = (type: VNodeTypes): type is ComponentOptions =>
-  typeof type === 'object'
-
-const isFunctionalComponent = (type: VNodeTypes): type is ComponentOptions =>
-  typeof type === 'function' && ('name' in type || 'displayName' in type)
 
 export function stubComponents(
   stubs: Record<any, any> = {},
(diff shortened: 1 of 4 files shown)


src/utils.ts
```diff
@@ -68,17 +68,19 @@ export const mergeDeep = (
...
 
-export function isClassComponent(component: any) {
-  // TypeScript
-  return (
-    component.toString().includes('_super.apply(this, arguments) || this') ||
-    // native ES6
-    (typeof component === 'function' &&
-      /^\s*class\s+/.test(component.toString()))
-  )
+export function isClassComponent(component: VNodeTypes) {
```
xanf, line 71: while we're here, let's clean up all three helpers and get rid of `any`

Current approach of `isClassComponent` is super-flaky and does not work if (for example) babel instruments code after TypeScript or you're running instanbul or Wallaby for code coverage (which instruments code) 

New approach relies on [implementation detail](https://github.com/vuejs/vue-class-component/blob/next/src/vue.ts#L194) of `vue-class-component` but feels more reliable, since it will not accept any class and if `vue-class-component` will change implementation in future tests will warn us

Feel free to discard this change if you don't feel it is correct

src/utils/find.ts
```diff
@@ -28,47 +33,47 @@ export function matches(
     return node.el?.matches?.(selector)
   }
 
-  const nodeType = node.type
-  if (typeof selector === 'object' && typeof nodeType === 'object') {
-    // we are looking for this exact component
-    if (selector === nodeType) {
```
xanf, line 34: This case is actually already covered on top

src/utils.ts
lmiller1990: This seems more reliable - we (Vue team) control the `vue-class-component` library, and core has some code that also relies on `__vccOpts` existing.

lmiller1990 (approved):
Great, thanks for all the PRs! GitLab is probably the biggest Vue x VTU code base there is, so if we can get VTU 2.x working on GL, then we can be fairly confident it works for most other use cases, too.

src/stubs.ts
```diff
@@ -22,6 +23,12 @@ interface StubOptions {
   renderStubDefaultSlot?: boolean
 }
 
+const stubsMap: WeakMap<ComponentOptions, VNodeTypes> = new WeakMap()
```
lmiller1990, line 26: I wonder if we should export this and do a `stubsMap.clear()` each time we mount a component? Although `WeakMap` does save the effort of cleaning up, it might be more reliable to additionally blow it away ourselves. What do you think?

lmiller1990 (commented):
One question, but looking good

src/stubs.ts
xanf: I've also thought about that, but would like to avoid this. If we have two wrappers mounted in parallel, and we interact with them, since stubbing occurs on each VNode tree update clearing will result in unexpected behaviour, since `stubsMap` is shared across all instances of "apps" created by `mount`

I definitely do not want (at least for now) making some "context" for separate trees and pushing it elsewhere and having "cleanup" happening in forced & uncontrolled manner (uncontrolled because we don't know when next `mount` is called) makes me believe I'm introducing more complexity than I should

lmiller1990: Alright lets go with this. If we notice any weirdness in the future, we can look into it.

PR #697, opened 2021-06-27
fix: do not stub slots content when using shallow mount

xanf:
Current implementation of `stubComponents` is extremely agressive and really stubs everything including wrappers created for slots and `SlotWrapper` component used internally

Fix this by adding set of components, which should not be stubbed and "whitelisting" SlotWrapper component and components created for slots

I find this approach way more clearer in terms of types, than adding extra field like "doNotStub" to any component, like it was done in @vue/test-utils 1.x

### DIFF
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -8,7 +8,8 @@ import {
   ComponentOptions,
   defineComponent,
   VNodeProps,
-  VNodeTypes
+  VNodeTypes,
+  ConcreteComponent
 } from 'vue'
 import { hyphenate } from './utils/vueShared'
 import { MOUNT_COMPONENT_REF, MOUNT_PARENT_NAME } from './constants'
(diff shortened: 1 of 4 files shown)


lmiller1990:
What happens if you pass a component as the content of the `<slot>`? Is it stubbed (it should be, right?)

For example, what happens in this scenario?

```js
it('correctly renders slot content', () => {
    const ComponentWithSlot = defineComponent({
      template: '<div><slot></slot></div>'
    })

    const Foo = defineComponent({ template: '<div>OK</div>' })

    const wrapper = shallowMount(ComponentWithSlot, {
      components: { Foo },
      slots: {
        default: '<Foo />
      }
    })
  })
```

xanf:
> I'd expect to get `<div><foo-stub></foo-stub></div>` given `shallowMount` should stub out any components.

image

Works as intended! We un-stub only "magic components" generated by us, this is other code path

lmiller1990:
Great! I was actually expecting `<anonymous-stub>` or `<foo-stub>`. Could you add `name: 'Foo'`? Does that give us `<foo-stub>`? Also, can you add this test to the code base? 

Thanks!

xanf:
@lmiller1990 it's because it has no source of truth for name :) we could potentially look for stub name in `global.components` too but this is subject for another PR

image

Added test case & rebased branch

PR #698, opened 2021-06-27
feat: add @vue/compat initial support

xanf:
Vue 3.1 brings compat build, which is extremely important milestone in migrating huge codebases to Vue3.
`@vue/test-utils` plays surprisingly well with compat build and codebase in compat mode with only few changes required in code

Why changes are required? [Migration guide](https://v3.vuejs.org/guide/migration/migration-build.html) suggests enabling features via `configureCompat` as part of certain migration steps. If `RENDER_FUNCTION` feature is enabled, signature of render function will be similar to Vue2, so our stubs will not render slots. Luckily, we can override it, by explicitly setting `compatConfig` on our stub components (this has no effect if no compat build is in play).

Additionally a few magic hops are required for example for finding components declared with `Vue.extend`.

Please, take a note - this is not a complete changes list (for example `setValue` should take into account that model event might be set back to legacy "input" if `COMPONENT_V_MODEL` flag is in play), but is a foundation introducing testing compat build to CI and a couple of fixes

### DIFF
--- a/src/utils/vueCompatSupport.ts
+++ b/src/utils/vueCompatSupport.ts
@@ -1,15 +1,21 @@
+import * as Vue from 'vue'
+import type { ComponentOptions } from 'vue'
 import { FindAllComponentsSelector } from '../types'
 
+function isCompatEnabled(key: string): boolean {
+  return (Vue as any).compatUtils?.isCompatEnabled(key) ?? false
+}
+
 export function convertLegacyVueExtendSelector(
-  selector: any
+  selector: FindAllComponentsSelector
 ): FindAllComponentsSelector {
-  if (
-    typeof selector === 'function' &&
-    selector.name === 'SubVue' &&
-    selector.options
-  ) {
-    return selector.options
+  if (!isCompatEnabled('GLOBAL_EXTEND') || typeof selector !== 'function') {
+    return selector
   }
 
-  return selector
+  // @ts-ignore Vue.extend is part of Vue2 compat API, types are missing
+  const fakeCmp = Vue.extend({})
+
+  // @ts-ignore TypeScript does not allow access of properties on functions
+  return fakeCmp.super === selector.super ? selector.options : selector
 }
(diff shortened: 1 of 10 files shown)


src/utils/vueCompatSupport.ts
```diff
@@ -0,0 +1,15 @@
+import { FindAllComponentsSelector } from '../types'
+
+export function convertLegacyVueExtendSelector(
+  selector: any
+): FindAllComponentsSelector {
+  if (
+    typeof selector === 'function' &&
+    selector.name === 'SubVue' &&
```
lmiller1990, line 8: I'm not super familiar with how the compat build works - what is `SubVue`? Where does this come from?

lmiller1990 (commented):
Awesome, great to hear things mostly work with the compat build. Just one question!

src/utils/vueCompatSupport.ts
xanf: It comes from here: https://github.com/vuejs/vue-next/blob/51d2be20386d4dc59006d31a1cc96676871027ce/packages/runtime-core/src/compat/global.ts#L229

Unfortunately there is no 100% reliable way to check that "component is coming from legacy extend" but I find `SubVue` name matching reliable enough to work in 99.9% cases (come-on, if you name your functional component `SubVue` you're in trouble :))

src/utils/vueCompatSupport.ts
```diff
@@ -1,15 +1,21 @@
...
+  if (!isCompatEnabled('GLOBAL_EXTEND') || typeof selector !== 'function') {
+    return selector
   }
 
-  return selector
+  // @ts-ignore Vue.extend is part of Vue2 compat API, types are missing
+  const fakeCmp = Vue.extend({})
+
+  // @ts-ignore TypeScript does not allow access of properties on functions
+  return fakeCmp.super === selector.super ? selector.options : selector
```
lmiller1990, line 20: Nice

lmiller1990 (approved):

PR #699, opened 2021-06-27
feat: support passing component with template to slots

xanf:
Pretty self-describing :)

Recover functionality which was available in v1 :)

### DIFF
--- a/docs/guide/advanced/slots.md
+++ b/docs/guide/advanced/slots.md
@@ -134,8 +134,6 @@ test('layout full page layout', () => {
 })
 ```
 
-Note: passing a component using `{ template: '<div /> }` is not supported. Use a HTML string, render function, plain text, or an SFC.
-
 [Refer to the tests](https://github.com/vuejs/vue-test-utils-next/blob/9d3c2a6526f3d8751d29b2f9112ad2a3332bbf52/tests/mountingOptions/slots.spec.ts#L124-L167) for more examples and use cases.
 
 ## Scoped Slots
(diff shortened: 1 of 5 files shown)


docs/guide/advanced/slots.md
```diff
@@ -134,8 +134,6 @@ test('layout full page layout', () => {
 })
 ```
 
-Note: passing a component using `{ template: '<div /> }` is not supported. Use a HTML string, render function, plain text, or an SFC.
```
afontcu, line 137: What do you think about adding an example? Just adding a simple case above (lines 127 and 133) would do

afontcu (approved):
LGTM! Added a small suggestion for docs

lmiller1990 (approved):
Nice, can't believe we forgot this use case.

docs/guide/advanced/slots.md
lmiller1990: Does anyone remember why this was not supported? 🤔

xanf: @afontcu makes sense, will do it ASAP! Also found a couple more places to clean up comments :)

afontcu:
> Does anyone remember why this was not supported? 🤔

I can't remember of any good reason other than setting the initial structure of the stubbing mechanism was complex enough that we decided to go with the simple solutions. That one seemed harder than it was 😇

docs/guide/advanced/slots.md
lmiller1990: LMK when this is done, we can merge it up. Thanks!

xanf: @lmiller1990 @afontcu Done :rocket: !

PR #702, opened 2021-06-28
fix: opt-out of stubbing by passing false as stub

xanf:
@vue/test-utils 1 [allowed to un-stub certain component](https://github.com/vuejs/vue-test-utils/blob/dev/test/specs/mounting-options/stubs.spec.js#L299-L310) by passing `false` as stub configuration

Recover and document this behavior

### DIFF
--- a/tests/shallowMount.spec.ts
+++ b/tests/shallowMount.spec.ts
@@ -91,6 +91,28 @@ describe('shallowMount', () => {
     )
   })
 
+  it('stubs all components, but allows disabling stub by passing false', () => {
+    const wrapper = mount(ComponentWithChildren, {
+      shallow: true,
+      global: {
+        stubs: {
+          Hello: false
+        }
+      }
+    })
+    expect(wrapper.html()).toEqual(
+      '<div class="ComponentWithChildren">\n' +
+        '  <div id="root">\n' +
+        '    <div id="msg">Hello world</div>\n' +
+        '  </div>\n' +
+        '  <component-with-input-stub></component-with-input-stub>\n' +
+        '  <component-without-name-stub></component-without-name-stub>\n' +
+        '  <script-setup-stub></script-setup-stub>\n' +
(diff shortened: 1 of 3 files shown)


tests/shallowMount.spec.ts
```diff
@@ -91,6 +91,28 @@ describe('shallowMount', () => {
...
+    const wrapper = mount(ComponentWithChildren, {
+      shallow: true,
+      global: {
+        stubs: {
+          Hello: false
+        }
+      }
+    })
+    expect(wrapper.html()).toEqual(
+      '<div class="ComponentWithChildren">\n' +
```
lmiller1990, line 104: I wonder if we can use the `inlineSnapshot` feature from Jest here? Or do you think the string assertion is easier to read?

Thanks for all these PRs - let's do a release as soon as they are all merged up.

xanf: I prefer to keep this as it is for now - just for consistency with style of other tests here. We could probably address it in separate clean-up MR

xanf: A bit offtopic, regarding release. I believe I have some more (~5-10 MR coming) somewhere in next two days. I'm working (for fun) on building a prototype migration of `gitlab-ui` to Vue3 using compat build and it is surprisingly satisfying. 

With new plugin system in play I was able to mimic a huge precent of old VTU v1 behavior in pretty clean manner (for now we would like to have tests which run on both versions of VTU and)

lmiller1990: Ok, happy to keep this as-is.

If you have 5-10 more PRs coming, we can wait until they are all merged up before doing a release.

lmiller1990 (approved):

tests/shallowMount.spec.ts
xanf: @lmiller1990 Sure, I'll drop a comment in "last PR" for letting you know. thank you for acting super-fast

PR #704, opened 2021-06-28
fix: find component by its definition when using stub without name

xanf:
Follow-up of #696 

When we're stubbing component (no matter with `mount` or `shallow: true`) we still want to find a component by searching original component definition

### DIFF
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -24,7 +24,8 @@ interface StubOptions {
   renderStubDefaultSlot?: boolean
 }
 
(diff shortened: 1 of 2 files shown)


afontcu (approved):
LGTM!

I was concerned with the situation where a not-stubbed component that doesn't render (by using a `v-if="false"` at root, for instance) might be found by using a stub. I checked it out and turns out we can already find() a _hidden_ component, so nothing to worry about!

xanf:
Rebased, should be ready to go

lmiller1990 (approved):

PR #705, opened 2021-06-28
fix: findComponent should work using same stub for different components

xanf:
This one is a bit tricky. Sometimes we're reusing same stub for multiple components (it's fine when it is some kind of dummy stub). In this case previous naive approach taken in #704 will fail because we're having `stub -> type` mapping with stub as key.

To fix this, we're now generating unique stub (a very thin wrapper), making sure we could distinguish "stubs" for componentA and componentB when they are using same stub in `mount`

Marking as draft before #704 is merged

### DIFF
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -31,7 +31,7 @@ interface StubOptions {
 const stubsMap: WeakMap<ConcreteComponent, VNodeTypes> = new WeakMap()
 
 export const getOriginalVNodeTypeFromStub = (
-  type: ComponentOptions
+  type: ConcreteComponent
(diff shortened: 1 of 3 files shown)


lmiller1990:
#704 is merged. Ok to review this now?

xanf:
@lmiller1990 sure, I will rebase this one in an hour

xanf:
@lmiller1990 we are ready to go!

afontcu:
I guess this needs some rebasing after #708!

xanf:
@afontcu you're right, done!

afontcu (approved):
smart!

PR #714, opened 2021-06-30
fix(shallow): correctly stub anonymous components

xanf:
#607 [introduced](https://github.com/vuejs/vue-test-utils-next/pull/607/files#diff-840c8e77baceaebec1b759efa8bbb8505149fbd81c64bb120801388b66d839c4R162) a bug, when anonymous component will not be stubbed even when using `shallow: true`

Remove this special case - now it is properly handled by rest of the code (the only difference is that stubs will be now named as `anonymous-stub` which is kind of expected)


lmiller1990 (approved):

lmiller1990:
Great, thanks

PR #718, opened 2021-06-30
fix(compat): do not stub legacy functional component root

xanf:
This originally started as fixing obvious issue - we should not stub root when it is legacy functional component (`{ functional: true }`)

But it turns out that instead of adding more & more conditions we could switch to explicitly adding our Parent & root component to do-not-stub list and this greatly simplifies logic


afontcu (approved):
neat, the stubs map keeps proving useful PR after PR 🎉

PR #736, opened 2021-07-06
docs: findAll().at() migration notes

### LINKED ISSUE #693: docs: Migration findAll.at(0) => findAll[0]
The migration guide should make it obvious that `.at(0)` has been removed from `.findAll`

* **Before:**
  * `wrapper.findAll('[data-test="token"]').at(0)`
* **After:**
  * `wrapper.findAll('[data-test="token"]')[0]`

* * *

The new API docs for `.findAll` should include an example of selecting a specific item.

```js
const wrapper = mount(BaseTable);
const thirdRow = wrapper.findAll('tr')[2];
```

TheJaredWilcurt:
fixes #693

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -1056,12 +1056,13 @@ findAll(selector: string): DOMWrapper<Element>[]
 
 ```js
 import { mount } from '@vue/test-utils'
-import Component from './Component.vue'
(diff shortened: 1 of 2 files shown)


cexbrayat (approved):
Thanks @TheJaredWilcurt 👍

PR #743, opened 2021-07-08
fix: correctly pass component definition as slot mounting option

### LINKED ISSUE #727: lifecycle methods in child components not invoked
Lifecycle methods in child components seem to be not invoked.

```js
  const Parent = {
    template: `
      <div>
        <p>I am a parent</p>
        <slot />
      </div>
    `,
  };
  const Child = {
    created() {
      throw new Error("error?");
      console.log("child created");
    },
    mounted() {
      console.log("child mounted");
    },
    render() {
      return "i am a child";
    },
  };

  const wrapper = mount(Parent, {
    slots: { default: Child },
  });

  console.log(wrapper.html());
```

↑ https://github.com/eunjae-lee/vue-test-utils-snapshot-2/blob/master/src/Test.spec.ts#L11-L37

I expect console log from `Child` or even `Error` being thrown at `created`, but I don't see anything and it successfully(?) prints out the html.

However, this workaround works:

```js
  const wrapper = mount({
    components: {Parent, Child},
    template: `
      <Parent>
        <Child />
      </Parent>
    `
  })
```

lmiller1990:
We were not passing the entire component to `slots`, just the `render` function, so things like the lifecycle hooks were not executed.

closes #727

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -263,7 +263,7 @@ export function mount(
   function slotToFunction(slot: Slot) {
     if (typeof slot === 'object') {
       if ('render' in slot && slot.render) {
-        return slot.render
+        return () => h(slot)
       }
 
       if ('template' in slot && slot.template) {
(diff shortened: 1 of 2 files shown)


cexbrayat (approved):
👍

eunjae-lee:
This looks good to me. Thank you 🙌🏼

PR #767, opened 2021-07-16
fix(setData): do not concat previous and new array values

### LINKED ISSUE #766: `setData` wrongly deep-merges
`setData` wrongly deep-merges.

```js
const Comp = {
  template: `
    <div>{{state.items.join(",")}}</div>
  `,
  data() {
    return {
      state: {
        items: ['1'],
      },
    };
  },
};

it('works in an object form', async () => {
  const wrapper = mount(Comp);
  expect(wrapper.html()).toMatchInlineSnapshot(`<div>1</div>`);

  await wrapper.setData({ state: { items: ['2', '3'] } });
  expect(wrapper.html()).toMatchInlineSnapshot(`<div>2,3</div>`);
});
```

The second `expect` clause is supposed to pass, but it isn't. The actual result is `<div>1,2,3</div>`.

[The implementation of setData](https://github.com/vuejs/vue-test-utils-next/blob/52b3777310dc8037a860805e79f2b75061aec8d9/src/vueWrapper.ts#L238-L241) seems to concat all the items in the array.

## reproduction
https://github.com/eunjae-lee/vtu-set-data

```
git clone git@github.com:eunjae-lee/vtu-set-data.git
cd vtu-set-data
yarn install
yarn test
```

↑ [the test](https://github.com/eunjae-lee/vtu-set-data/blob/main/tests/unit/example.spec.ts) fails

eunjae-lee:
fixes #766

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -57,7 +57,7 @@ export const mergeDeep = (
     const sourceValue = source[key]
 
     if (Array.isArray(targetValue) && Array.isArray(sourceValue)) {
-      target[key] = targetValue.concat(sourceValue)
+      target[key] = sourceValue
     } else if (isObject(targetValue) && isObject(sourceValue)) {
       target[key] = mergeDeep(Object.assign({}, targetValue), sourceValue)
     } else {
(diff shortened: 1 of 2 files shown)


cexbrayat (approved):
LGTM, thanks for the PR.

I'll wait for @afontcu review before merging, as maybe it was using `concat` for a purpose.

afontcu (approved):
thanks for this!

PR #777, opened 2021-07-18
fix(compat): Do not wrap legacy options component with extra wrap

xanf:
New day, new compat issue (this one discovered by working on migrating `bootsrap-vue` to compat build)

We need to unwrap component before mount, or (since typeof component will be function after `Vue.extend`)) it will be treated as function component and wrapped inside another one with "setup"

### DIFF
--- a/tests-compat/compat.spec.ts
+++ b/tests-compat/compat.spec.ts
@@ -130,4 +130,17 @@ describe('@vue/compat build', () => {
 
     expect(wrapper.html()).toBe('<div>stubbed</div>')
   })
+
+  it('wrapper.vm points to correct instance when component is wrapped with Vue.extend', () => {
+    const Component = extend({
+      data() {
+        return { foo: 'bar' }
+      },
+      template: '<div></div>'
+    })
+
+    const wrapper = mount(Component)
+
+    expect(wrapper.vm.foo).toBe('bar')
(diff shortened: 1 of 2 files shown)


tests-compat/compat.spec.ts
```diff
@@ -130,4 +130,18 @@ describe('@vue/compat build', () => {
...
+
+  it('wrapper.vm points to correct instance when component is wrapped with Vue.extend', () => {
+    const Component = extend({
+      data() {
+        return { foo: 'bar' }
+      },
+      template: '<div></div>'
+    })
+
+    debugger
```
AlexeyKorkoza, line 142: @xanf , Probably, you forgot to remove the debugger statement

xanf: Lol. thank you for catching!

lmiller1990 (approved):
Thanks for finding all these issues!

lmiller1990:
Just did a release, but we can do another one in the next day or two to get this out.

xanf:
Thank you!

No rush, I'm working on couple big ones this week

ср, 28 июл. 2021 г., 03:20 Lachlan Miller ***@***.***>:

lmiller1990:
I will wait on those before releasing then 👍

PR #789, opened 2021-07-23
docs: update name

### LINKED ISSUE #787: Wrong config property name for `renderDefaultStubSlot` into migration guide
Package types and API mention `renderStubDefaultSlot`, while migration guide uses `renderDefaultStubSlot`.
https://next.vue-test-utils.vuejs.org/migration/#shallowmount-and-renderdefaultstubslot

Possibly a rename which has not been propagated

DominicGBauer:
Closes #787

## Issue
Wrong name for heading

## Fix
Update docs to use correct name

### DIFF
--- a/docs/migration/index.md
+++ b/docs/migration/index.md
@@ -143,7 +143,7 @@ const wrapper = mount(App, {
 }
 ```
 
-### `shallowMount` and `renderDefaultStubSlot`
+### `shallowMount` and `renderStubDefaultSlot`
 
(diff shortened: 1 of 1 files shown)


afontcu (changes requested):
Nice catch! I think we should also fix line 209 of this very same file because it also mentions `renderDefaultStubSlot`.

DominicGBauer:
@afontcu Ah missed that sorry, added the change now :)

afontcu (approved):
Looks great now! 🚀  thank you!

PR #811, opened 2021-07-30
chore: update to jest 27

cexbrayat:
Updates jest, ts-jest and babel-jest to v27.
vue-jest v5 does not support jest v27,
so a new vue3-jest package has been introduced (see https://github.com/vuejs/vue-jest/pull/343),
with a version number aligned on the jest version (as ts-jest and babel-jest do).

This commits replaces vue-jest by vue3-jest, and updates the jest config to use this new trasnformer.
This will be the default Vue CLI setup in the future.


cexbrayat:
This is currently blocked by https://github.com/vuejs/vue-next/pull/4216 in vue-next, as this update surfaced an issue when spying on proxy (it in fact occurs with jest v26.6.1, but we were using an older version than that)

cexbrayat:
facebook/jest#11721 landed, unblocking part of the issue with proxy spying.
Jest v27.1 has been released with the fix, and this PR now uses it, allowing to remove the dirty workaround we had to add `hasOwnProperty` (see the second commit).

Note that users will have to update to Jest v27.1 if they want to spy on the VM object, so I reflected it by adding Jest as an optional peer dependency with at least version 27.1. I'm happy to discuss it (or to push it to a separate PR if we want to).

We are still waiting on vuejs/vue-next#4216 to completely unblock this PR.

cexbrayat:
The proxy issue in core has been fixed and released in v3.2.31, so this PR can now be merged after review

lmiller1990 (approved):

PR #813, opened 2021-07-31
fix(stubs): Do not create stubs across multiple renders, fix #810

### LINKED ISSUE #810: Need to reassign stubbed component with findComponent after update to 2.0.0-rc.10
I am currently having the issue, that some of my tests fail after updating from `2.0.0-rc.9` to `2.0.0-rc.10`. When manually stubbing a component then I need to reassign the wrapper with `findComponent` to receive the correct values after making any update (e.g. emits or setProps).

After some checking I created a code snippet to test with. If you run the snippet with `2.0.0-rc.9` the tests passes. If running with `2.0.0-rc.10` the test "should test with stub" fails. In comparision with "should test with stub and reassign" there is only the reassign `child = wrapper.findComponent({ name: 'ChildComponent' })` missing which is not necessary when testing without manually stub or with `2.0.0-rc.9`.

```js

const ChildComponent = defineComponent({
  name: 'ChildComponent',
  props: {
    modelValue: {
      type: String,
      required: true
    }
  },
  template: '<span>{{ modelValue }}</span>'
})

const MainComponent = defineComponent({
  components: { ChildComponent },
  props: {
    modelValue: {
      type: String,
      required: true
    }
  },
  template: '<ChildComponent v-model="modelValue" />'
})

describe('test stub', () => {
  it('should test without stub', async () => {
    const wrapper = mount(MainComponent, {
      props: { modelValue: '1' }
    })

    const child = wrapper.findComponent({ name: 'ChildComponent' })
    expect(child.props('modelValue')).toBe('1')

    await wrapper.setProps({ modelValue: '2' })

    expect(child.props('modelValue')).toBe('2')
  })

  /**
   * Test stub component with reassign wrapper with findComponent
   * Works with rc.9 and rc.10
   */
  it('should test with stub and reassign', async () => {
    const wrapper = mount(MainComponent, {
      props: { modelValue: '1' },
      global: {
        stubs: {
          ChildComponent
        }
      }
    })

    let child = wrapper.findComponent({ name: 'ChildComponent' })
    expect(child.props('modelValue')).toBe('1')

    await wrapper.setProps({ modelValue: '2' })

    child = wrapper.findComponent({ name: 'ChildComponent' })
    expect(child.props('modelValue')).toBe('2')
  })

  /**
   * Test stub component without reassign wrapper with findComponent
   * Works with rc.9. Fails with rc.10
   */
  it('should test with stub', async () => {
    const wrapper = mount(MainComponent, {
      props: { modelValue: '1' },
      global: {
        stubs: {
          ChildComponent
        }
      }
    })

    const child = wrapper.findComponent({ name: 'ChildComponent' })
    expect(child.props('modelValue')).toBe('1')

    await wrapper.setProps({ modelValue: '2' })

    expect(child.props('modelValue')).toBe('2')
  })
})
```

xanf:
Fixes #810 

Basically we need to cache our stubs so we do not create new stubs each render.
We already had this thing in place long ago for our automatic stubs (created either by shallow or with stub value set to `true`)

This PR uses same logic and unifies "storage" of created stubs

### DIFF
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -126,12 +126,35 @@ const getComponentName = (type: VNodeTypes): string => {
   return ''
 }
 
+function createStubOnceForType(
+  type: {} & VNodeTypes,
+  factoryFn: () => ConcreteComponent,
+  cache: WeakMap<{} & VNodeTypes, ConcreteComponent>
+): ConcreteComponent {
+  const cachedStub = cache.get(type)
+  if (cachedStub) {
+    return cachedStub
+  }
+
+  const stub = factoryFn()
+  stubsMap.set(stub, type)
+  cache.set(type, stub)
+  return stub
+}
+
 export function stubComponents(
   stubs: Record<any, any> = {},
   shallow: boolean = false,
   renderStubDefaultSlot: boolean = false
 ) {
-  const components: Record<string, ComponentOptions> = {}
+  const createdStubsMap: WeakMap<{} & VNodeTypes, ConcreteComponent> =
+    new WeakMap()
+
+  const createStubOnce = (
+    type: {} & VNodeTypes,
+    factoryFn: () => ConcreteComponent
+  ) => createStubOnceForType(type, factoryFn, createdStubsMap)
+
   transformVNodeArgs((args, instance: ComponentInternalInstance | null) => {
     const [nodeType, props, children, patchFlag, dynamicProps] = args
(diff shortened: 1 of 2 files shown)


lmiller1990 (approved):
Nicely done. The stub madness begins again - so far this is much cleaner than V1, let's try to keep it that way. Thanks for this!

PR #815, opened 2021-08-01
fix: Always wrap string passed to slot in a helper component

xanf:
I've originally [encountered](https://github.com/bootstrap-vue/bootstrap-vue/blob/5bf6733/src/components/table/table-colgroup.spec.js#L31) this issue while porting bootstrap-vue to Vue3

All known (at least to me and in mentioned stack overflow :stuck_out_tongue: question) methods of checking if something is HTML reports false for `<col><col><col>` because `<col>` is considered valid tag only inside `table -> colgroup`

In order to fix that I simply removed check if it is HTML and always wrap slot provided as string into `<SlotWrapper>`
This has a downside - if code under test relies on knowing that slot will always have raw strings - it will be confused by wrapper. User could workaround it with passing slot as function:

```js
slots: {
  'some-slot': () => h('Something which should be just plain old string')
}
```

I consider this way more rare case, and what is more important - people who manipulate slot contents in that way are usually far more expert to understand the root cause of the issue and to discover (or find this PR) workaround.

For original issue - it took me a while to realize why slot work for certain HTML string and does not work for another when the only difference is tag

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -27,7 +27,6 @@ import {
 import { MountingOptions, Slot } from './types'
 import {
   isFunctionalComponent,
-  isHTML,
   isObjectComponent,
   mergeGlobalProperties,
   isObject
@@ -281,14 +280,7 @@ export function mount(
     }
 
     if (typeof slot === 'string') {
-      // if it is HTML we process and render it using h
-      if (isHTML(slot)) {
-        return (props: VNodeProps) => h(processSlot(slot), props)
-      }
-      // otherwise it is just a string so we just return it as-is
-      else {
-        return () => slot
-      }
+      return (props: VNodeProps) => h(processSlot(slot), props)
     }
 
     throw Error(`Invalid slot received.`)
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -93,20 +93,6 @@ export function isObjectComponent(
   return Boolean(component && typeof component === 'object')
 }
 
-// https://stackoverflow.com/questions/15458876/check-if-a-string-is-html-or-not/15458987#answer-15458968
-export function isHTML(str: string) {
-  var a = document.createElement('div')
-  a.innerHTML = str
-
-  for (let c = a.childNodes, i = c.length; i--; ) {
-    if (c[i].nodeType == 1) {
-      return true
-    }
-  }
-
-  return false
-}
-
 export function textContent(element: Element): string {
   // we check if the element is a comment first
   // to return an empty string in that case, instead of the comment content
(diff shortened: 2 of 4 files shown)


lmiller1990:
`isHTML` was introduced here: https://github.com/vuejs/vue-test-utils-next/pull/599 I am surprised this did not break one of those tests.

This would technically be a breaking change (at least for existing V2 users). Any chance you could write something I can include in the release notes explain the breaking change and how to work around it? Just a paragraph with a snippet or two would be great.

lmiller1990 (commented):
Seems reasonable, just left one comment - request for something to include in release notes explaining the breaking change.

xanf:
> This would technically be a breaking change (at least for existing V2 users). Any chance you could write something I can include in the release notes explain the breaking change and how to work around it? Just a paragraph with a snippet or two would be great.

Well, I'm not sure this is a breaking change. For me it feels more like a fix for perfectly valid use case:

```js
 it('allows passing a scoped slot via string with no HTML inside without template tag', () => {
      const wrapper = mount(ComponentWithSlots, {
        slots: {
          scoped: 'Just a plain {{ params.aBoolean }} {{ params.aString }}'
        }
      })

      expect(wrapper.find('.scoped').text()).toEqual('Just a plain true string')
    })
```

I've added this test in a recent commit.
WDYT of something like:

Note: if you're passing raw strings to a slot of the component - these are now always wrapped into helper component (previously wrapping occurred only if string contained HTML. If your component relies on raw string being passed to slot without any specific wrappers, pass render function to slot content wrapping your string in `h()`:

```js
 it('allows passing a scoped slot via string with no HTML inside without template tag', () => {
      const wrapper = mount(ComponentWithSlots, {
        slots: {
          // this one will be wrapped with helper component
          scoped: 'Just a plain string',
          // this one will be passed to slot as plain string
          otherScoped: () => h('Just a plain string'),
        }
      })

      expect(wrapper.find('.scoped').text()).toEqual('Just a plain true string')
    })
```

Feel free to edit this for making most easy-to-understand description

xanf:
@lmiller1990 Ignore my explanation, if https://github.com/vuejs/vue-test-utils-next/pull/844 this will effectively hide any "slot wrapper" stuff from end users, making this explanation wrong and unneeded

lmiller1990:
Do we still need this or will #844 take it's place?

lmiller1990 (approved):
LGTM I think - still a little uneasy, hope we won't break anything (eg if ppl are relying on some specific logic around `slots[0].children`).

Issue #716, opened 2021-06-30
Support `find({ ref })` syntax


kanryu:
### Subject of the issue

Calling findComponent () with the ref method causes a run-time error.

### Steps to reproduce

```html
<template>
  <div class="hello">
    <h1>{{ msg }}</h1>
    <button ref="some_ref" v-on:click="$emit('accepted')">OK</button>
  </div>
</template>
```

```js
import { mount } from '@vue/test-utils';
import Component from '@/components/HelloWorld.vue'

describe('button-exists', () => {
    const wrapper = mount(Component)

    it('ref check', () => {
        const some_ref = wrapper.findComponent({ref: 'some_ref'})
        expect(some_ref.exists()).toBe(true)
    })
})
```

- write a template like the one above
- call `wrapper.findComponent({ref: 'some_ref'})`

### Expected behaviour

findComponent() should return a valid node.

### Actual behaviour

runtime error.

```
    TypeError: Cannot read property 'emits' of undefined

      42 |     it('ref check', () => {
      43 |         console.log(wrapper)
    > 44 |         const some_ref = wrapper.findComponent({ref: 'some_ref'})
         |                                    ^
      45 |         expect(some_ref.exists()).toBe(true)
      46 |     })
      47 | })

      at VueWrapper.Object.<anonymous>.VueWrapper.attachNativeEventListener (node_modules/@vue/test-utils/dist/vue-test-utils.cjs.js:7225:33)
      at new VueWrapper (node_modules/@vue/test-utils/dist/vue-test-utils.cjs.js:7202:15)
      at createWrapper (node_modules/@vue/test-utils/dist/vue-test-utils.cjs.js:7384:12)
      at VueWrapper.findComponent (node_modules/@vue/test-utils/dist/vue-test-utils.cjs.js:7305:24)
      at Object.<anonymous> (tests/App.spec.js:44:36)
```

### Possible Solution

I can't comment because I don't know the detailed implementation of this framework, but when I checked it with the debugger, `VueWrapper.prototype.attachNativeEventListener()` is executed twice.

The first time, `vm` points to a Proxy and it runs success.
The second time it points to a HTMLButtonElement and it doesn't have a `$options` property, which causes a run-time error.

### versions

- "vue": "^3.0.0"
- "core-js": "^3.6.5",
- "@vue/compiler-sfc": "^3.0.0",
- "@vue/component-compiler-utils": "^3.2.2",
[...]

PR #817, opened 2021-08-01
fix: do not crash findComponent if ref is pointing to html element

xanf:
This is first part of fix related to #716 

After certain consideration, I've decided that expected behavior should be not throwing error, but returning empty wrapper instead.
My reasoning is based on three points:

1. Passing `ref` to `findComponent` is perfectly valid syntax. So if we're looking for component with specific ref and it does not exists - why should we complain about HTML element with same ref? We simply should not care
2. (Real world example from gitlab codebase). Consider having `<component ref="hello" :is="foo">` where `foo` under certain conditions is resolved either to `<div>` or to component (let's say `<Note>`). Consider having table syntax test, where we pass certain props... and if we throw, it definitely will look ugly and wrong instead of just checking for `.exists()`
3. `findComponent` does not have `throw` behavior and I'm resisting adding one (keeping mental mode simple - `getComponent` throws, `findComponent` - does not thow)

### DIFF
--- a/src/vueWrapper.ts
+++ b/src/vueWrapper.ts
@@ -159,8 +159,11 @@ export class VueWrapper<T extends ComponentPublicInstance>
   ): VueWrapper<T> {
     if (typeof selector === 'object' && 'ref' in selector) {
       const result = this.vm.$refs[selector.ref]
-      if (result) {
+      console.log(result)
+      if (result && !(result instanceof HTMLElement)) {
         return createWrapper(null, result as T)
+      } else {
+        return createWrapperError('VueWrapper')
       }
     }
 
--- a/src/vueWrapper.ts
+++ b/src/vueWrapper.ts
@@ -159,7 +159,6 @@ export class VueWrapper<T extends ComponentPublicInstance>
   ): VueWrapper<T> {
     if (typeof selector === 'object' && 'ref' in selector) {
       const result = this.vm.$refs[selector.ref]
-      console.log(result)
       if (result && !(result instanceof HTMLElement)) {
         return createWrapper(null, result as T)
       } else {
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -1143,6 +1143,10 @@ test('findComponent', () => {
 })
 ```
 
+:::warning
+If `ref` in component points to HTML element, `findComponent` will return empty wrapper. This is intended behaviour
+:::
+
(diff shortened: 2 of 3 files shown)


src/vueWrapper.ts
```diff
@@ -159,8 +159,11 @@ export class VueWrapper<T extends ComponentPublicInstance>
   ): VueWrapper<T> {
     if (typeof selector === 'object' && 'ref' in selector) {
       const result = this.vm.$refs[selector.ref]
-      if (result) {
+      console.log(result)
```
afontcu, line 162:
```suggestion
```

afontcu (approved):
Looks good to me! 🚀  @lmiller1990 since you expressed some concerns in https://github.com/vuejs/vue-test-utils-next/issues/716, do you believe this makes sense?

I also suggested throwing, but I agree that the differente between `find` and `get` is easier to grasp if the former never throws – even on edge cases.

src/vueWrapper.ts
xanf: @afontcu Thank you! One day I will remember to have another SPECIFIC look for console.logs and debugger :)

lmiller1990:
Seems fine! This is the V1 behavior, right? 

I really regret the decision to support both `findComponent` AND `getComponent`. The differences are so subtle, and it's really hard to remember with different libraries using `get`, `find`, and `query` all slightly differently.

xanf:
> My concern is that it'll be confusing to return an empty wrapper that doesn't actually do anything. Does this really provide any value other than some fairly specific cases? Eg, even if is returns a wrapper, you cannot do very much except assert exists (and there should be another way to write that same test). Should we not just return undefined (since there is no component with the ref you are querying for? This seems like the more logical behavior to me.

Well, if we're speaking in scope of this PR - it definitely should be an empty wrapper, just because this is what `findComponent` (and `find`) returns when no result is found, no matter which selector is used, and keeping this behavior consistent is pretty important IMO

If you're speaking about returning undefined instead of empty wrapper "in general" for any find operation - I do not have strong preference towards any of the approaches and both makes sense to me, but I know that at least some of my colleagues like "JQuery-like" behavior of `find / findComponent` - when you could always be sure you will have wrapper as a result even if it is empty

afontcu:
I think I also lean towards returning an empty Wrapper – as we've already pointed out, we should teach users to use `get(Component)` if they are certain an item is there (let's bear in mind we added `find(Component)` to allow people to make sure a component/node is present!)

lmiller1990:
Ok, I will trust you guys on this one 👍

PR #825, opened 2021-08-04
docs: Add learning links

afontcu:
This could be greatly improved in both quality and quantity, but just thought that it's better than having an empty page.


lmiller1990 (approved):
Thanks!

Issue #457, opened 2021-03-11
Mocked (vuex) store doesn't show up inside setup method of composition API


irongollem:
**Versions:** 
Vue: 3.0.5
@vue/test-utils: 2.0.0-rc.3 
jest: 26.6.3,

**Problem:**
When using the composition API I get a store.x => cannot get property x of undefined.
When using the option API using the same test there is no issue

**example:**

```javascript
// comp.spec.ts
const store = {
    state: {
      blocks: [],
    },
    dispatch: jest.fn(),
  };

const wrapper = shallowMount(Comp, {
  global: {
    mocks: [store]
  },
});
```

```javascript
// comp.vue working example

// ...
setup() {
  const store = useStore();
  return { store }; 
},
computed: {
  getSomething() { this.store.state.thing },
},
// ...

// OR

// comp.vue not working composition api example

// ...
setup() {
const store = useStore{};
const getSomething = computed(() => store.state.thing);
// ...
}
```

**Expected**
Mocked state should be available inside the setup method

lmiller1990:
I don't think this should work; you are not passing a plugin (which should have an `install` method), you are just passing a plain object.

See this example: https://github.com/vuejs/vue-test-utils-next/blob/84f875fbb6120ffaa597d7a702732e4f79be363d/tests/features/vuex.spec.ts#L15

Blackfaded:
You could also mock the store like so:

```ts
jest.mock('@/libs/store', () => {
  return {
    useStore: () => createStoreMock(),
  };
});
```
Just adjust the importpath and write a createStoreMock function like so:
```ts
import { cloneDeep } from 'lodash';
import { createStore } from 'vuex';

import { RootStore } from '@/libs/store';
const { modules } = jest.requireActual('@/libs/store');

export const createStoreMock = () => {
  return createStore({
    modules: cloneDeep(modules),
  }) as RootStore;
};

```

irongollem:
ah tnx guys, I will try these solutions! Would be nice we get some descriptor of this in the docs so people are made aware (even though you answer seems locical). It depends where you come from I guess, but since it was possible in the past I assumed it still should be :P

irongollem:
Not sure if this is really a bug or just some config or whatever on my side, but I tried both suggested solutions, and neither did the trick. 
* `jest.mock('@/store, () => ...)` didn't seem to do anything (I have the store in root) (i didnt clone the story as your example but tried created a very minimal one on the fly)
[...]

Issue #407, opened 2021-02-22
injection "store" not found


techlab23:
Hello, 

I'm trying to setup tests for the component using mock vuex store, however i'm running into an issue where i get the following warning and subsequent error in setup function of the component. 

Here is the sample component that i used to reproduce this issue

https://github.com/techlab23/task-management-app-ts/blob/tests/tests/App.spec.ts

```console.warn
      [Vue warn]: injection "store" not found. 
        at <Anonymous ref="VTU_COMPONENT" > 
        at <VTUROOT>

      16 |   components: { AppHeader, AppLoadingIndicator },
      17 |   setup() {
    > 18 |     const store = useStore();
         |                   ^
      19 |     store.dispatch('fetchData');
      20 |     const isLoading = computed<boolean>(() => store.getters.isLoading);
      21 |     return {

```

Following above  error i get another error in calling a dispatch function

```

App › calls fetchData to load initial data

    TypeError: Cannot read property 'dispatch' of undefined

      17 |   setup() {
      18 |     const store = useStore();
    > 19 |     store.dispatch('fetchData');
         |           ^
      20 |     const isLoading = computed<boolean>(() => store.getters.isLoading);
      21 |     return {
      22 |       isLoading,

      at setup (src/App.vue:19:11)

```
Appreciate any pointers is resolving this issue.

cexbrayat:
@techlab23 The error `injection "store" not found. ` indicates that in your current test there is not store provided.

You need to provide the store to your component, using https://next.vue-test-utils.vuejs.org/api/#global
Something like:
```ts
const wrapper = mount(App, { 
  global: { 
    provide: { 
      store: createStore(...) 
    } 
  } 
});
```

maurer2:
Hello, how would one provide a mock store when using an injection key with a custom useStore function as recommended here https://next.vuex.vuejs.org/guide/typescript-support.html#simplifying-usestore-usage ?

I can get the previous example and also another approach from here (https://github.com/vuejs/vue-test-utils-next/issues/457#issuecomment-799191975) running without an injection key, but once I use a key, the store can't be found. 

main
```ts
import { store, key } from './store/index';

createApp(App)
  .use(store, key)
  .mount('#app');
````

store
```ts
import { createStore, useStore as genericUseStore } from 'vuex';

const key: InjectionKey<Store> = Symbol('key');

function useStore(): Store {
  return genericUseStore(key);
}
```

[...]

PR #829, opened 2021-08-04
Added explanation for testing useStore/compositionAPI

maurer2:
Hello, I tried to document the approach for testing components that use the useStore-helper function with and without Typescript. Most information are taken from https://github.com/vuejs/vue-test-utils-next/issues/407 and https://github.com/vuejs/vue-test-utils-next/issues/457). 

Cheers

### DIFF
--- a/docs/guide/advanced/vuex.md
+++ b/docs/guide/advanced/vuex.md
@@ -181,56 +181,112 @@ By creating a `createVuexStore` function that takes an initial state, we can eas
 
 The [Vue Testing Handbook](https://lmiller1990.github.io/vue-testing-handbook/testing-vuex.html) has more examples for testing Vuex. Note: the examples pertain to Vue.js 2 and Vue Test Utils v1. The ideas and concepts are the same, and the Vue Testing Handbook will be updated for Vue.js 3 and Vue Test Utils 2 in the near future.
 
-## Testing a store when using the composition API
-When using the Composition API, a Vuex store is usually accessed via a `useStore` helper function. More about it here: https://next.vuex.vuejs.org/guide/composition-api.html.
-`useStore` can be used with an optional and unique injection key to get a properly typed store. To avoid repeating the key parameter passing whenever `useStore` is used, it is recommended to extract that logic into a helper function and reuse that function instead of the default `useStore`-function. More about it here: https://next.vuex.vuejs.org/guide/typescript-support.html#typing-usestore-composition-function.
-The approach to provide a store for a unit test depends on the way the `useStore` function is used in the component.
+## Testing using the Composition API
 
-### Testing components that utilize `useStore` without an injection key
+Vuex is accessed via a `useStore` function when using the Composition API. [Read more about it here](https://next.vuex.vuejs.org/guide/composition-api.html).
 
-Without an injection key, the store data can just be injected into the component via the global `provide` mounting option. The name of the injected store must be the same as the one in the component, e.g. "store". Instead of creating the store on the fly, it can also be imported from a file and therefore be shared between unit tests.
+`useStore` can be used with an optional and unique injection key as discussed [in the Vuex documentation](https://next.vuex.vuejs.org/guide/typescript-support.html#typing-usestore-composition-function).
+
+It looks like this:
+
+```js
+import { createStore } from 'vuex'
+import { createApp } from 'vue'
+
+// create a globally unique symbol for the injection key
+const key = Symbol()
+
+const App = {
+  setup () {
+    // use unique key to access store
+    const store = useStore(key)
+  }
+}
+
+const store = createStore({ /* ... */ })
+const app = createApp({ /* ... */ })
+
+// specify key as second argument when calling app.use(store)
+app.use(store, key)
+```
+
+To avoid repeating the key parameter passing whenever `useStore` is used, the Vuex documentation recommends extracting that logic into a helper function and reuse that function instead of the default `useStore` function. [Read wore about it here](https://next.vuex.vuejs.org/guide/typescript-support.html#typing-usestore-composition-function). The approach providing a store using Vue Test Utils depends on the way the `useStore` function is used in the component.
+
+### Testing Components that Utilize `useStore` without an Injection Key
+
+Without an injection key, the store data can just be injected into the component via the global `provide` mounting option. The name of the injected store must be the same as the one in the component, e.g. "store". 
 
 #### Example for providing the unkeyed `useStore`
-```ts
+
+```js
 import { createStore } from 'vuex'
 
-const myStore = createStore({
+const store = createStore({
   // ...
 })
 
 const wrapper = mount(App, {
   global: {
-      provide: {
-        store: myStore
-      },
+    provide: {
+      store: store
     },
+  },
 })
 ```
 
-### Testing components that utilize `useStore` with an injection keys
+### Testing Components that Utilize `useStore` with an Injection Key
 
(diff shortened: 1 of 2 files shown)


docs/guide/advanced/vuex.md
```diff
@@ -181,6 +181,48 @@ By creating a `createVuexStore` function that takes an initial state, we can eas
 
 The [Vue Testing Handbook](https://lmiller1990.github.io/vue-testing-handbook/testing-vuex.html) has more examples for testing Vuex. Note: the examples pertain to Vue.js 2 and Vue Test Utils v1. The ideas and concepts are the same, and the Vue Testing Handbook will be updated for Vue.js 3 and Vue Test Utils 2 in the near future.
 
+## Testing a store when using the composition API
+When using the composition API, the Vuex store is usually accessed via a `useStore`-helper function. In Javascript projects the built-in `useStore` function should be used. More about it here: https://next.vuex.vuejs.org/guide/composition-api.html.
```
lmiller1990, line 185:
```suggestion
When using the Composition API, a Vuex store is usually accessed via a `useStore` helper function. In JavaScript projects the built-in `useStore` function should be used. More about it here: https://next.vuex.vuejs.org/guide/composition-api.html.
```

lmiller1990: I don't understand what "In JavaScript projects the built-in `useStore` function should be used" means. Isn't this the same information as the first sentence?

docs/guide/advanced/vuex.md
```diff
@@ -181,6 +181,48 @@ By creating a `createVuexStore` function that takes an initial state, we can eas
...
+In TypeScript projects the `useStore`-function can be used with an optional and unique injection key to get a properly typed store. To avoid repeating the key parameter passing whenever `useStore` is used, it is recommended to extract that logic into a helper function and reuse that function instead of the default `useStore`-function. More about it here: https://next.vuex.vuejs.org/guide/typescript-support.html#typing-usestore-composition-function.
```
lmiller1990, line 186: The API and usage should be the same in both TS and JS - I am not sure we need language specific information. At the end of the day, it's the same code and API.

docs/guide/advanced/vuex.md
```diff
@@ -181,6 +181,48 @@ By creating a `createVuexStore` function that takes an initial state, we can eas
...
+The approach to provide a store for a unit test depends on the way the `useStore`-function is used in the component.
```
lmiller1990, line 187:
```suggestion
The approach to provide a store for a unit test depends on the way the `useStore` function is used in the component.
```

docs/guide/advanced/vuex.md
```diff
@@ -181,6 +181,48 @@ By creating a `createVuexStore` function that takes an initial state, we can eas
...
+
+### Testing components that utilize `useStore` without an injection keys
+Without an injection key, the store data can just be injected into the component on mount via the provide mounting-option. The name of the injected store must be the same as the one in the component, e.g. "store". Instead of creating the store on the fly, it can also be imported from a file and therefore be shared between unit tests.
```
lmiller1990, line 190:
```suggestion
When using the composition API, the Vuex store is usually accessed via a `useStore`-helper function. In Javascript projects the built-in `useStore` function should be used. More about it here: https://next.vuex.vuejs.org/guide/composition-api.html.
In TypeScript projects the `useStore`-function can be used with an optional and unique injection key to get a properly typed store. To avoid repeating the key parameter passing whenever `useStore` is used, it is recommended to extract that logic into a helper function and reuse that function instead of the default `useStore`-function. More about it here: https://next.vuex.vuejs.org/guide/typescript-support.html#typing-usestore-composition-function.
The approach to provide a store for a unit test depends on the way the `useStore`-function is used in the component.

### Testing components that utilize `useStore` without an injection key

Without an injection key, the store data can just be injected into the component via the global `provide` mounting option. The name of the injected store must be the same as the one in the component, e.g. "store". Instead of creating the store on the fly, it can also be imported from a file and therefore be shared between unit tests.
```

docs/guide/advanced/vuex.md
```diff
@@ -181,6 +181,48 @@ By creating a `createVuexStore` function that takes an initial state, we can eas
...
+        store: createStore({
+          .....
+        }),
+      },
+    },
+})
+```
+
+### Testing components that utilize `useStore` with an injection keys
+When using the store with an injection key the previous approach won't work. The store instance won't be returned from `useStore`. In oder to access the correct store the correct store identifier needs to be provided on mount.
```
lmiller1990, line 206:
```suggestion

When using the store with an injection key ,the previous approach won't work. The store instance won't be returned from `useStore`. In order to access the correct store the identifier needs to be provided.
```

docs/guide/advanced/vuex.md
```diff
@@ -181,6 +181,48 @@ By creating a `createVuexStore` function that takes an initial state, we can eas
...
+It needs to be the exact key that is passed to `useStore` in the `setup`-function of the component or to `useStore` within the custom helper-function. Since JavScript-symbols are unique and can't be recreated, it is best to export the key from the real store.
+As before the mock store can also be imported from a file to improve code reusability.
```
lmiller1990, line 208:
```suggestion
It needs to be the exact key that is passed to `useStore` in the `setup` function of the component or to `useStore` within the custom helper function. Since JavaScript symbols are unique and can't be recreated, it is best to export the key from the real store.

The mock store can also be imported from a file to improve code reusability.
```

docs/guide/advanced/vuex.md
```diff
@@ -181,6 +181,48 @@ By creating a `createVuexStore` function that takes an initial state, we can eas
...
+
+const wrapper = mount(App, {
+  global: {
+      provide: {
+        [key as symbol]: createStore({
+          .....
+        }),
+      },
+    },
+})
```
lmiller1990, line 222:
```suggestion
import { key } from 'store'

const wrapper = mount(App, {
  global: {
      provide: {
        [key as symbol]: createStore({
          // ...
        }),
      },
    },
})
```

docs/guide/advanced/vuex.md
```diff
@@ -181,6 +181,48 @@ By creating a `createVuexStore` function that takes an initial state, we can eas
...
+```
+
```
lmiller1990, line 224: `createStore` is mentioned, but it's not clear where the method comes from. Maybe clarify this to make it clear to the reader?

Thanks for this!

lmiller1990 (commented):
Some comments - mostly small grammatical improvements.

I don't see any real difference between JS and TS - not sure we need to differentiate between the two.

docs/guide/advanced/vuex.md
maurer2: Okay, I removed it.

docs/guide/advanced/vuex.md
maurer2: Okay, I removed it.

docs/guide/advanced/vuex.md
maurer2: Okay, I added it to both code examples.

lmiller1990:
Cool, pushed a few minor changes.

Also added a test showing how to use a symbol as the key and use `global.plugins` instead of `global.provide`: https://github.com/vuejs/vue-test-utils-next/pull/829/files#diff-a19e19208dfc531e739d821349619a56a6f044435c4b0085e90acca2a78378afR44

maurer2:
Okay, nice. Thanks

PR #835, opened 2021-08-06
chore(find): simplify vnode traversal

xanf:
Basically, this is a proper finisher for #188 - we can avoid going over same vnodes again and again if we do not queue them multiple times - as long as we queue `node.component.subTree` - no need to queue`subTree.children` - they will be aggregated later by calling `aggregateChildren(nodes, node.children)`

@lmiller1990 I think you will be happy to see we're finally simplifying this part a bit :)

### DIFF
--- a/src/utils/find.ts
+++ b/src/utils/find.ts
@@ -127,11 +127,8 @@ function findAllVNodes(
   const nodes: VNode[] = [vnode]
   while (nodes.length) {
     const node = nodes.shift()!
-    // match direct children
     aggregateChildren(nodes, node.children)
     if (node.component) {
-      // match children of the wrapping component
-      aggregateChildren(nodes, node.component.subTree.children)
       aggregateChildren(nodes, [node.component.subTree])
     }
     if (node.suspense) {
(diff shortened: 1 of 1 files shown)


src/utils/find.ts
```diff
@@ -127,19 +127,16 @@ function findAllVNodes(
...
-      aggregateChildren(nodes, node.component.subTree.children)
       aggregateChildren(nodes, [node.component.subTree])
     }
     if (node.suspense) {
       // match children if component is Suspense
       const { activeBranch } = node.suspense
       aggregateChildren(nodes, [activeBranch])
     }
-    if (matches(node, selector) && !matchingNodes.includes(node)) {
+    if (matches(node, selector)) {
```
lmiller1990, line 139: Nice

lmiller1990 (approved):
Great! I don't like how this is implemented in general - all the mutation is really confusing, I think we should really be using recursion here. Anyway, this is a good place to start simplifying things - we have lots of tests, so if someone does want to rework it for whatever reason, we can do so with confidence.

PR #836, opened 2021-08-06
fix(find): allow finding root without name

xanf:
Pretty self-describing, I hope :)

### DIFF
--- a/tests/getComponent.spec.ts
+++ b/tests/getComponent.spec.ts
@@ -48,7 +48,7 @@ describe('getComponent', () => {
   })
 
   it('should throw if not found with a component selector that has no name', () => {
-    const wrapper = mount(compA)
+    const wrapper = mount(compB)
     expect(() => wrapper.getComponent(compA)).toThrowError(
       'Unable to get specified component within: <div class="A"></div>'
     )
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -35,7 +35,12 @@ import {
 import { processSlot } from './utils/compileSlots'
 import { createWrapper, VueWrapper } from './vueWrapper'
 import { attachEmitListener } from './emit'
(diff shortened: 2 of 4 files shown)


tests/getComponent.spec.ts
```diff
@@ -48,7 +48,7 @@ describe('getComponent', () => {
   })
 
   it('should throw if not found with a component selector that has no name', () => {
-    const wrapper = mount(compA)
```
lmiller1990, line 51: This test is not relevant anymore, is it? It was testing for the missing behavior which you implemented.

lmiller1990 (commented):
One comment. Also just to confirm my understanding here, the way this works is we create a stub (which is not really a stub, just maps to itself) to be able to search via constructor?

tests/getComponent.spec.ts
xanf: It is still relevant - we still should `throw` getComponent if we were unable to find component without name. It just updated to reflect changes in this PR - previously we were searching for compA inside compA - and it throws (not the case anymore), now we search for compA inside compB and it also throws (which is what this test about - about throwing, not about matching root)

xanf:
> One comment. Also just to confirm my understanding here, the way this works is we create a stub (which is not really a stub, just maps to itself) to be able to search via constructor?

Sometimes (when the component is created via options API or is a functional component) - instead of mounting the component we're providing we're making its "clone" (this is required to modify mixins field for `data()` and `components` for #845). Luckily we already have logic in place "when you're searching for component X - match componentY" which is used to allow finding stubs by original component object. This PR simply reuses this logic for root component (I'm happy that we did not introduce MORE workarounds here) :)

lmiller1990 (approved):
Great - just need to resolve the conflict.

One all these are merged up, let's do a release.

PR #840, opened 2021-08-07
fix(mount): support passing data when immediate watchers are present

xanf:
* Fix issue that data was set too late after watchers with `immediate: true` fired
* Throw when using `data()` option on functional or class component


src/mount.ts
```diff
@@ -328,11 +328,20 @@ export function mount(
...
-    ;(component as any).mixins = [
-      ...((component as any).mixins || []),
-      dataMixin
-    ]
+    const providedData = options.data()
+    if (isObjectComponent(originalComponent)) {
+      // component is guaranteed to be the same type as originalComponent
+      const objectComponent = component as ComponentOptions
+      const originalDataFn = originalComponent.data || (() => ({}))
+      objectComponent.data = (vm) => ({
```
lmiller1990, line 336: I don't understand how this makes immediate watchers work.

lmiller1990 (commented):
Why does calling this function instead of using the mixin make immediate watchers work?

src/mount.ts
xanf: Unfortunately at `beforeCreate` we do not have `this` available and `getCurrentInstance` too.

How it works: we replace `data()` function on component we're actually mounting (since we create a copy of the component in any case - either to change `data()` or to add mixin) with function which calls original `data()`, `data()` provided by `mount` and merges the result

I personally also do not like how intrusive it is and in my perfect world `setData` and `data()` are interfering with component implementation details which is bad, but frankly speaking I see no major difference - either we interfere with `mixins` field or just swap `data()` function

lmiller1990: Ok - glad we both are on the same page about intrusiveness, but that's kind of the nature of something like Test Utils that let's you fine-tune your component for testing.

PR #841, opened 2021-08-07
chore(stubs): simplify global stubs registration

xanf:
Good code is dead code :)

Stubs created inside `mount.ts` for `global.stubs` were never called, since they were replaced by `stubComponents during `transformVNodeArgs` pass (which is intended and correct)

Removed unneeded code and provide explanation why it works in this way
Super happy to have stubs code living in one place instead of two now :)


lmiller1990 (approved):

lmiller1990:
Great catch. RIP dead code, you will be forgotten but you won't be missed.

PR #842, opened 2021-08-08
docs: match findAll example template and test

Thomasan1999:
findAll example test looks for <tr> elements but <span> elements are used inside template


afontcu (approved):
nice catch, thanks!

PR #844, opened 2021-08-09
fix(slots): does not expose extra wrapper of slot content, fix #389

### LINKED ISSUE #389: Slot content doesn't match browser behaviour
In tests if you use `wrapper.vm.$slots.default()` and you have passed multiple html elements in the default slot, there will be only 1 vnode returned in vtu while in the browser, the correct number of vnodes will get returned.
[Reproduction repo](https://github.com/nandi95/vue-test-utils-next/commit/c4309b40654bc48cade5cfb0d835452f8662998c)

xanf:
Exposing extra "SlotWrapper" in certain cases makes VTU behavior different from browser one, as mentioned in #389 

This PR gets rid of "exposing" anything wrapping slot, by providing same function, as if slot was passed in template

~~This PR depends on #815 so it is marked as draft for now :)
Changes specific to this PR could be viewed at d3a147c~~

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -263,27 +263,16 @@ export function mount(
   }
 
   function slotToFunction(slot: Slot) {
-    if (typeof slot === 'object') {
-      if ('render' in slot && slot.render) {
+    switch (typeof slot) {
+      case 'function':
+        return slot
+      case 'object':
         return () => h(slot)
-      }
-
-      if ('template' in slot && slot.template) {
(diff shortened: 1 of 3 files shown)


lmiller1990:
#815 is merged

After all these PRs are merged we should do a release.

xanf:
@lmiller1990 :rocket: ready to go! No more ugly wrappers of components in any case :)

afontcu (approved):

PR #867, opened 2021-08-15
Fix typo in get method description

erikgall:
Fixed a typo in the `get` method description

## Before

Gets **an an** element and returns a `DOMWrapper` if found. Otherwise it throws an error.

## After

Gets **an** element and returns a `DOMWrapper` if found. Otherwise it throws an error.


afontcu (approved):
nice catch, thanks!

Issue #868, opened 2021-08-16
Example code for mocking HTTP requests appears to be broken - out of scope variable for jest.mock()


meanspa:
Ran into trouble trying to follow the documentation's instructions for mocking HTTP requests here - https://next.vue-test-utils.vuejs.org/guide/advanced/http-requests.html

**Steps to reproduce:** 
- Create Vue CLI project with Vue 3, babel, and jest unit testing options
- Create example component and example test from [this docs page](https://next.vue-test-utils.vuejs.org/guide/advanced/http-requests.html)
- Run the test, get this error: `babel-plugin-jest-hoist: The module factory of `jest.mock()` is not allowed to reference any out-of-scope variables.`

**Secondary wording suggestion:**
For me, I found this sentence confusing: "Our first goal is to test this component without actually reaching the API. This would create a fragile and potentially slow test." I wasn't sure what "this" referred to until I researched it later. One alternative that I would find clearer: "Our first goal is to test this component without actually reaching the API,  _because using an external API in a test_ would create a fragile and potentially slow test."

Thanks!

standbyoneself:
The problem here is that when you use `jest.mock()`, it is hoisted by default, so you cannot reference `fakePostList` variable. In order to get this example work you may rewrite implementation of the `axios.get()` for the specific test:

```js
import { mount, flushPromises } from '@vue/test-utils'
import axios from 'axios'
import PostList from '../../src/components/PostList.vue'

const fakePostList = [
  { id: 1, title: 'title1' },
  { id: 2, title: 'title2' }
]

// Following lines tell Jest to mock any call to `axios.get`
// and to return `fakePostList` instead
jest.mock('axios', () => ({
  get: jest.fn()
}))

test('loads posts on button click', async () => {
  jest.spyOn(axios, 'get').mockImplementation(() => fakePostList)

  const wrapper = mount(PostList)

  await wrapper.get('button').trigger('click')

  // Let's assert that we've called axios.get the right amount of times and
  // with the right parameters.
  expect(axios.get).toHaveBeenCalledTimes(1)
  expect(axios.get).toHaveBeenCalledWith('/api/posts')

  // Wait until the DOM updates.
  await flushPromises()

  // Finally, we make sure we've rendered the content from the API.
  const posts = wrapper.findAll('[data-test="post"]')

  expect(posts).toHaveLength(2)
  expect(posts[0].text()).toContain('title1')
  expect(posts[1].text()).toContain('title2')
})
```

[...]

PR #870, opened 2021-08-16
docs: fix test in making http requests

standbyoneself:
This PR fixes the test in the section "Making HTTP Requests" (https://next.vue-test-utils.vuejs.org/guide/advanced/http-requests.html) and makes some explain.

**Why it is necessary?**

Because it has been mentioned in #868.

The previous test is broken because you cannot access variables/methods unprefixed with `mock`. Follow this: https://jestjs.io/docs/es6-class-mocks#calling-jestmock-with-the-module-factory-parameter.

> A limitation with the factory parameter is that, since calls to jest.mock() are hoisted to the top of the file, it's not possible to first define a variable and then use it in the factory. An exception is made for variables that start with the word 'mock'. It's up to you to guarantee that they will be initialized on time! For example, the following will throw an out-of-scope error due to the use of 'fake' instead of 'mock' in the variable declaration.

### DIFF
--- a/docs/guide/advanced/http-requests.md
+++ b/docs/guide/advanced/http-requests.md
@@ -51,15 +51,15 @@ import { mount, flushPromises } from '@vue/test-utils'
 import axios from 'axios'
 import PostList from './PostList.vue'
 
-const fakePostList = [
+const mockPostList = [
   { id: 1, title: 'title1' },
   { id: 2, title: 'title2' }
 ]
 
 // Following lines tell Jest to mock any call to `axios.get`
-// and to return `fakePostList` instead
+// and to return `mockPostList` instead
 jest.mock('axios', () => ({
-  get: jest.fn(() => fakePostList)
+  get: jest.fn(() => mockPostList)
 }))
 
 test('loads posts on button click', async () => {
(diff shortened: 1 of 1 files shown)


docs/guide/advanced/http-requests.md
```diff
@@ -84,7 +84,10 @@ test('loads posts on button click', async () => {
 })
 ```
 
-Notice how we awaited `flushPromises` and then interacted with the Component. We do so to ensure that the DOM has been updated before the assertions run.
+Pay attention that we added prefix `mock` to the variable `mockPostList`. If not, we will get the error: "The module factory of jest.mock() is not allowed to reference any out-of-scope variables."
+Additional info is located here: https://jestjs.io/docs/es6-class-mocks#calling-jestmock-with-the-module-factory-parameter.
```
afontcu, line 88:
```suggestion
Pay attention that we added prefix `mock` to the variable `mockPostList`. If not, we will get the error: "The module factory of jest.mock() is not allowed to reference any out-of-scope variables.". This is jest-specific, and you can read more about this behavior [in their docs](https://jestjs.io/docs/es6-class-mocks#calling-jestmock-with-the-module-factory-parameter).
```

afontcu (approved):
Thanks for this! Nice catch. We might have changed the name at some point without realizing we broke the "magic" Jest performs with `mock-` variables.

meanspa:
Thanks for this quick fix!

Issue #876, opened 2021-08-20
Composition API first Class support


Xenoha:
Hello Team, 

This has been a great suite through the evolution of Vue, However, I feel like with this being a RC, there seems to be a lot of working around happening for simple Vue3 components. The documentation doesn't offer much for testing when using the compositions API and with the release of the setup SFC from the vue team this utils seems even further behind.

I'm not trying to be a bad news bear but I am dealing with a lot of issues which I would think should be first class with a test utils. At the very least, can the docs be updated to show composition api examples for each case?

1. When using the globals.provide with the mount method, the values can be brought into a component as proxies or variable depending on how the user provides them in the parent. This isn't handled well inside the current mount method and when accessing an injected variable with the `myVariable.value` the test utils seems to no longer track correctly.

2. The same occurs when bringing in props. Props are usually defined now through `defineProps`, and the user has the change to destructure with `toRefs` or use the prop itself. As soon as the user uses the property with `toRefs` the user is dealing with a proxy that the suite just doesn't handle.

The below highlights a simple example: (Just to highlight my thoughts above ) 
We have a simple component that renders the provided prop `title`. Then based on `title` and the injected `type` we grab some kind of data and set the ref `data` with the new data array.

```vue
<template>
  <div>
    {{ title }}
  </div>
  <div v-if="data.length" id="data">
    Has Data
  </div>
</template>
```

```vue
<script setup>
import { toRefs, ref, inject } from "vue";

const props = defineProps({
	title: { required: true, type: String },
});

const { title } = toRefs(props);

const type = inject("parentType"); // proxy`

const data = ref([]);

function getData() {
	 // get data based on the title and type.value then set it to data.value.
	 // data.value = newData
}

getData();
</script>
```

I would love your assistance here, but I don't see a way to test the template for the `div#data` element? I don't have a first class way to: `expect(wrapper.find('#data').exists()).toBe(true)`.

To start _I think_  I would have to:

```js
const mockGetData = jest.fn()
global.provide = { parentType: 'something' }
```

Then type.value wouldn't exist because the utils doesn't handle this case, correct? Moreover, checking title.value in the setup won't have a value because I would provide props of `title: 'Some Title'`. 

Using the composition API, how do I toggle the ref `data`? How do I both check the template text and do some calculation of the provided prop `title`? How do I toggle and add data to data.value?

[...]

PR #890, opened 2021-08-27
chore: add script setup tests

lmiller1990:
Added some tests based on the feedback in https://github.com/vuejs/vue-test-utils-next/issues/876


tests/components/ScriptSetup_ToRefsInject.vue
```diff
@@ -0,0 +1,27 @@
...
+    {{ title }}
+  </div>
+  <p>{{ type }}</p>
+  <div v-if="data.length" id="data">Has Data</div>
+</template>
+
+<script setup>
+import { toRefs, ref, inject } from 'vue'
+
+const props = defineProps({
```
cexbrayat, line 12: in TS, this can be written more elegantly using `defineProps<{ title: string}>()`

tests/features/ScriptSetup_ToRefsInject.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+      title: 'Some title'
+    },
+    global: {
+      provide: {
+        parentType: 'Parent Type'
+      }
+    }
+  })
+
+  console.log(wrapper.html())
```
cexbrayat, line 16: this should be removed

tsconfig.json
```diff
@@ -5,6 +5,7 @@
     "module": "esnext",
     "lib": ["DOM", "ES2015", "ES2017"],
     "strict": true,
+    "allowJs": true,
```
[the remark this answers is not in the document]
cexbrayat: I think it would be better to stick to TS components as this ensures vue-tsc can properly check them. And if it works in TS, it works in JS 🙂 So it would be better to remove this, and convert the script setup to TS I think.

cexbrayat (changes requested):
Thanks @lmiller1990 

The lockfile received quite a few changes, maybe this is what caused the failing build?
A few changes are needed, ping when done if you want another review

tests/components/ScriptSetup_ToRefsInject.vue
lmiller1990: Sure. Do you think there's a use case for including some plain JS tests, too? Ideally we should running tests against a wide variety of configs - not sure type safety is a big deal in these kind of fixtures.

lmiller1990:
@cexbrayat thanks for the review, updated the PR.

tests/components/ScriptSetup_ToRefsInject.vue
cexbrayat: It would be nice to have integration tests as a safety net against various config indeed (something like a mini CLI project in TS and in JS, using the built version of VTU). But I think for our unit tests we should be fine if sticking to TS: it's pretty much the hardest thing to get right from our tooling perspective.

cexbrayat:
Awesome, thanks Lachlan!

PR #896, opened 2021-08-28
fix: remove ability to use findComponent with DOM selector

### LINKED ISSUE #689: findAllComponents() returning the same component twice?
Is there a reproduction template for https://codesandbox.io? I'm having trouble finding one.

My component looks something like this:
```
<template>
    <component1 v-for="item in array" :key="item+ '_item'">
        <component2 data-some-attr="attrvalue" />
    </component1>
</template>
<script>
return {
    data() {
        return {
            array: ['foo']
        };
    }
};
</script>
```

Test is trying to retrieve all component instances by the custom attribute:
```
it('test', () => {
    const wrapper = mount(MyComponent);
    const components = wrapper.findAllComponents('[data-some-attr="attrvalue"]');
    expect(components.length).toBe(1);
});
```

It fails on the length assertion, appears to return the same component twice, even though array controlling the loop has 1 item.

If this problem is not familiar to you, I'd like to know if there's an easy reproduction template so I can sort this out.

xanf:
This brings VTU2 behavior in line with VTU1 
Closes #689

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -42,7 +42,7 @@ test('mounts a component', () => {
 })
 ```
 
-Notice that `mount` accepts a second parameter to define the component's state configuration. 
+Notice that `mount` accepts a second parameter to define the component's state configuration.
 
 **Example: mounting with component props and a Vue App plugin**
 ```js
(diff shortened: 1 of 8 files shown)


lmiller1990:
Thanks!

Issue #936, opened 2021-09-10
Build fail due to wrong import in domWrapper.ts


alfeg:
There is seems to be an error that causes our build to fail after library update to v2.0.0-rc.14
https://github.com/vuejs/vue-test-utils-next/blob/df6c234d2511f500be996620118c99625a86e165/src/domWrapper.ts#L8

I guess proper line should be:
```typescript
import { VueWrapper } from '.';
```
As there is no `src` fodler in npm package:
![image](https://user-images.githubusercontent.com/118516/132848036-43db0833-d8dd-4666-9467-3e9669ecf5cf.png)

cexbrayat:
I see you're preparing a PR 👍

You can use `skipLibCheck: true` in the meantime in your projects (it will be faster to compile, and it will skip the libraries compilation, which is usually useless).

alfeg:
@cexbrayat I've prepared PR and forgot to press create pull request =)

Thanks for workaround!

designervoid:
@cexbrayat 
maybe this is good as a temporary solution, but why not use relative imports with dot or two dots (`.`, `..`)
`2.0.0-rc.18` version
![image](https://user-images.githubusercontent.com/43455658/147373211-afe14fbe-f9b1-4afc-ae09-650aae2dd546.png)

designervoid:
with `skipLibCheck: false`:
![image](https://user-images.githubusercontent.com/43455658/147373227-8a2243b7-c48a-459a-b945-840bcfcffafd.png)

designervoid:
with `skipLibCheck: false` and relative imports:
![image](https://user-images.githubusercontent.com/43455658/147373238-0f829139-a4da-4cc3-8a5c-1c07e9e080c5.png)
🥳

cexbrayat:
Sure, if that fixes an issue, feel free to open a PR, and we'll gladly merge it!

designervoid:
@cexbrayat mentioned this PR

catrope:
This is broken again in 2.0.0-rc.19, because #1180 added more broken imports, but #1184 was submitted before #1180 was merged.

```
../../node_modules/@vue/test-utils/dist/interfaces/wrapperLike.d.ts:1:42 - error TS2307: Cannot find module 'src/constants/dom-events' or its corresponding type declarations.

1 import { DomEventNameWithModifier } from 'src/constants/dom-events';
                                           ~~~~~~~~~~~~~~~~~~~~~~~~~~

[...]

PR #937, opened 2021-09-10
fix: domWrapper.d.ts imports

alfeg:
Fix for issue #936 
Created issue, then realized that fix is quite simple.


cexbrayat (approved):
Thanks @alfeg !

cj848:
When will this issue be released? In my project the build is not working because of that problem.

cexbrayat:
A release should occur soon.

In the meantime check the workaround here https://github.com/vuejs/vue-test-utils-next/issues/936#issuecomment-916844985

cj848:
The same option came to my mind and applied.
However, for users who have not applied this option, please release it as soon as possible. Thank you.

PR #971, opened 2021-09-24
chore: bump to vue v3.2.16 and fix devtools

cexbrayat:
There is a breaking change in the devtools API as `setDevtoolsHook` now needs a second parameter.
https://github.com/vuejs/vue-next/commit/2476eaad6e9d68f0b75772456775a0a8165631c0#diff-56930a63772d2e89f19432918b9d6ed851ede4159b0d32404e0c1d8c159fec99R346


src/emit.ts
```diff
@@ -30,7 +30,7 @@ export function emitted<T = unknown>(
 export const attachEmitListener = () => {
   events = {}
   // use devtools to capture this "emit"
-  setDevtoolsHook(createDevTools(events))
+  setDevtoolsHook(createDevTools(events), this)
```
cexbrayat, line 33: I _think_ `this` is the proper value to pass here

cexbrayat:
Merging this one even if not reviewed as it unblocks dependabot

src/emit.ts
cexbrayat: Ho right, sorry I missed that. Opened a PR to fix it https://github.com/vuejs/vue-test-utils-next/pull/1011

PR #994, opened 2021-10-03
fix: return abitility to use findComponent with DOM selector

### LINKED ISSUE #904: Request to reinstate ability to use CSS selectors in findComponent
It said in the release notes that if we felt strongly about this feature we could open an issue, so that's what I'm doing. I don't think it makes sense to remove the ability to use CSS selectors in findComponent just to make VTU2 more in line with VTU1, and I think there are some notable benefits to using CSS selectors to find components. 

The most significant benefit (in my opinion) is the ability to more specifically select components. Say I have a component like so, that imports a Spinner component and then shows it on two buttons depending on which is clicked
```vue
<template>
  <button @click="buttonASpinning = true">A <Spinner id="btnASpinner" v-if="buttonASpinning" /></button>
  <button @click="buttonBSpinning = true">B <Spinner id="btnBSpinner" v-if="buttonBSpinning" /></button>
</template>

<script>
import { defineComponent, ref } from 'vue'
import Spinner from '@/components/Spinner.vue'

  components: { Spinner },
  setup() {
    const buttonASpinning = ref(false)
    const buttonBSpinning = ref(false)
    return { buttonASpinning, buttonBSpinning }
  },
})
</script>
```
Without CSS selectors I have to use `findAllComponents(Spinner)[1]` to access Spinner B, which will break if the buttons switch positions, if there are any other spinners in any other components before this component on the page, etc. With large components or views this can become a nightmare of remembering state for completely unrelated parts of the page. 

It was also just nice to have another option for selecting components that didn't come with the downsides of the other three. Selecting by constructor requires you to import the constructor into your test, selecting by name requires you to know the name of the component which isn't always readily available or even present for 3rd party components, and selecting by ref requires you to add refs to components that don't already have them just for testing. Personally I like to use data attributes for selectors in tests since they can be easily removed in production.

xanf:
This PR consists of two commits:

* one is just a revert of #896 
* second one makes `find` (which is used underneath of `findComponent` / `findAllComponents`) to return only one vnode for same element when called with CSS selector (fixing #689). This behavior is in line with VTU v1 behavior (see #904 for more context)

Closes #904

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -42,7 +42,7 @@ test('mounts a component', () => {
 })
 ```
 
-Notice that `mount` accepts a second parameter to define the component's state configuration. 
+Notice that `mount` accepts a second parameter to define the component's state configuration.
 
 **Example: mounting with component props and a Vue App plugin**
 ```js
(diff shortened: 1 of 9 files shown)


lmiller1990 (approved):
Nice, I like how little code you had to change to implement this.

Can release on the weekend.

Issue #806, opened 2021-07-29
Script Setup and Stubs


robokozo:
Stubs are not working when using the latest script setup syntax.

Versions:
`"vue": "^3.1.5"`
`"@vue/test-utils": "^2.0.0-rc.12",`
(Bootstrapped from the vue-cli)

This seems to only matter for the component under test. The stubbed component can be written in either style.

HelloWorld.vue (Default Style)
```vue
<template>
  <div class="hello">
    <h1>{{ msg }}</h1>
    <NamedSlot>
      <template #inside="{ person }">
        <div>
          {{ person.firstName }}
          {{ person.lastName }}
        </div>
      </template>
    </NamedSlot>
  </div>
</template>

<script lang="ts">
import { defineComponent } from "vue";
import NamedSlot from "./NamedSlot.vue";

export default defineComponent({
  props: {
    msg: String,
  },
  components: {
    NamedSlot,
  },
});
</script>
```

HelloWorld.vue (Script Setup)
```vue
<template>
  <div class="hello">
    <h1>{{ msg }}</h1>
    <NamedSlot>
      <template #inside="{ person }">
        <div>
          {{ person.firstName }}
          {{ person.lastName }}
        </div>
      </template>
    </NamedSlot>
  </div>
</template>

<script lang="ts" setup>
import NamedSlot from "./NamedSlot.vue";

defineProps(["msg"]);
</script>
```

example.spect.ts
```
describe("HelloWorld.vue", () => {
  it("renders props.msg when passed", () => {
    const msg = "new message";
    const wrapper = mount(HelloWorld, {
      props: {
        msg,
      },
      global: {
        stubs: {
          NamedSlot: {
            template: `
              <div class="stubbed">
                <slot name="inside" :person="{ firstName: 'A', lastName: 'B' }"></slot>
              </div>
            `,
          },
        },
      },
    });
    expect(wrapper.html()).toMatchSnapshot();
  });
});
```

example.spec.ts.snap (Default Style)
```
exports[`HelloWorld.vue renders props.msg when passed 1`] = `
<div class="hello">
  <h1>new message</h1>
  <div class="stubbed">
    <div>A B</div>
  </div>
</div>
`;
```

example.spec.ts.snap (Script Setup)
```
exports[`HelloWorld.vue renders props.msg when passed 1`] = `
<div class="hello">
  <h1>new message</h1>
  <div>
    <div>DONT SHOW ME IN TESTS</div>
[...]

PR #1066, opened 2021-11-12
fix: [BREAKING] Use default name for components with setup script and nameless compon…

khalilst:
This PR in the following of issue #806 replaces the anonymous naming with the default component name:

So the following component's stub name will be `script-setup-stub`:
```js
//filename: ScriptSetup.vue
<script setup>
...
</script>
```

Also, the same approach works on nameless components.
The test coverage applied for this fix.

### DIFF
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -109,9 +109,14 @@ const getComponentRegisteredName = (
   return null
 }
 
-const getComponentName = (type: VNodeTypes): string => {
+const getComponentName = (
+  instance: any | null,
+  type: VNodeTypes
+): string => {
   if (isObjectComponent(type)) {
-    return type.name || ''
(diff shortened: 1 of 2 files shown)


cexbrayat (approved):
Thanks @khalilst for the PR!

LGTM, but I'd like another pair of eyes before merging this.

cexbrayat:
@lmiller1990 @xanf @afontcu I added you as reviewers, as I don't know if that's too breaking or OK to merge. What do you think?

khalilst:
@cexbrayat 
My pleasure (^_^)
Sure, you need to check more. Just for your consideration, I've applied it in a local test env and everything works fine.

lmiller1990 (commented):
I think this makes sense. Bit of a breaking change for snapshots, but this seems like a overall improvement.

PR #1078, opened 2021-11-16
fix(findComponent): favor displayName over name when searching

xanf:
This fix swaps priority order when searching for component by name to make `displayName` respected.
It is important for functional components: 
```js
const Cmp = () => 'hello'
// hey, I gave this one a displayName for A REASON
Cmp.displayName = 'FooComponent';
```
but is even more important for legacy functional components when used with `@vue/compat` - every single functional component will be named `Func` without this fix :)

### DIFF
--- a/src/utils/find.ts
+++ b/src/utils/find.ts
@@ -45,14 +45,16 @@ export function matches(
   }
 
   let componentName: string | undefined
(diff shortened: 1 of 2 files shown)


lmiller1990 (approved):
Good catch 🎉

PR #1079, opened 2021-11-16
fix(mount): pass unknown options to $options on mount

xanf:
Vue 3.x keeps the behaviour we've had in Vue2: all unknown fields in component are passed as $options. This allows you to do following:
```
createApp({ ...App, test: 123 }).mount("#app");
```
and access test on `this.$options.test`
While not in a great use in modern Vue 3 code, this is still a great importance when we work with `@vue/compat` - you are still want to be able to pass router/store/etc. in "legacy" way to your component

This PR restores this behavior


afontcu (approved):
nice catch!

PR #1086, opened 2021-11-18
fix: remove symbol props from stubs

### LINKED ISSUE #1076: using shallowMount on a component that uses Symbol causes tests to crash
When using `shallowMount` on a component that uses `Symbol` as a value for a property, Vue Test Utils fail to run. I receive the following error: 

```
 TypeError: Failed to execute 'setAttribute' on 'Element': parameter 2 is a symbol, which cannot be converted to a string.

      at Object.exports.DOMString (node_modules/webidl-conversions/lib/index.js:228:15)
      at patchAttr (node_modules/@vue/runtime-dom/dist/runtime-dom.cjs.js:207:16)
      at patchProp (node_modules/@vue/runtime-dom/dist/runtime-dom.cjs.js:406:9)
      at mountElement (node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:3751:25)
      at processElement (node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:3724:13)
      at patch (node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:3644:21)
      at ReactiveEffect.componentUpdateFn [as fn] (node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:4181:21)
      at ReactiveEffect.run (node_modules/@vue/reactivity/dist/reactivity.cjs.js:164:29)
      at setupRenderEffect (node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:4300:9)
```

I would expect there to be no error when using `shallowMount`. When using `mount`, the tests run as expected.

------

Steps to reproduce:

1. Clone https://github.com/liamdebeasi/vue-test-utils-symbol.
2. Run `npm install`.
3. Run `npm run test:unit`. Observe that the tests fail on the error above.
4. Open `src/MyCmp.vue` and change the default value for `myProp` from `Symbol()` to `'123'`.
5. Run `npm run test:unit`. Observe that the tests pass as expected.

This is a blocker for some Vue users upgrading to Ionic 6, so it would be great if this could get fixed: https://github.com/ionic-team/ionic-framework/issues/24181. We make use of Symbols to account for this change in behavior with Vue 3.1: https://github.com/vuejs/vue-next/issues/3889

lmiller1990:
resolves #1076

More context: https://github.com/ionic-team/ionic-framework/issues/24181

Basically when you do a stub that receives a `Symbol` as a prop, it ends up doing `<blah-stub sym="Symbol()">` which isn't allowed. Try it in your browser: `document.body.setAttribute('sym', Symbol())`.

I just filter out all the symbol props. Bit of a hack, but not a breaking change (passing a symbol as prop to a stub causes an error in both VTU v1 and v2). 

Happy to solicit any better work arounds for this issue, best I could come up with for now to unblock the issue in Ionic team repo.

### DIFF
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -22,7 +22,7 @@ import { Stub, Stubs } from './types'
 
 interface StubOptions {
   name: string
-  propsDeclaration: ComponentPropsOptions
+  propsDeclaration?: ComponentPropsOptions
   renderStubDefaultSlot?: boolean
 }
 
@@ -43,7 +43,7 @@ const shouldNotStub = (type: ConcreteComponent) => doNotStubComponents.has(type)
 export const addToDoNotStubComponents = (type: ConcreteComponent) =>
   doNotStubComponents.add(type)
 
-const removeSymbols = <T = ComponentPropsOptions>(props: T): T => {
+const stringifySymbols = (props: ComponentPropsOptions) => {
   // props are always normalized to object syntax
   const $props = props as unknown as ComponentObjectPropsOptions
   return Object.keys($props).reduce((acc, key) => {
@@ -51,7 +51,7 @@ const removeSymbols = <T = ComponentPropsOptions>(props: T): T => {
       return { ...acc, [key]: $props[key]?.toString() }
     }
     return { ...acc, [key]: $props[key] }
-  }, {}) as T
+  }, {})
 }
 
 export const createStub = ({
@@ -68,8 +68,8 @@ export const createStub = ({
     // something like `el.setAttribute('val', Symbol())` which is not valid and
     // causes an error.
     // Only a problem when shallow mounting. For this reason we iterate of the
-    // props that will be passed and remove any that are symbols.
-    const propsWithoutSymbols = removeSymbols(ctx.$props)
+    // props that will be passed and stringify any that are symbols.
+    const propsWithoutSymbols = stringifySymbols(ctx.$props)
 
     return h(
       tag,
@@ -189,7 +189,6 @@ export function stubComponents(
       return [
         createTransitionStub({
           name: 'transition-stub',
-          propsDeclaration: {}
         }),
         undefined,
         children
@@ -205,7 +204,6 @@ export function stubComponents(
       return [
         createTransitionStub({
           name: 'transition-group-stub',
-          propsDeclaration: {}
         }),
         undefined,
         children
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -7,7 +7,8 @@ import {
   defineComponent,
   VNodeTypes,
   ConcreteComponent,
-  ComponentPropsOptions
+  ComponentPropsOptions,
+  ComponentObjectPropsOptions
 } from 'vue'
 import { hyphenate } from './utils/vueShared'
 import { matchName } from './utils/matchName'
@@ -42,6 +43,17 @@ const shouldNotStub = (type: ConcreteComponent) => doNotStubComponents.has(type)
 export const addToDoNotStubComponents = (type: ConcreteComponent) =>
   doNotStubComponents.add(type)
 
+const removeSymbols = <T = ComponentPropsOptions>(props: T): T  => {
+  // props are always normalized to object syntax
+  const $props = props as unknown as ComponentObjectPropsOptions
+  return Object.keys($props).reduce((acc, key) => {
+    if (typeof $props[key] === 'symbol') {
+      return acc
+    }
+    return {...acc, [key]: $props[key]}
+  }, {}) as T
+}
+
 export const createStub = ({
   name,
   propsDeclaration,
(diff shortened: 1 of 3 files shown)


cexbrayat (commented):
This makes me think that the fix should not be in VTU but in Vue itself (as it won't be in webidl-conversion).
The same will happen in Vue if someone binds a Symbol to an attribute right?
So instead of having this dirty hack, I thhink the issue should be opened upstream on vue-next, and fix it there.

If we ever end up merging this hack in VTU, maybe we should consider not removing the Symbol, but replacing its value with a string equivalent, so the stub will properly reflect that there was a Symbol.

cexbrayat:
I opened a PR to fix this in vue-next instead https://github.com/vuejs/vue-next/pull/4964

If that gets merged we won't need this PR, and a simple upgrade to the latest vue-next will solve the issue.
Otherwise, we'll go with Lachlan's hack 😉

lmiller1990:
Nice, I added a comment to your PR https://github.com/vuejs/vue-next/pull/4964

I like your fix, although Vue core has so many open PRs - if it won't be merged for a while, we could use this until that PR is merged, then change to that implementation.

We should ping Evan for a review on that PR, as well as the `jest.spyOn` fix.

src/stubs.ts
```diff
@@ -42,6 +43,17 @@ const shouldNotStub = (type: ConcreteComponent) => doNotStubComponents.has(type)
 export const addToDoNotStubComponents = (type: ConcreteComponent) =>
   doNotStubComponents.add(type)
 
+const removeSymbols = <T = ComponentPropsOptions>(props: T): T => {
+  // props are always normalized to object syntax
+  const $props = props as unknown as ComponentObjectPropsOptions
+  return Object.keys($props).reduce((acc, key) => {
+    if (typeof $props[key] === 'symbol') {
+      return acc
```
cexbrayat, line 51: WDYT of stringifying the Symbol instead of completely removing it? It would then show in the snapshot. (a simple `$props[key].toString()` should do the trick)

tests/props.spec.ts
```diff
@@ -224,4 +225,55 @@ describe('props', () => {
...
+          }
+        }
+      })
+
+      const wrapper = shallowMount(Comp)
+
+      expect(wrapper.html()).toBe('<div>Symbol: Symbol()</div>')
+    })
+
+    it('works with Symbol an array syntax', () => {
```
cexbrayat, line 246: an -> as

src/stubs.ts
```diff
@@ -62,7 +86,7 @@ export const createStub = ({
   })
 }
 
-const createTransitionStub = ({ name }: StubOptions) => {
+const createTransitionStub = ({ name, propsDeclaration }: StubOptions) => {
```
cexbrayat, line 89: Are the changes to createTransitionStub necessary? I fail to see the link. If it is necessary maybe add a test case with a transition?

cexbrayat (changes requested):
As the PR in vue-next was not merged, let's go with the hacky patch in VTU 😉

src/stubs.ts
afontcu: +1

given that this is only an issue when shallow mounting, I guess stringifying makes sense as people use shallow mounting to do things such as snapshotting or checking against a prop value. I never use it, but I guess stringifying the symbol aligns with the same mental model of "exploring" the rendered stuff

src/stubs.ts
lmiller1990: reverted

src/stubs.ts
lmiller1990: Done!

lmiller1990:
@cexbrayat @afontcu updated 👍

src/stubs.ts
```diff
@@ -42,6 +43,17 @@ const shouldNotStub = (type: ConcreteComponent) => doNotStubComponents.has(type)
 export const addToDoNotStubComponents = (type: ConcreteComponent) =>
   doNotStubComponents.add(type)
 
+const removeSymbols = <T = ComponentPropsOptions>(props: T): T => {
```
cexbrayat, line 46: nit: this could be renamed to `stringifySymbols` as it is not removing them anymore?

src/stubs.ts
```diff
@@ -51,7 +63,19 @@ export const createStub = ({
   const tag = name ? `${hyphenate(name)}-stub` : anonName
 
   const render = (ctx: ComponentPublicInstance) => {
-    return h(tag, ctx.$props, renderStubDefaultSlot ? ctx.$slots : undefined)
+    // https://github.com/vuejs/vue-test-utils-next/issues/1076
+    // Passing a symbol as a static prop is not legal, since Vue will try to do
+    // something like `el.setAttribute('val', Symbol())` which is not valid and
+    // causes an error.
+    // Only a problem when shallow mounting. For this reason we iterate of the
+    // props that will be passed and remove any that are symbols.
```
cexbrayat, line 71: description is not accurate anymore: this should mention that they are stringified

src/stubs.ts
```diff
@@ -51,7 +63,19 @@ export const createStub = ({
...
+    const propsWithoutSymbols = removeSymbols(ctx.$props)
```
cexbrayat, line 72: nit: same here, the variable could be renamed

src/stubs.ts
```diff
@@ -164,7 +188,8 @@ export function stubComponents(
     if (type === Transition && 'transition' in stubs && stubs['transition']) {
       return [
         createTransitionStub({
-          name: 'transition-stub'
+          name: 'transition-stub',
+          propsDeclaration: {}
```
cexbrayat, line 192: If the option is reverted to optional, this can be removed.

src/stubs.ts
```diff
@@ -179,7 +204,8 @@ export function stubComponents(
     ) {
       return [
         createTransitionStub({
-          name: 'transition-group-stub'
+          name: 'transition-group-stub',
+          propsDeclaration: {}
```
cexbrayat, line 208: same here

src/stubs.ts
cexbrayat: ooc: is the generic type `T` necessary here? As it is only called once, can't we use `ComponentPropsOptions` directly?

src/stubs.ts
```diff
@@ -21,7 +22,7 @@ import { Stub, Stubs } from './types'
 
 interface StubOptions {
   name: string
-  propsDeclaration?: ComponentPropsOptions
+  propsDeclaration: ComponentPropsOptions
```
cexbrayat, line 25: is this necessary? Can't that be left optional, and have a default value in the stubs?

lmiller1990: Done, fixed all comments, thanks for the review!

src/stubs.ts
lmiller1990: Removed that

afontcu (approved):

cexbrayat (approved):

Issue #1070, opened 2021-11-13
Using teleport with shallowMount


doutatsu:
I am getting a Vue warning for my teleport, when doing a shallowMount. I am unable to find a way to get a workaround on this - tried stubbing it out, but that also didn't do anything...

### Warning
```js
console.warn
    [Vue warn]: Non-function value encountered for default slot. Prefer function slots for better performance.
      at <AnonymousStub to="nav" >
      at <Anonymous ref="VTU_COMPONENT" >
      at <VTUROOT>
```

### Component template
```pug
<template lang="pug">
teleport(to='nav')
  .fixed.bottom-5.right-5(ref='fab')
    span.relative.z-50
      button.fab.ripple(
        @click='open = !open'
        type='button'
        aria-haspopup='true'
        aria-expanded='false'
      )
        .absolute.transform.transition.duration-150(
          :class="open ? 'rotate-90 opacity-0' : 'rotate-0 opacity-100'"
        )
          slot
        icon-x-outline.absolute.w-6.h-6.inline-block.text-white.transform.transition.duration-150(
          ref='toggleIcon'
          style="padding: 2px 0 0 2px"
          :class="open ? 'rotate-0 opacity-100' : '-rotate-90 opacity-0'"
        )
    transition(name="overlay-fade")
      button.click-off-button(
        ref="clickOffButton"
        v-show='open'
        @click='open = false'
        tabindex='-1'
      )
        .absolute.inset-0.bg-white.opacity-95.dark_bg-gray-900.dark_opacity-90
    template(v-if="options.length")
      transition(name="dropdown")
        .dropdown-body.w-full.sm_w-40(v-show='open')
          .py-1
            a.group.justify-end.space-x-4(
              v-for="option in options"
              href='#'
              @click.prevent="handleClick(option.action)"
            )
              span(v-if="option.text" v-text="option.text")
              .secondary-fab.ripple
                component.h-5.w-5.text-gray-400.dark_text-white(:is="option.icon")
</template>
```

### Test
```js
import BaseFab from '@/components/base_components/BaseFab.vue';

describe('BaseFab.vue', () => {
  let baseFab;

  describe('when open', () => {
    beforeEach(() => {
      baseFab = shallowMount(BaseFab, {
        slots: { default: '<div>My slot content</div>' },
        data() { return { open: true }; },
      });
    });

    it('shows cross icon', () => {
      expect(baseFab.findComponent({ ref: 'toggleIcon' }).element)
[...]

PR #1087, opened 2021-11-18
feat: add ability to stub teleport

TheDutchCoder:
Adds stubbing support for the `teleport` Vue component

### Notes
Jest throws a warning:
```
  console.warn
    [Vue warn]: Non-function value encountered for default slot. Prefer function slots for better performance.
      at <TeleportStub>
      at <Anonymous ref="VTU_COMPONENT" >
      at <VTUROOT>
```

I believe this has something to do with how teleport consumes slots and how the renderer provides them, but I couldn't figure out exactly what the issue was. Providing them as an object with a function didn't solve it (i.e. `{ default: getSlots }`.

Teleport seems a bit special in Vue land, I think this is okay, but please make sure it is ;)

### DIFF
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -171,7 +171,7 @@ export function stubComponents(
 
   transformVNodeArgs((args, instance: ComponentInternalInstance | null) => {
     const [nodeType, props, children, patchFlag, dynamicProps] = args
-    const type = nodeType as (VNodeTypes | typeof Teleport)
+    const type = nodeType as VNodeTypes | typeof Teleport
 
     // stub transition by default via config.global.stubs
     if (type === Transition && 'transition' in stubs && stubs['transition']) {
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -206,7 +206,7 @@ export function stubComponents(
           name: 'teleport-stub'
         }),
         undefined,
-        children
+        () => children
       ]
     }
 
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -2,6 +2,7 @@ import {
   transformVNodeArgs,
   Transition,
   TransitionGroup,
+  Teleport,
   h,
   ComponentPublicInstance,
   defineComponent,
@@ -9,7 +10,6 @@ import {
   ConcreteComponent,
   ComponentPropsOptions
 } from 'vue'
-import type { Teleport } from 'vue'
 import { hyphenate } from './utils/vueShared'
 import { matchName } from './utils/matchName'
 import { isComponent, isFunctionalComponent, isObjectComponent } from './utils'
@@ -200,7 +200,7 @@ export function stubComponents(
     }
 
     // stub teleport by default via config.global.stubs
-    if ((type as typeof Teleport).__isTeleport && 'teleport' in stubs && stubs['teleport']) {
+    if (type === Teleport && 'teleport' in stubs && stubs['teleport']) {
       return [
         createTeleportStub({
           name: 'teleport-stub'
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -9,6 +9,7 @@ import {
   ConcreteComponent,
   ComponentPropsOptions
 } from 'vue'
+import type { Teleport } from 'vue'
 import { hyphenate } from './utils/vueShared'
 import { matchName } from './utils/matchName'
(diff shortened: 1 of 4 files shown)


tests/mountingOptions/global.stubs.spec.ts
```diff
@@ -419,6 +419,35 @@ describe('mounting options: stubs', () => {
...
+    }
+    const wrapper = mount(Comp, {
+      global: {
+        stubs: {
+          teleport: false
+        }
+      }
+    })
+
+    expect(wrapper.html()).toContain('<!--teleport start-->')
```
lmiller1990, line 447: I'm guessing we also want `not.toContain('<div id="content></div>')` as well.

src/stubs.ts
```diff
@@ -186,6 +199,17 @@ export function stubComponents(
       ]
     }
 
+    // stub teleport by default via config.global.stubs
+    if ((type as typeof Teleport).__isTeleport && 'teleport' in stubs && stubs['teleport']) {
```
lmiller1990, line 203: Can't we just do `type === Teleport`? Same as we do with `Transition`.

src/stubs.ts
```diff
@@ -9,6 +9,7 @@ import {
   ConcreteComponent,
   ComponentPropsOptions
 } from 'vue'
+import type { Teleport } from 'vue'
```
lmiller1990, line 12: Instead of grabbing the type, grab the actual component then we can do the comparison (as commented below). Eg `import { Teleport } from 'vue'`.

lmiller1990 (commented):
I am not sure we want to auto stub teleport - can we default to *not* stubbing it?

We can always change this later, but since it would be breaking change, I'd like to get more feedback. If we make it opt in, (no stub by default) we can get this merged up sooner, and talk more about breaking changes prior to 2.0.0.

src/stubs.ts
TheDutchCoder: At first that didn't work, because Teleport was not included in the `const type = nodeType as VNodeTypes` declaration.

It's kind of odd, because `VNodeTypes` literally has `| typeof Teleport` in the type, but now that I explicitly added `const type = nodeType as (VNodeTypes | typeof Teleport)` it does seem to work, let me try that again.

src/stubs.ts
TheDutchCoder: See my previous comment. I got this working so I'll remove it!

TheDutchCoder:
> I am not sure we want to auto stub teleport - can we default to _not_ stubbing it?
> 
> We can always change this later, but since it would be breaking change, I'd like to get more feedback. If we make it opt in, (no stub by default) we can get this merged up sooner, and talk more about breaking changes prior to 2.0.0.

Good point, I'll reverse it instead!

TheDutchCoder:
As a side note. The master branch has 5 failing tests currently, just so you know, do you want me to open an issue for that?

tests/mountingOptions/global.stubs.spec.ts
TheDutchCoder: I've made a stricter check on the DOM to counter this.

lmiller1990 (commented):
Code seems fine, last thing would be adding a quick paragraph at the top of this file: https://github.com/vuejs/vue-test-utils-next/blob/master/docs/guide/advanced/teleport.md. You could say something like "If you just want to test the rest of the component, ignoring teleport, you can pass `teleport: false`.

src/stubs.ts
```diff
@@ -186,6 +199,17 @@ export function stubComponents(
...
     }
 
+    // stub teleport by default via config.global.stubs
+    if (type === Teleport && 'teleport' in stubs && stubs['teleport']) {
+      return [
+        createTeleportStub({
+          name: 'teleport-stub'
+        }),
+        undefined,
+        children
```
lmiller1990, line 209:
```ts
 if (type === Teleport && 'teleport' in stubs && stubs['teleport']) {
      return [
        createTeleportStub({
          name: 'teleport-stub'
        }),
        undefined,
        () => children // <====== added function
      ]
    }
```

afontcu (approved):

lmiller1990 (approved):

doutatsu:
This is great - addressed https://github.com/vuejs/vue-test-utils-next/issues/1070 for me

Issue #899, opened 2021-08-30
Improve typing of findAllComponents


xanf:
See https://github.com/vuejs/vue-test-utils-next/pull/897#discussion_r697885293 for original discussion

`findAllComponents` is incorrectly typed by VueWrapper<T>[], where `T` - is a type of original wrapper, not the selector. We should explore possibility of improving this typing to provide better error-free experience for TypeScript users

xanf:
:white_check_mark: closed via #1110

PR #1094, opened 2021-11-22
chore(findComponent): refactor find & findComponent (fixes #716, #899)

### LINKED ISSUE #716: Support `find({ ref })` syntax

xanf:
Sorry for such a big one, I was working on typings and improving typings revealed multiple issues

Basically major changes in this PR are:
  * `find` now finds DOM nodes by ref (#716)
  * `findAllComponents` / `findComponent` / `getComponent` migrated to `baseWrapper` since they are now officially supported in both wrappers 
  * major types rework for `findComponent` / `findAllComponents` with better type inference by large margin
  * better typing for `.element` - it was originally `Element` which is actually wrong - it could be Element, Comment (when rendering empty v-if), Text (if component renders just raw text). Fixed this typings and added guards everywhere
  * `findComponent` and `findAllComponents` are now silently returning DOMWrappers for functional components encountered. This mimics VTU v1 behavior and really helps with migration
  * add type tests for findComponent


src/types.ts
```diff
@@ -5,27 +5,24 @@ import {
...
-}
-
-interface RefSelector {
-  ref: string
-}
-
-interface NameSelector {
+export interface NameSelector {
   name: string
+  length?: never
```
xanf, line 17: This was a nasty one. `{ name: string }` definition matches Function (because functions have `name` property) which messes with our type inference. Putting `length` to `never` allows us to distinguish between finding a function and finding by object with name field

tests/findAllComponents.spec.ts
```diff
@@ -66,16 +66,17 @@ describe('findAllComponents', () => {
     const wrapper = mount(RootComponent)
 
     expect(wrapper.findAllComponents('.in-root')).toHaveLength(1)
-    expect(wrapper.findAllComponents('.in-root')[0].vm.$options.name).toEqual(
-      'NestedChild'
-    )
+    expect(
+      wrapper.findAllComponents<DefineComponent>('.in-root')[0].vm.$options.name
```
xanf, line 70: I like that when you're doing something **nasty** (working outside `WrapperLike` interface which is unified across `VueWrapper` and `DOMWrapper` - you need to explicitly specify "hey, I know that I'm doing, there will be no functional component here". Type safety as it is!

src/vueWrapper.ts
```diff
@@ -1,35 +1,37 @@
...
-  implements WrapperLike
-{
+export class VueWrapper<
+  T extends Omit<
+    ComponentPublicInstance,
+    '$emit' | keyof ComponentCustomProperties
+  > & {
+    $emit: (event: any, ...args: any[]) => void
+  } & ComponentCustomProperties = ComponentPublicInstance
+> extends BaseWrapper<Node> {
```
xanf, line 26: TypeScript as usually sucks in co- and contravariance, that's why we need to relax check on `$emit` a bit (in other words `$emit` generated by `InstanceType<DefineComponent<...>>` which is `ComponentPublicInstance<...>` is not a subclass of `ComponentPublicInstance` (with default params) due to params type mismatch in `EmitFn` (the code of this helper type in vue-next is a bit insane)

But I wish someone told me why i need to decompose & put back `ComponentCustomProperties` - or `mount` will fail typecheck otherwise :confused:

xanf:
@cexbrayat will be great to have another pair of eyes, especially on TypeScript typings :)

@lmiller1990 it seems that I've occasionally implemented circuit-breaker similar to your one at 2700e88351fe9853288f974b9d0469f6084380d7, so this commit kinda reverts that change, sorry (I think separate `createWrapper` approach is a bit cleaner)

src/types.ts
afontcu: damn, nasty but smart. thanks for the explanation, I would've missed it

src/vueWrapper.ts
```diff
@@ -1,35 +1,37 @@
...
-export class VueWrapper<T extends ComponentPublicInstance>
-  extends BaseWrapper<T['$el']>
-  implements WrapperLike
-{
+export class VueWrapper<
+  T extends Omit<
+    ComponentPublicInstance,
+    '$emit' | keyof ComponentCustomProperties
+  > & {
+    $emit: (event: any, ...args: any[]) => void
```
lmiller1990, line 24: Should this be `event: string`?

lmiller1990 (commented):
Looks good - I don't fully see how the circular dep is cleaned up, but assuming if you do `yarn build` there is no warnings?

Left one minor comments. Could probably add more tsd tests if you like (or at a future time). Excited for this PR!

src/vueWrapper.ts
xanf: Unfortunately no. `ComponentPublicInstance<>` (with all default parameters) infers $emit exactly as `(event: string, ...args: any[]) => void`. When you pass here a component with defined emits (for example `hi`) $emit will be `(event: "hi") => void'` which is **not** a subtype of `(event: string, ...args: any[]) => void` (event argument is too narrow). That's why we need to cheat with `any`

I've tried to add `event: EmitType` and `EmitType extends string = string` as parameter to VueWrapper generic to express this limitiation , but it seems typescript still infers wrong type

cexbrayat (approved):
Nice work! To be honest, it's hard to review, but LGTM overall.

@pikax maybe you can take a look for the types part?

@lmiller1990 As this might introduce unwanted regressions/behavior differences, maybe we could release it on a special tag, before making it rc.18?

xanf:
> Nice work! To be honest, it's hard to review, but LGTM overall.

Sorry for that, I promise not to do such big one in future :hugs:. It was unexpectedly too interconnected
 
> @lmiller1990 As this might introduce unwanted regressions/behavior differences, maybe we could release it on a special tag, before making it rc.18?

I would like **not** to do that. We have relatively small user base, so adding a special tag will not help that much, making sure it is even less tested. I've ran this change accross big suite (2k+ tests) with no changes, so I'm pretty confident in it except types part :)

> Looks good - I don't fully see how the circular dep is cleaned up, but assuming if you do yarn build there is no warnings?

Sorry. I totally overlooked this one. I pushed separate commit e5e42dc to fix warning, WDYT of that? 
This pattern (factory with runtime registration) feels a bit cleaner for me

PR #1113, opened 2021-11-29
fix(find): allow finding stubs by stub definition

xanf:
Long-long ago, in the galaxy far-far away it started pretty simple: we've passed a stub for a component and VTU rendered this stub via `transformVNodeArgs` provided by Vue core:

Untitled-2021-11-29-1408

We were able to find stub by stub (obviously, because VNode matched) and component by stub because [we've tracked relation between component and it's stub](https://github.com/vuejs/vue-test-utils-next/pull/696) (blue line)

Unfortunately, soon we've discovered that sometimes people use same stub for stubbing multiple components:

Untitled-2021-11-29-1408-1

This was a problem, because we were unable to tell, if `Stub 1` was the stub of `Original component 1` or `Original component 2`

So, we've [introduced](https://github.com/vuejs/vue-test-utils-next/pull/705) a "specialized stub" concept - it is simply a copy of original stub, so for each original component stub is "unique":

Untitled-2021-11-29-1408-2

We were still tracking relation between specialized stubs and original components (red dotted line) to keep everything consistent.

Unfortunately, this broke use case, when people are searching stubs by passing stub as component - if stub has no name, it will not match anything (and it worked in VTU v1). So this PR is pretty simple: it adds tracking relations between specialized stubs and stubs too, so we could match specialized stubs when searching for stubs (green dotted line):

Untitled-2021-11-29-1408-3

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -277,7 +277,7 @@ export function mount(
   }
 
   addToDoNotStubComponents(component)
-  registerStub(originalComponent, component)
+  registerStub({ source: originalComponent, stub: component })
   const el = document.createElement('div')
 
   if (options?.attachTo) {
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -26,17 +26,29 @@ interface StubOptions {
   renderStubDefaultSlot?: boolean
 }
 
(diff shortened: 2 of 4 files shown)


cexbrayat (approved):
LGTM, and this PR probably wins the best PR description of the year 😉

PR #1136, opened 2021-12-07
Fix: Use defaultName when it exists to correctly stub async components

Evertvdw:
When mounting a component that has components defined with `defineAsyncComponent` the result of `type.name` is `AsyncComponentWrapper`. The `defaultName` contains the filename which should get precedence over the `type.name` to make stubbing work correctly.

### DIFF
--- a/tests/components/ScriptSetupWithChildren.vue
+++ b/tests/components/ScriptSetupWithChildren.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { defineAsyncComponent } from '@vue/compat'
+import { defineAsyncComponent } from 'vue'
 
 import Hello from './Hello.vue'
 import ComponentWithInput from './ComponentWithInput.vue'
--- a/tests/components/ScriptSetupWithChildren.vue
+++ b/tests/components/ScriptSetupWithChildren.vue
@@ -1,7 +1,10 @@
 <script setup lang="ts">
+import { defineAsyncComponent } from '@vue/compat'
+
 import Hello from './Hello.vue'
 import ComponentWithInput from './ComponentWithInput.vue'
 import ComponentWithoutName from './ComponentWithoutName.vue'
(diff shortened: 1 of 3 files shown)


tests/components/ScriptSetupWithChildren.vue
```diff
@@ -1,7 +1,10 @@
 <script setup lang="ts">
+import { defineAsyncComponent } from '@vue/compat'
```
cexbrayat, line 2: Do you use `@vue/compat` here on purpose instead of `vue`?

cexbrayat (commented):
Looks reasonable.

@xanf is that OK with you, as you changed this recently?

tests/components/ScriptSetupWithChildren.vue
Evertvdw: No, IDE imported that when I  used `defineAsyncComponent`. Figured it was something project specific as I don't have that in my own projects. But can change to `vue`.

cexbrayat: I think it would be better, thanks!

Evertvdw: Will do that!

Evertvdw:
@cexbrayat I just ran `yarn run lint` in the project in my branch but that has 100+ warnings. Seemed to also be indicated here: https://github.com/vuejs/vue-test-utils-next/runs/4445921989?check_suite_focus=true. Is that something I need to fix because I did not change anything in those files :)

cexbrayat:
@Evertvdw It looks like it fails in files you updated (`tests/components/ScriptSetupWithChildren.vue`). Run `yarn lint:fix` and you should be good to go. If you have other errors, make sure you have up-to-date dependencies, as it should not be the case?

Evertvdw:
Probably some git glitch, I ran the `lint:fix` and git changes said 105 changed files, but when comparing there were no differences for 104 out of 105. Fixed the one error that was there though :)

Issue #972, opened 2021-09-24
Better types for `wrapper.vm` when the component is closed


cexbrayat:
Since `2.0.0-rc.15`, we are now exposing `$.proxy` as `wrapper.vm` to simplify testing of `script setup` component,
allowing to test the following component:

```vue
<script setup lang="ts">
import { ref } from 'vue'
const count = ref(0)
</script>
```

with `expect(wrapper.vm.count).toBe(0)`, even if `count` is not exposed by the component (see PR #931 ).

This works, but for TS (`vue-tsc`), `wrapper.vm` does not have a `count` property, because `wrapper.vm` is typed as `ComponentPublicInstance`.

Ideally, we would like `wrapper.vm` type to reflect that it is actually not the component public instance, but the properties exposed to the template. Is there a way to infer that?

To reproduce, remove `"exclude": ["tests/expose.spec.ts"]` from `tsconfig.volar.json`, and run `yarn vue-tsc`. It fails with:

```
tests/expose.spec.ts:38:23 - error TS2339: Property 'count' does not exist on type '{ $: ComponentInternalInstance; $data: {}; $props: Partial<{}> & Omit<Readonly<{} & {} & {}> & VNodeProps & AllowedComponentProps & ComponentCustomProps, never>; ... 10 more ...; $watch(source: string | Function, cb: Function, options?: WatchOptions<...> | undefined): WatchStopHandle; } & ... 4 more ... & ComponentC...'.

38     expect(wrapper.vm.count).toBe(1)
```

cexbrayat:
@pikax do you have an idea how we could solve that?

pikax:
Currently that's not possible, I'm working on this PR https://github.com/vuejs/vue-next/pull/4465 to improve some types, I can have a look on how to also add that info there

cexbrayat:
For anyone keeping up with this, @johnsoncodehk released vue-tsc@0.30.0 with a new option that improves the situation. See #1170

felixzapata:
The explicit cast solution stops working with the last Vue release (3.2.45).

cexbrayat:
@felixzapata are you using VTU v2.2.2 ? There was a breaking change in Vue v3.2.45, but the latest VTU release should work around it.

felixzapata:
sorry, I forgot to mention that with VTU 2.2.2 also fails

cexbrayat:
[...]

PR #1170, opened 2021-12-22
chore: use experimentalExposeScriptSetupContext for vue-tsc

cexbrayat:
This new option allows a better typechecking of components that expose variables.
See https://github.com/johnsoncodehk/volar/issues/805

Refs #972 as this partially fixes it

### DIFF
--- a/tests/expose.spec.ts
+++ b/tests/expose.spec.ts
@@ -25,10 +25,11 @@ describe('expose', () => {
     const wrapper = mount(DefineExposeWithRenderFunction)
 
     // other is exposed vie `expose`
+    // @ts-ignore upstream issue, see https://github.com/vuejs/vue-next/issues/4397#issuecomment-957613874
     expect(wrapper.vm.other).toBe('other')
     // can't access `msg` as it is not exposed
     // and we are in a component with a setup returning a render function
-    expect(wrapper.vm.msg).toBeUndefined()
+    expect((wrapper.vm as unknown as { msg: undefined }).msg).toBeUndefined()
   })
 
   it('access vm with <script setup> and defineExpose()', async () => {
--- a/tests/emit.spec.ts
+++ b/tests/emit.spec.ts
@@ -1,5 +1,4 @@
 import {
-  DefineComponent,
   defineComponent,
   FunctionalComponent,
   getCurrentInstance,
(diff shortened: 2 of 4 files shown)


cexbrayat:
@johnsoncodehk I gave a try to the option, as we have a test file `expose.spec.ts` in the codebase that was not type checking previously (and was excluded form what vue-tsc checks).

 Using the option, 2 errors went away 👍

We still have one unexpected error, for a component defined with `defineComponent` and exposing its variable with `expose`: in that case, vue-tsc still thinks `wrapper.vm` does not have access to the property:

```
tests/expose.spec.ts:28:23 - error TS2339: Property 'other' does not exist on type '{ $: ComponentInternalInstance; $data: {}; $props: Partial<{}> & Omit<Readonly<{} & {} & {}> & VNodeProps & AllowedComponentProps & ComponentCustomProps, never>; ... 10 more ...; $watch(source: string | Function, cb: Function, options?: WatchOptions<...> | undefined): WatchStopHandle; } & ... 4 more ... & ComponentC...'.

28     expect(wrapper.vm.other).toBe('other')
```

It also raises an error that was not existing before in `emit.spec.ts`, where we have `[EmitsEventSFC, EmitsEventScriptSetup] as DefineComponent[]`:
```
tests/emit.spec.ts:349:11 - error TS2352: Conversion of type 'DefineComponent<{}, { click: () => void; }, {}, {}, {}, ComponentOptionsMixin, ComponentOptionsMixin, "bar"[], "bar", PublicProps, Readonly<...> & { ...; }, {}>[]' to type 'DefineComponent<{}, {}, {}, ComputedOptions, MethodOptions, ComponentOptionsMixin, ComponentOptionsMixin, ... 4 more ..., {}>[]' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.

349   it.each([EmitsEventSFC, EmitsEventScriptSetup] as DefineComponent[])(
```

We can workaround all this, but maybe that should be fixed in vue-tsc?

tests/expose.spec.ts
```diff
@@ -28,7 +28,7 @@ describe('expose', () => {
     expect(wrapper.vm.other).toBe('other')
     // can't access `msg` as it is not exposed
     // and we are in a component with a setup returning a render function
-    expect(wrapper.vm.msg).toBeUndefined()
+    expect((wrapper.vm as unknown as { msg: undefined }).msg).toBeUndefined()
```
lmiller1990, line 31: Should we be ts-ignoring these or is there more work to come (maybe in Volar) before merging this?

cexbrayat: This one is expected to not compile without `as`, as the test is just to check it is _not_ exposed 😉 So I think it's fine to explicitly cast it to make TS happy.

cexbrayat:
@johnsoncodehk mentioned on Discord that the first part of the issue is an upstream one https://github.com/vuejs/vue-next/issues/4397#issuecomment-957613874

I think I'll rewrite the test to get rid of the second one, and maybe ts-ignore the first error while we wait for the issue resolution in vue-next

cexbrayat:
Done. At the price of one `ts-ignore` due to the upstream issue, vue-tsc is now capable of type-checking all our tests and components. I also had to rewrite a test to split it in 2, but I don't think this is big deal.

PR #1172, opened 2021-12-23
repro: find inside suspense when root element

cexbrayat:
This is a regression with rc.18


lmiller1990:
Should be good now? https://github.com/vuejs/test-utils/pull/1397

cexbrayat:
We can 👍

Issue #945, opened 2021-09-14
Stubs rendering default props


robokozo:
When rendering snapshots, stubbed components are actually being evaluated in order to apply the default props to the element.

For example if my component imports and contains the following code:

```      
<Button
      ref="cancel-button"
      label="Cancel"
      :disabled="isLoading"
      icon="pi pi-times"
      @click="cancelClickHandler"
      class="p-button-text"
    />
```

And I stub out the `Button` component in my test

```
mountOptions = {
      global: {
        stubs: ["Button"]
      }
    }
```
My snapshot will render with
`
<button-stub label="Cancel" icon="pi pi-times" iconpos="left" loading="false" loadingicon="pi pi-spinner pi-spin" disabled="false" class="p-button-text"></button-stub>
`

`iconpos="left"` and `loadingicon="pi pi-spinner pi-spin" ` should not be rendered.

This is annoying for me because whenever I update my `Button` component my snapshots for anything using the `Button` component need to be updated. I already have my `Button` component covered with its own tests so this is unnecessary.

lmiller1990:
Where are `loadingicon` and `iconpos` coming from?

Snapshots + stubs are difficult to handle - seems like there are varying opinions on what should and should not be rendered. I wonder if we can provide a few different snapshot serializers, so people can choose the snapshot behavior they'd like. 🤔

robokozo:
They are defined on the actual "Button" component that I'm stubbing out. These specific props are defined with default values that are being rendered by the snapshot. (In this particular case they are defined in the PrimeVue component library I'm using)

I'm assuming that the actual Button component is being evaluated before the stubbing process takes place.

robokozo:
> I'm assuming that the actual Button component is being evaluated before the stubbing process takes place.

I ran into another issue which leads me to believe this is the case and the stubbing process should be examined. In this new situation, one of my stubbed components has dependencies which are not accounted for in the context of the test. The test crashes... but it shouldn't matter because the component should be stubbed out.

lmiller1990:
What do you mean by "has dependencies"? As in it imports other modules? The stubbing VTU offers is just before the component is rendered; it will still get imported (as well as all its dependencies).

Assuming my understanding of your question is correct, if you want to completely and totally stub out a component (and the modules it imports) you could use `jest.mock` potentially.

robokozo:
Yeah I think you understood correctly. Your idea to use jest.mock sounds really interesting! I'll give it a go!

[...]

PR #1174, opened 2021-12-23
feat: Hide props with default values within snapshot

freakzlike:
This PR will avoid props with their default values within snapshot (Fixes https://github.com/vuejs/vue-test-utils-next/issues/945).

```vue
<template>
  <my-component />
</template>
```
Will now be stubbed to `<my-component-stub />` and no longer to `<my-component-stub some-props="default-value" />`.

This may lead to invalid snapshots and could be a breaking change?

I currently use `JSON.stringify` for object comparison but I want to replace it.

https://github.com/freakzlike/vue-test-utils-next/blob/efca188cf62bea1f0e4ed07c9cd4766016c1c71c/src/stubs.ts#L115-L116
```ts
// TODO: use some different object comparison
return JSON.stringify(defaultValue) === JSON.stringify(propValue)
```

Is there already any existing util for this? (I haved found one). If not I will create one.


lmiller1990:
Will this only impact `shallowMount`? Eg, for components using `mount`, I think all the prop values should be shown?

freakzlike:
It only impact stubbed components so `shallowMount` and `stubs` in `mount`.
I'm also thinking if the props should be sorted by name. I faced some discrepancies of snapshots a while ago because I just changed the order of how the props are passed to the component.

freakzlike:
I have added a util function `deepCompare` and determine the default values of the props during stub setup and not on every render

lmiller1990:
I'd like some more feedback on this one, maybe @xanf or @cexbrayat or @afontcu ?

My concern is PR adds a lot of complexity that I foresee as difficult to maintain and also will make it very hard to make changes without breaking snapshots. 

I'm thinking that VTU should not even be concerned with snapshots; this is a Jest only feature, which can be heavily customized with their snapshot serializer API. Furthermore, different people have different opinions on snapshots, and I don't think we should be enforcing any opinions on them. Knowing the default prop value is likely important for some people.

Finally, this fix is specific to snapshots with **shallowMount**, a type of test which is going to be brittle and break at the slightest change to any implementation detail (good tests should be resilient to implementation details changing). How useful and beneficial is providing a highly opinionated snapshot, a feature where the desired behavior is different for each dev and testing framework?

I feel like this is going to cause a lot of overhead in the future and we may unintentionally break snapshots in minor versions. I'd like to get some more feedback before we merge this. Thanks for the work so far!

freakzlike:
I understand your concerns. Another proposal could be to override the default stub generation by setting a function in the global mounting options. So something like:

```ts
const wrapper = shallowMount(MyComponent, {
  global: {
    createStubs: ({ name, component }) => {
      // Implement own logic to create a stub
      return defineComponent({
        name,
        render: () => // Custom render logic
      })
    }
  }
})
```

This will allow the user the customize the default stub generation when using `shallowMount` or `stubs: { ChildComponent: true }`. Maybe some generic function could also be provided by a plugin package

cexbrayat:
I also tend to think it adds too much complexity for the benefits...

afontcu:
Hi! First of all, thank you for your work on this, I actually learned a couple of things reading your code 🙌 

That being said – I agree on the overall feeling of too much complexity – and also, a bit of opinionated. Assuming that component snapshot testing is useful, why wouldn't I get the default values of props? Or why wouldn't I want to break if prop order changes?

With that, I believe this PR should not be merged, as the behavior of snapshot testing is not something Vue Test Utils should control extensively.

Again, thanks!

lmiller1990:
If you want to make a snapshot serializer that implements this same behavior and need VTU to expose things, we could do that.

PR #1184, opened 2021-12-25
fix: add relative imports in wrapperLike interface

designervoid:
the problem was described here:
https://github.com/vuejs/vue-test-utils-next/issues/936#issuecomment-1000939410


lmiller1990 (approved):

designervoid:
☄️

PR #1185, opened 2021-12-25
docs: Correct difference in GlobalMountOptions type

freakzlike:
The API documentation of `GlobalMountOptions` has some difference

https://next.vue-test-utils.vuejs.org/api/#global vs. https://next.vue-test-utils.vuejs.org/api/#config-global


lmiller1990:
thanks!

PR #1186, opened 2021-12-27
Capitalize the first letter

victory820:
The content is wrong, when I run the test it will fail.

Best Wishes.


afontcu (approved):
nice catch, thanks!

PR #1193, opened 2021-12-29
repro: checkbox bug

lmiller1990:
Attempting to reproduce https://www.reddit.com/r/vuejs/comments/rqer1q/help_with_issue_testing_custom_checkbox_component/ but no luck yet.


lmiller1990:
User needed to add `attachTo`. Weird! We should doc that.

PR #1203, opened 2022-01-03
fix: support other class component implement

agileago:
vue3 class component shape is here 

https://github.com/vuejs/vue-next/blob/7007ffb2c796d6d56b9c8e278c54dc1cefd7b58f/packages/runtime-core/src/component.ts#L122-L125

```typescript
export interface ClassComponent {
  new (...args: any[]): ComponentPublicInstance<any, any, any, any, any>
  __vccOpts: ComponentOptions
}
```

but the library `mount` type shape is not adjust it. so we should support other class component implement

### DIFF
--- a/test-dts/mount.d-test.ts
+++ b/test-dts/mount.d-test.ts
@@ -1,8 +1,15 @@
 import { expectError, expectType } from './index'
 import {
+  ComponentOptions,
   DefineComponent,
   defineComponent,
-  FunctionalComponent
+  FunctionalComponent,
+  getCurrentInstance,
+  h,
+  ref,
+  SetupContext,
+  Prop,
+  VNodeChild
 } from 'vue'
 import { Options, Vue } from 'vue-class-component'
 import { mount } from '../src'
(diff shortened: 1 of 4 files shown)


cexbrayat:
@agileago Thanks for the PR!

Would you mind adding a test to show what this fixes? You can check `mount.d.ts` for examples.

agileago:
ok

test-dts/mount.d-test.ts
```diff
@@ -211,6 +212,48 @@ class ClassComponent extends Vue {
...
+  changeCount(count: number) {
+    this.count.value = count
+  }
+
+  render() {
+    return h('div', `hello world ${this.count.value}`)
+  }
+}
+
+// @ts-expect-error changeMessage expects an argument
```
cexbrayat, line 252: changeMessage -> changeCount

test-dts/mount.d-test.ts
```diff
@@ -1,8 +1,9 @@
 import { expectError, expectType } from './index'
 import {
+  ComponentOptions,
   DefineComponent,
   defineComponent,
-  FunctionalComponent
+  FunctionalComponent, h, ref, SetupContext
```
cexbrayat, line 6: nit: this should be formatted properly

test-dts/mount.d-test.ts
```diff
@@ -211,6 +212,48 @@ class ClassComponent extends Vue {
...
+      props: this.defaultProps,
+      setup(props, ctx) {
+        const instance = new CompConstructor()
+        instance.props = props
+        instance.context = ctx
+        ctx.expose(instance)
+        return instance.render.bind(instance)
+      }
+    }
+  }
```
cexbrayat, line 232: OOC: Is this something that developers really write? Why not extend the base `Vue` class?

agileago: where is the Vue class? i use it in my library [vue3-oop ](https://github.com/agileago/vue3-oop)

cexbrayat: `vue-class-component` exports a `Vue` class that components should extend (if I remember correctly).

cexbrayat:
@agileago The type check tests are failing, so this change is probably breaking something for the type inference. Can you take a look?

agileago:
git rebase is so hard...

PR #1212, opened 2022-01-04
feat: support custom class component

agileago:


src/mount.ts
```diff
@@ -73,6 +74,25 @@ function getInstanceOptions(
   return resultOptions
 }
 
+// Standard Class component
```
cexbrayat, line 77: nit: I would remove "Standard" (here and everywhere), as I suspect most people won't write this component like this. I would use "Class component (without vue-class-component)" instead.

src/mount.ts
```diff
@@ -73,6 +74,25 @@ function getInstanceOptions(
...
+export function mount<V>(
+  originalComponent: {
+    new (...args: any[]): V
+    __vccOpts: any
+  },
+  options?: MountingOptions<any> & Record<string, any>
+): VueWrapper<ComponentPublicInstance<V>>
+
+// Standard Class component - props
```
cexbrayat, line 86: Same here "Class component (without vue-class-component) - props"

test-dts/mount.d-test.ts
```diff
@@ -211,6 +218,78 @@ class ClassComponent extends Vue {
 expectError(mount(ClassComponent, {}).vm.changeMessage())
 mount(ClassComponent, {}).vm.changeMessage('')
 
+// region standard class component implement
+class StandardClassComponent<Props extends {} = {}> {
```
cexbrayat, line 222: `class CustomClassComponent`

cexbrayat: (remove standard in the comment as well)

test-dts/mount.d-test.ts
```diff
@@ -211,6 +218,78 @@ class ClassComponent extends Vue {
...
+  render() {
+    return h('div', `hello world ${this.count.value}`)
+  }
+}
+
+// @ts-expect-error changeCount expects an argument
+expectError(mount(NoPropClassComponent, {}).vm.changeCount())
+mount(NoPropClassComponent, {}).vm.changeCount(2)
+
+interface WithProp_Props {
```
cexbrayat, line 265: We usually don't use underscores in interface names. Maybe just `ComponentProps`?

cexbrayat (changes requested):
Just a few name to change and we'll be good to merge 👍

agileago:
@cexbrayat thanks for your patient

lmiller1990 (approved):

PR #1213, opened 2022-01-05
docs: Add Transitions page

### LINKED ISSUE #1163: docs: how to test transitions still blank
https://next.vue-test-utils.vuejs.org/guide/advanced/transitions.html
the docs about how to test transition are still blank any updates about it?
image

afontcu:
Closes #1163

- Add `Transition` page. It's quite basic and simple, but it exposes the auto-stubbing behavior from VTU.
- Removes `Third-party integration` as I believe it is sensible to wait for some libraries to release stable versions for Vue 3 before suggesting ways to test them.

With that, the only page left blank is `Reusability & Composition`, which I believe is quite important and should be filled before hitting a stable version. I'll try to find time to work on it soon.


docs/guide/advanced/transitions.md
```diff
@@ -1 +1,53 @@
...
+
+<script>
+export default {
+  data() {
+    return {
+      show: false
+    }
+  }
+}
+</script>
```
cexbrayat, line 24: Maybe we could use a more modern way of defining components in the docs?
Something like:

```vue
<script setup>
import { ref } from 'vue'

const show = ref(false)
</script>
```

docs/guide/advanced/transitions.md
```diff
@@ -1 +1,53 @@
...
+  transition: opacity 0.5s ease;
+}
+
+.fade-enter-from, .fade-leave-to {
+  opacity: 0;
+}
+</style>
+```
+
+Since Vue Test Utils stubs built-in transitions, you could test the component above as you'd test any other component:
```
cexbrayat, line 37: could -> can (but I'm not a native English speaker, so you probably know better ^^)

docs/guide/advanced/transitions.md
```diff
@@ -1 +1,53 @@
...
+
+```js
+import Component from './Component.vue'
+import { mount } from '../../src'
```
cexbrayat, line 41: import { mount } from '@vue/test-utils'

cexbrayat (approved):

docs/guide/advanced/transitions.md
afontcu: I'm not native either, so I'll go ahead and change it because "can" sounds good to me, too 😂

docs/guide/advanced/transitions.md
afontcu: that's a good debate – we've been using SFC or the "old way" of defining components throughout the docs where needed, but I also feel we should demonstrate VTU works fine with `script setup` and refs and whatnot.

Do you think we should use arbitrary ways of defining components? Or maybe we should write a specific page for `script setup`?

pinging @lmiller1990 @xanf to share their thoughts :)

cexbrayat: Yes, it's hard to know what would be best. I would at least not use `data` and replace it with a setup function if we don't want to use `script setup` directly.

afontcu: OK 👌  so for the time being I'll replace it with a `setup` function, but let's keep on exploring the best way to showcase that VTU works with any combination of options API/options API with setup/`<script setup>`

lmiller1990: I think we should pick 1 style and standardize on in.

I'd say "just do what Vue core docs are doing" but those are all over the shop at the moment. It feels weird to push TS and script setup (although this is my favorite way to code Vue now) if the docs for other core libs are not doing the same.

From what I can see, we are using the options API mostly in the Vue next docs - I think we should use that here (for now) to be consistent. I'd like to move towards TS and script setup, but we should do this in one single go.

lmiller1990 (commented):
one comment around style, but looks great!

docs/guide/advanced/transitions.md
afontcu: I guess options API + `setup()` were sensible is a good default?

lmiller1990: Seems fine - I'd like to actually move everything to TS and script setup soonish, but we should discuss this first, and do it all in one go.

afontcu: great, then we're good to go! Already changed this page to `setup()`

afontcu:
I'll go ahead and merge this one, feel free to open up a new PR if you feel something needs changing!

PR #1226, opened 2022-01-08
chore: Use setup instead of compat render for stubs

freakzlike:
This PR will use the setup function for stubs instead of the render function with compatibility mode. Also some refactoring so transition and teleport stubs don't use separate function

### DIFF
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -4,7 +4,6 @@ import {
   TransitionGroup,
   Teleport,
   h,
-  ComponentPublicInstance,
   defineComponent,
   VNodeTypes,
   ConcreteComponent,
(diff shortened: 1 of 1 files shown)


cexbrayat (approved):
I like it, thanks @freakzlike 
I'll let @lmiller1990 or @xanf take another look, but LGTM

src/stubs.ts
```diff
@@ -190,8 +169,9 @@ export function stubComponents(
       stubs['transition-group']
     ) {
       return [
-        createTransitionStub({
-          name: 'transition-group-stub'
+        createStub({
+          name: 'transition-group',
+          renderStubDefaultSlot: true
```
lmiller1990, line 174: I guess this will break some shallow snapshots, but probably fine.

lmiller1990 (approved):
Thanks!

src/stubs.ts
freakzlike: The `-stub` is added inside of `createStub`

https://github.com/vuejs/vue-test-utils-next/blob/cee2c9ccceef04b256968dd0010d924e21b4c000/src/stubs.ts#L70-L76

Issue #1173, opened 2021-12-23
regression: can not find element inside suspense if root element


cexbrayat:
After the refactoring released in rc.18, we stumbled into a regression when updating in our projects

Failing test:

```ts
  test('returns the element if it is a root element inside Suspense', () => {
    const Async = defineComponent({
      // works if there is a root element
      // template: '<div><h1>Hello</h1><span id="my-span">There</span></div>'
      // otherwise does not find the element
      template: '<h1>Hello</h1><span id="my-span">There</span>'
    })
    const Component = defineComponent({
      components: { Async },
      template: '<Suspense><Async/></Suspense>'
    })

    const wrapper = mount(Component)
    expect(wrapper.get('#my-span')).not.toBeNull()
  })
```

Repro in #1172

cexbrayat:
@xanf Does that ring a bell? I can dig deeper, but you know these parts better than any of us now ^^

xanf:
Yeah. I'll take a look in 12h and provide an update

чт, 23 дек. 2021 г., 17:09 Cédric Exbrayat ***@***.***>:

cexbrayat:
👍 Thanks (there is no rush, take your time)

tangjinzhou:
`wrapper.findAll('#my-span')`  also not work at rc.18

PR #1227, opened 2022-01-09
fix: findAll root nodes when using render function

### LINKED ISSUE #1201: `findAll` returns nothing when using a render function
### Subject of the issue
`findAll()` returns nothing when the component is using a render function.

### Steps to reproduce
This as `TestComponent.vue`:

```vue
<template>
<render></render>
</template>

<script setup lang="ts">
import { h } from 'vue';

const render = () => [
  h('a', {}, 'first'),
  h('a', {}, 'second'),
  h('a', {}, 'third'),
];
</script>
```

Then in `TestComponent.spec.ts`:

```ts
import TestComponent from './TestComponent.vue';

describe('mcve', () => {
  test('should have all three anchors', () => {
    const wrapper = mount(TestComponent);

    expect(wrapper.findAll('a').length).toBe(3);
  });
});

```

### Expected behaviour
The test should pass.

### Actual behaviour
The test fails because `findAll` is unable to find the three anchors.

freakzlike:
This PR will fix #1201.

I just used the same logic of `find` for `findAll` and unified this into one method. So same logic for `find` and `findAll` (except ref selector for `find`)

### DIFF
--- a/src/baseWrapper.ts
+++ b/src/baseWrapper.ts
@@ -1,4 +1,4 @@
-import { isNotNullOrUndefined, textContent } from './utils'
+import { textContent } from './utils'
 import type { TriggerOptions } from './createDomEvent'
 import {
   ComponentInternalInstance,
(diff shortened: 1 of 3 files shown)


cexbrayat (approved):

cexbrayat:
Thanks @freakzlike for the PR!

Does it also fix #1173 ? Would you mind adding a test in the PR to check?

cc @xanf on his, as he may have started to work on that, and knows this code better than me.

freakzlike:
@cexbrayat I ran the unit test from #1172 with this changes but it was the same result as before.

lmiller1990 (approved):
Seems fine to me - all tests passing, which is the vector we need to rely on :+1:

Issue #1179, opened 2021-12-24
Extend stub via mounting options


ol1s:
I want to be able to "fake" the method of a stubbed component. Currently this is possible by providing my own defined stub at mount time:

```
const ParentComponent = {
  template: `<div><child-component id="childComponent" ref="childComponent" /></div>`,
  components: { ChildComponent },
  methods: {
    myAction() {
      return this.$refs.childComponent.childMethod()
    },
  },
}
```
```
test("calls method on child component", async () => {
  const wrapper = shallowMount(ParentComponent, {
    global: {
      stubs: {
        ChildComponent: {
          methods: {
            childMethod() {
              return "Test 1"
            },
          },
        },
      },
    },
  })
  expect(wrapper.vm.myAction()).toEqual("Test 1")
})
```

However, this approach fully replaces the stub that would be produced automatically by `shallowMount()` and I end up with this warning:

```
WARNING: Component is missing template or render function
```

Is there are way for to "extend" a stub so I can add faked methods etc but without having to redefine the template or render function each time?

freakzlike:
I'm looking to make some improvements to stubs. Maybe I can add this too

cexbrayat:
@ol1s There is currently no way to do this. I usually provide an empty template or render function. I tend to think this is not worth adding another feature to stubs (which are already quite complicated to maintain). What do the rest of team think?

ol1s:
Thanks both for the replies. 

I am currently providing a hardcoded template to solve my situation:

```
const wrapper = shallowMount(ParentComponent, {
    global: {
      stubs: {
        ChildComponent: {
          template: `<div v-bind="$props" />`,
          methods: {
            childMethod() {
              return "Test 1"
            },
          },
        },
      },
    },
  })
```

On reflection, this is a pretty small price to pay in order to avoid bloating the API with additional configuration options.

Issue #773, opened 2021-07-16
Stubbed components slots aren't been rendered when 2 deep.


rikbrowning:
I put together a test case to show what I mean. It is very difficult to describe it in words. Essentially the produced snapshot of the test is incorrect. It should include all stubbed components and their slot contents as well.

https://github.com/rikbrowning/vue-test-utils-next/commit/53a0a13dc65935b45d63ba0abbb7179811eb7670

rikbrowning:
I did some digging around the code and think I may have found the issue.

When creating a stub we are creating a render function and passing the ctx.$slots in directly. The issue here is I think that any child slots inside that stub are not getting passed through the vnodeTransform and therefore not being rendered correctly.

I changed the render function to the follow:
```js
  const render = (ctx: ComponentPublicInstance) => {
    const slots = ctx.$slots
      ? Object.keys(ctx.$slots)
          .map((k) => ctx.$slots[k])
          .map((f) => {
            if (f) return f()
          })
      : []
    return h(tag, ctx.$props, renderStubDefaultSlot ? slots : undefined)
  }
```
and it started rendering out stubs inside of the first stub.

I am sure that code could be far neater, I apparently suck at writing typescript, but I think that we need to call the slot functions before passing them into the `h` function. Reading the examples in the vue docs they all seem to call the slot function before passing the result to the `h` function.

rikbrowning:
I have taken a stab a reworking renderStubDefaultSlot, the issue being that it doesnt really provide full backwards compatibility with v1.
I pushed a commit https://github.com/rikbrowning/vue-test-utils-next/commit/d9309d48264031ec83a9211c8baced8f5e87a088 and I am looking for feedback.

`renderStubSlot` can be a `boolean` or an `object`. When `false` nothing is rendered in the slots. When `true` all slots passed in are rendered. Note that slots that do not exist in the component you are stubbing will still be rendered. This may cause some false positives by rendering slots that won't actually be rendered in the browser. When an `object` you can specify slot property values into the slots. Essentially allowing you to stub slot prop values.

```js
 config.renderStubSlot = {
        'component-with-slots-stub': {
          scoped: { boolean: true, string: 'test' }
        }
      }
```

The above config will pass in `{ boolean: true, string: 'test' }` to the `scoped` slot of `component-with-slots-stub`.

Thoughts?

lmiller1990:
I didn't get a chance to look at the reproduction or look into this deeply yet, but 

[...]

PR #1241, opened 2022-01-17
feat: Add createStubs plugin hook

### LINKED ISSUE #1216: Proposal: Hook to override default stub creation
As mentioned in https://github.com/vuejs/vue-test-utils-next/pull/1174#issuecomment-1002890497. I want to be able to customize the default creation of a stubbed component, when using `shallowMount` or `stubs: { ChildComponent: true }`.

This could look something like:
```typescript
const wrapper = shallowMount(MyComponent, {
  global: {
    createStubs: ({ name, component }) => {
      // Implement own logic to create a stub
      return defineComponent({
        name,
        render: () => // Custom render logic
      })
    }
  }
})
```

If set the hook would be called once for every component instead of `createStub`.
https://github.com/vuejs/vue-test-utils-next/blob/5bfe95cc71f6d07aff576a66bf450e6f23a26a36/src/stubs.ts#L282-L288

Unfortunately I was not able to find another way, without manually create a stub for every component. I tried to use `transformVNodeArgs` by myself but it will be replaced when I do `mount`.

freakzlike:
Adds a new plugin API to `config.plugins` called `createStubs` which will override the default stub creation provided by VTU.

Usage:
```ts
config.plugins.createStubs = ({ name, component }) => {
  return defineComponent({
    render: () => h(`custom-${name}-stub`)
  })
}
```
When VTU creates a stub either from `shallow: true` or `stubs: { ChildComponent: true }`, then the provided function will be called.

This will allow to implement different custom logics which are mentioned here:
* https://github.com/vuejs/vue-test-utils-next/issues/773#issuecomment-885182483
* https://github.com/vuejs/vue-test-utils-next/issues/945
* https://github.com/vuejs/vue-test-utils-next/pull/1174
* https://github.com/vuejs/vue-test-utils-next/issues/1179
* https://github.com/vuejs/vue-test-utils-next/issues/69

I have not updated the docs yet, but I will if it is good.

Closes #1216

### DIFF
--- a/docs/guide/extending-vtu/plugins.md
+++ b/docs/guide/extending-vtu/plugins.md
@@ -10,7 +10,9 @@ Some use cases for plugins:
 1.  Attaching matchers to the Wrapper instance
 1.  Attaching functionality to the Wrapper
 
-## Using a Plugin
+## Wrapper Plugin
+
+### Using a Plugin
 
 Install plugins by calling the `config.plugins.VueWrapper.install()` method
 . This has to be done before you call `mount`.
@@ -43,12 +45,12 @@ once. Follow the instructions of the plugin you're installing.
 
 Check out the [Vue Community Guide](https://vue-community.org/v2/guide/ecosystem/testing.html) or [awesome-vue](https://github.com/vuejs/awesome-vue#test) for a collection of community-contributed plugins and libraries.
 
-## Writing a Plugin
+### Writing a Plugin
 
 A Vue Test Utils plugin is simply a function that receives the mounted
 `VueWrapper` or `DOMWrapper` instance and can modify it.
 
-### Basic Plugin
+#### Basic Plugin
 
 Below is a simple plugin to add a convenient alias to map `wrapper.element` to `wrapper.$el`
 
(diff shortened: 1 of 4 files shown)


cexbrayat (commented):
This might be a good idea. @lmiller1990 WDYT?

lmiller1990:
Cool, but isn't this exactly the same as

```ts
import { config } from '@vue/test-utils'
config.globals.stubs = {
  // your stub
}
```

Maybe I did not read it properly or misunderstood - don't we already have the ability to create a stub globally like this?

freakzlike:
Ok, I will add the docs soon :)

It is not used to stub one specific global component. Instead this will be called for all stubs so something similar like this:
```typescript
import { config } from '@vue/test-utils'
config.globals.stubs = {
  '*': GlobalStub
}
```
Also the generated stub can by generic depending on the component which should be stubbed

lmiller1990:
I guess we can have this, although I'm not sure how much it'll be used.

I'd say this should be last feature before we move to 2.0 stable.

Issue #1277, opened 2022-01-25
Bug: Built-in `<component />` functionality seems altered when `shallow: true`


phobetron:
**Describe the bug**
When shallow-mounting a component that contains the built-in `<component />` component, its behavior seems to be altered.

**To Reproduce**
In my current case, I'm testing an SFC with `script setup` that has `<component :is="someComponent" />` in its template. `someComponent` is a computed property that returns a value of the type `Component`.

When testing with `shallow: true`,  a component named `<some-component />` is rendered. It seems to match the name of the computed property.

**Expected behavior**
When testing with `shallow: true`,  a component named `<custom-component />` is rendered. It should match the return value of the computed property.

**Related information:**
- `@vue/test-utils` version: 2.0.0-rc.18
- `Vue` version: 3.2.26
- `node` version: 16.13.0
- `npm` (or `yarn`) version: `yarn` 1.22.17

**Additional context**
When using `shallow: false`, or running normally in the browser, the built-in `<component />` component functions as expected, and `<custom-component />` is rendered.

I believe it is not possible to stub the functionality of `<component />` as it is a built-in component.

The components that are rendered using `<component />` are fairly complex, and are tested separately, so we are hoping to merely inspect the values passed to those components instead of reaching through their hierarchies and inspecting their behaviors.

phobetron:
I've typed it up based on what's in my codebase, but haven't actually run it. Hopefully, it will help explain the issue I'm running into, either way.

`OneOrOtherComponent.vue`
```vue
<template>
  <component :is="someComponent" />
</template>

<script setup>
import { computed } from 'vue'
import ComponentA from './ComponentA.vue'
import ComponentB from './ComponentB.vue'

interface Props {
  isB?: boolean
}

const props = defineProps<Props>()

const someComponent = computed(() => {
  return isB ? ComponentB : ComponentA
})
</script>
```

`OneOrOtherComponent.spec.ts`
```ts
import { mount } from '@vue/test-utils'
import { OneOrOtherComponent } from './'

describe('OneOrOtherComponent', () => {
  it('renders component B by default', () => {
    const wrapper = mount(OneOrOtherComponent, { shallow: true })

    // This outputs "<some-component />"
    console.log(wrapper.html())

    // This fails
    expect(wrapper.findComponent({ name: 'component-b' }).exists()).toBe(true)
  })
})
```

lmiller1990:
Thanks for the great report - I will try get some :eyes: on this soon.

[...]

PR #1312, opened 2022-02-14
test: Add tests for shallowMount issue with dynamic components

phobetron:
This PR adds a minimal set of tests to describe issue #1277, which regards how dynamic components, when shallow mounted, are stubbed with the name of a computed property that returns a component's name instead of the component itself.

The test component outlines the most extreme error case: a computed property does not need to be referenced in order to trigger the issue, it merely needs to both exist and return the component that is directly referenced.

This PR now also implements a possible simple fix for the issue.

cc: @cexbrayat

### DIFF
--- a/tests/shallowMount.spec.ts
+++ b/tests/shallowMount.spec.ts
@@ -198,24 +198,14 @@ describe('shallowMount', () => {
     )
   })
 
-  it('stubs a given component that is not returned by a computed property', () => {
+  it('stubs a given component that is also returned by a computed property', () => {
     const wrapper = shallowMount(DynamicComponentWithComputedProperty)
 
     expect(wrapper.find('hello-stub').exists()).toBe(true)
   })
 
-  it('stubs a given component that is also returned by a computed property', () => {
-    const wrapper = shallowMount(DynamicComponentWithComputedProperty, {
-      props: { isHello: true }
-    })
-
-    expect(wrapper.find('hello-stub').exists()).toBe(true)
-  })
-
   it('does not attempt to stub a dynamic component based on the name of a computed property', () => {
-    const wrapper = shallowMount(DynamicComponentWithComputedProperty, {
-      props: { isHello: true }
-    })
+    const wrapper = shallowMount(DynamicComponentWithComputedProperty)
 
     expect(wrapper.find('computed-property-stub').exists()).toBe(false)
   })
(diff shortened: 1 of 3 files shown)


tests/shallowMount.spec.ts
```diff
@@ -2,6 +2,7 @@ import { defineAsyncComponent, defineComponent } from 'vue'
 import { mount, shallowMount, VueWrapper } from '../src'
 import ComponentWithChildren from './components/ComponentWithChildren.vue'
 import ScriptSetupWithChildren from './components/ScriptSetupWithChildren.vue'
+import ScriptSetupWithBuiltinComponent from './components/ScriptSetupWithBuiltinComponent.vue'
```
cexbrayat, line 5: nit: I would rename to `DynamicComponentWithComputedProperty` or something like that.

tests/components/ScriptSetupWithBuiltinComponent.vue
```diff
@@ -0,0 +1,19 @@
+<template>
+  <component :is="Hello" />
```
cexbrayat, line 2: shouldn't that use `computedProperty`?

phobetron: @cexbrayat This is what I meant when I added the note. It does not matter whether you _use_ the computed property, its name is still what `shallowMount` will give to `component`. That's how weird this bug is.

phobetron: I'll add a third test to better outline the issue.

cexbrayat:
Hmm strange indeed, thanks for the repo 👍

If you feel adventurous, you can try to inspect what's going on when the tests run, and you may be able to fix the issue: we would be very happy to accept the PR!

phobetron:
@cexbrayat The issue was exactly where @freakzlike had suggested. My fix is to ensure `getComponentNameInSetup` doesn't get tripped up by any getter functions that may return a component object. All the tests now pass. Please let me know what you think.

lmiller1990 (approved):
Seems fine! @cexbrayat ?

lmiller1990:
@phobetron can you fix the linting errors?

phobetron:
@lmiller1990 Sorry about that! I think my local has a different default line length. Should be fixed now.

cexbrayat (approved):
Awesome, thanks 👍

phobetron:
@cexbrayat That makes sense. I've reduced the test code to the minimum necessary to cover the error case. Should pass everything now 😅

phobetron:
@cexbrayat Sorry, decided to make another minor style adjustment.

cexbrayat:
Looks great, thank you for taking the time to investigate and fix this, we really appreciate it!

PR #1314, opened 2022-02-14
Update vue router docs

### LINKED ISSUE #1315: Mocking router (components) in tests not fully documented
### Documentation Feedback

After updating my project to the latest next versions of Vue/test-utils etc. I got into an issue were using a real router would cause side effects, cause it's a singleton and Jest runs tests in parallel, so a route change in your component would affect another test and hence it'd fail.

Reading the docs I felt it wasn't obvious that you should *never* import your router, instead you should instantiate a new router with the routes from your router file for each test, like so: https://github.com/Joralf/vue-test-utils-router/blob/master/tests/unit/component-without-mock-with-router-per-test.spec.ts#L11

I think this can be added to this section: https://next.vue-test-utils.vuejs.org/guide/advanced/vue-router.html#using-a-real-router

Secondly, I feel that testing with "a real router" is unnecessary, so I started using mocks. However, from the documentation here https://next.vue-test-utils.vuejs.org/guide/advanced/vue-router.html#testing-userouter-and-useroute-within-setup it is not clear how you should stub `router-link` and `router-view` in case you want to mock your router. The documentation only explains how to get rid of the warnings related to these components when using a real router: https://next.vue-test-utils.vuejs.org/guide/advanced/vue-router.html#using-a-real-router

Google provides a few outdated answers, so imho it would be best if the docs would describe how to do this like so: https://github.com/Joralf/vue-test-utils-router/blob/master/tests/unit/component-with-mock.spec.ts#L31
or so https://github.com/Joralf/vue-test-utils-router/blob/master/tests/unit/component-with-mock.spec.ts#L61 and maybe add a link to the global.stub section in the docs (https://next.vue-test-utils.vuejs.org/api/#global).

I know these are really minor changes and you're probably busy, so I've added a PR below to change some things myself.

**Affected documentation page:** Insert a link to the affected page
https://next.vue-test-utils.vuejs.org/guide/advanced/vue-router.html

Joralf:
As described in the following issue (https://github.com/vuejs/test-utils/issues/1315) I felt the documentation regarding vue-router was lacking a bit, therefore I created a PR to improve the docs. Let me know if anything should be changed.

All the best!

resolves #1315

### DIFF
--- a/docs/guide/advanced/vue-router.md
+++ b/docs/guide/advanced/vue-router.md
@@ -301,7 +301,7 @@ test('routing', async () => {
 
 It _finally_ passes. Great! This is all very manual, however - and this is for a tiny, trivial app. This is the reason using a mocked router is a common approach when testing Vue components using Vue Test Utils.
 
-## Testing useRouter and useRoute within setup
+## Using a mocked router with Composition API
 
 Vue router 4 allows for working with the router and route inside the `setup` function with the composition API.
 
@@ -361,9 +361,7 @@ test('allows authenticated user to edit a post', () => {
       isAuthenticated: true
     },
     global: {
-      mocks: {
-        // No need for the mocks
-      }
+      stubs: ["router-link", "router-view"], // Stubs for router-link and router-view in case they're rendered in your template
     }
   })
 
@@ -388,6 +386,9 @@ test('redirect an unauthenticated user to 404', () => {
   const wrapper = mount(Component, {
     props: {
       isAuthenticated: false
+    }
+    global: {
+      stubs: ["router-link", "router-view"], // Stubs for router-link and router-view in case they're rendered in your template
     }
   })
 
--- a/docs/guide/advanced/vue-router.md
+++ b/docs/guide/advanced/vue-router.md
@@ -188,9 +188,11 @@ The `<router-link>` and `<router-view>` component are not found. We need to inst
 ```js {10,11,12}
 import { mount } from '@vue/test-utils'
 import { createRouter, createWebHistory } from 'vue-router'
+import { routes } from "@/router" // This import should point to your routes file declared above
 
 const router = createRouter({
-  // omitted for brevity
+  history: createWebHistory(),
+  routes: routes,
 })
 
 test('routing', () => {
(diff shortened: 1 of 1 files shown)


docs/guide/advanced/vue-router.md
```diff
@@ -134,26 +134,32 @@ const Posts = {
...
-    },
-    {
-      path: '/posts',
-      component: Posts
-    }
-  ]
+  routes: routes,
 })
+
+export { routes };
```
cexbrayat, line 160: I suppose you exported the routes to reuse them later, but I don't see where.

cexbrayat (commented):
Nice, thank you for the PR!

In my opinion, we should show more of vue-router-mock, as it simplifies the tests a great deal in real applications. But that's still better, so thanks 👍

docs/guide/advanced/vue-router.md
Joralf: > Nice, thank you for the PR!
>
>In my opinion, we should show more of vue-router-mock, as it simplifies the tests a great deal in real applications. But that's still better, so thanks 👍

Is there a ticket / discussion for this somewhere? I'd be willing to try and create better examples with regards to mocking vue-router, but I'd need some input on what you folks think is good practice. Let me know if I can help!

cexbrayat (approved):

docs/guide/advanced/vue-router.md
cexbrayat: 👍

I don't think there is a ticket to track this, but the idea is to use `createRouterMock`.
For example, in a route `/users/:userId` that displays a `User` component:
```ts
const router = createRouterMock();

function userWrapper() {
  injectRouterMock(router);
  // fake router params
  router.setParams({
    userId: 12
  });
  return mount(User);
}
```

and then in the tests, you can easily check when the router has been called to navigate, because `router.push` is spied on:
```ts
expect(router.push).toHaveBeenCalledWith('/');
```

The advantage is that you don't have to export your routes and really trigger a navigation: you just test your component in isolation, set the params you want, and can check on the navigations triggered.

If you want to open another PR, that would be awesome 👍

lmiller1990 (approved):

docs/guide/advanced/vue-router.md
Joralf: So far I stayed away from vue-router-mock, cause I feel adding more external dependencies on top of other dependencies increases the risk of it not being compatible in case Vue has another major upgrade. In this case we're talking about Vue > Jest > TestUtils and then adding vue-router-mock as another dependency. That's 4 levels deep :)

But I'm willing to at least give other people the option with some better docs, so let me explore that next week :)

PR #1326, opened 2022-02-16
The wrapper methods props support functional props

edisonLzy:


cexbrayat:
That seems a reasonable fix, thanks @edisonLzy 👍

You need to reformat the code though, the CI is failing on the lint task. Run `yarn lint:fix` and that should be good to go

PR #1345, opened 2022-02-24
Add docs for find by element ref

visualjerk:
Since https://github.com/vuejs/test-utils/pull/1094 find supports element refs. This PR adds the new use case to the docs.


cexbrayat (approved):
👍 thanks!

PR #1364, opened 2022-03-08
chore: give credit to original flushPromises lib

marina-mosti:
I think due credit is deserved here to original repo


cexbrayat (approved):
Thanks @marina-mosti 👍

PR #1377, opened 2022-03-21
fix: do not include self when using findAll

lmiller1990:
resolves https://github.com/vuejs/test-utils/issues/1233

Seems too easy 🤔

### DIFF
--- a/src/baseWrapper.ts
+++ b/src/baseWrapper.ts
@@ -41,7 +41,10 @@ export default abstract class BaseWrapper<ElementType extends Node>
     this.wrapperElement = element
   }
 
-  protected findAllDOMElements(selector: string, ignoreSelf = false): Element[] {
+  protected findAllDOMElements(
+    selector: string,
+    ignoreSelf = false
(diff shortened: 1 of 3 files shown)


lmiller1990:
cc @freakzlike @aethr WDYT?

xanf:
:eyes: Looking on this right now

lmiller1990:
@xanf pushed a commit to fix that.

xanf:
@lmiller1990 looking

xanf:
@lmiller1990 while reviewing your PR I've found another failing case - https://github.com/vuejs/test-utils/pull/1392/files#diff-9835123bc80f420fd7ebbcc25a72c73e7f2c8934bac14a9a3bbe6e1dca3569f7R356

And I believe I've figured more elegant approach in #1392 
Could you please look at that (basically we simply separated implementations of `findAll` for VueWrapper and DOMWrapper.
Feel free to close that PR is favor of this one if you feel that this approach is better

/cc @cexbrayat

PR #1381, opened 2022-03-22
fix: pass RouterLink slot props to RouterLinkStub slot

### LINKED ISSUE #848: RouterLinkStub doesn't support scoped slot used with `custom` prop
The RouterLink component in VueRouter supports a `custom` prop that enables use of a scoped slot (https://router.vuejs.org/api/#v-slot-api-3-1-0). It can be used to create custom link element like:

```html
<router-link
  to="/about"
  custom
  v-slot="{ href, route, navigate, isActive, isExactActive }"
>
  <NavLink :active="isActive" :href="href" @click="navigate"
    >{{ route.fullPath }}</NavLink
  >
</router-link>
```

Testing a component that uses this feature using the `RouterLinkStub` provided by VTU causes an error because no value is available to the scoped slot:

`TypeError: Cannot destructure property 'href' of 'undefined' as it is undefined.`

I fixed this in my broken test by copying RouterLinkStub and passing some dumb values through the scoped slot in the render function:

```js
const RouterLinkStub = defineComponent({
  name: 'RouterLinkStub',
  props: ['to'],

  render() {
    return h('a', undefined, this.$slots?.default?.({
      href: '/places',
      navigate: () => {},
      isActive: false,
    }));
  },
});
```

I wonder if something similar would be a useful addition to the provided stub in VTU?

aethr:
resolves #848 

Added a default "empty" route object based on what I've seen coming out of vue-router, and basic stupid values for the rest of the slot props. In my opinion, anyone who needs `RouterLink` to be a bit smarter can create their own stub or use the real thing.

I also added the `custom` prop behaviour since it was easy to lift from the real `RouterLink` but it's probably not necessary, I'd be happy to remove that.

I added some tests around `RouterLinkStub` just to verify my implementation, also happy to remove them if they add unnecessary weight to the test suite.

### DIFF
--- a/src/components/RouterLinkStub.ts
+++ b/src/components/RouterLinkStub.ts
@@ -1,4 +1,18 @@
-import { defineComponent, h } from 'vue'
+import { defineComponent, h, computed } from 'vue'
+
+// match return type of router.resolve: RouteLocation & { href: string }
+const defaultRoute = {
+  path: '/',
+  name: undefined as string | undefined,
+  redirectedFrom: undefined as object | undefined,
+  params: {},
+  query: {},
+  hash: '',
+  fullPath: '/',
+  matched: [] as object[],
+  meta: {},
+  href: '/'
+}
 
 // TODO: Borrow typings from vue-router-next
 export const RouterLinkStub = defineComponent({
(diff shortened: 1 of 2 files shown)


aethr:
I also noticed there doesn't appear to be documentation around using `RouterLinkStub`. Happy to add something if this is a gap that should be closed, although some advice one where to best do that would be appreciated.

aethr:
Took me a while to get the linter/TS happy. Not much of a TypeScript developer, but I reckon borrowing some types from vue-router might be handy here, if we can do it without adding a full dependency?

lmiller1990:
Nice job. I'm not sure we can import the type from Vue Router, unfortunately - although if someone is using this component, they'd almost certainly have it installed. Not sure on the best practice here, but this is a good enough start.

We should likely document this, we have it in the v1 docs: https://v1.test-utils.vuejs.org/api/components/ if you want to make a follow up PR, that'd be great.

PR #1391, opened 2022-03-26
docs: add RouterLinkStub to api docs

aethr:
Copies `RouterLinkStub` portion from the v1 api docs with a small addendum re: slot content.

Added after a comment in #1381 pointed out these were missing.


cexbrayat (approved):
LGTM, thanks!

PR #1392, opened 2022-03-26
fix(find): find and findAll should not find itself on DOM wrappers

### LINKED ISSUE #1233: Bug: `findAll` returns parent node in result array
**Describe the bug**
`findAll` returns the parent node in the result set if it matches the selector when chained from `get`

**To Reproduce**

`ParentComponent.vue`
```vue
<template>
  <div class="parent">
    <div>child</div>
    <div>child</div>
    <div>child</div>
  </div>
</template>
```
`ParentComponent.spec.js`
```js

import ParentComponent from '../ParentComponent';

describe('parent', () => {
  test('finds 3 children', () => {
    const wrapper = mount(ParentComponent);

    const parent = wrapper.get('.parent');

    expect(parent.findAll('div').length).toBe(3);
  });
});
```
`result`
```
parent
    ✕ finds 3 children (21 ms)

  ● parent › finds 3 children

    expect(received).toBe(expected) // Object.is equality

    Expected: 3
    Received: 4
```

**Expected behavior**
`findAll` returns only the 3 child nodes in the result array. This is the behaviour of `vue-test-utils@1`

**Related information:**
- `@vue/test-utils` version: `2.0.0-rc.18`
- `Vue` version: `3.2.26`
- `node` version: `16.13.1`
- `npm` (or `yarn`) version: `npm@8.1.2`

**Additional context**
I'm aware of the open PR reworking `find` and `findAll`, as well as a few mentions in other issues relating to `findAll` of a contributor currently working on changes to `findAll`. Maybe the open PR or the work in progress fixes this, but I thought it was worth documenting.

xanf:
This is alternate approach to fix #1233 
Other approach: #1377

What I'm liking about this approach:

* we're not introducing any flags 
* we're just moving `findAll` implementation to `DOMWrapper`/`VueWrapper` to distinguish behavior on Vue wrapper / DOM (semantics :1st_place_medal: )

### DIFF
--- a/tests/find.spec.ts
+++ b/tests/find.spec.ts
@@ -1,7 +1,7 @@
 import { defineComponent, h, nextTick, Fragment } from 'vue'
-
 import { mount, VueWrapper } from '../src'
 import SuspenseComponent from './components/Suspense.vue'
+import ParentComponent from './components/ParentComponent.vue'
 import MultipleRootRender from './components/MultipleRootRender.vue'
 
 describe('find', () => {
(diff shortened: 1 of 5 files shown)


tests/find.spec.ts
```diff
@@ -93,18 +94,6 @@ describe('find', () => {
     expect(wrapper.find('.foo').exists()).toBe(true)
   })
 
-  it('returns the root element from dom wrapper if it matches', () => {
```
xanf, line 96: This test now makes no sense, it captured previous behavior

I've added alternate test at the bottom of this file

lmiller1990 (approved):

lmiller1990:
Let's ship it :muscle:

PR #1180, opened 2021-12-24
fix: add methods from baseWrapper to the wrapperLike interface

kensodemann:

PR #1401, opened 2022-04-06
fix: make imports in wrapperLike relative (again)

### LINKED ISSUE #936: Build fail due to wrong import in domWrapper.ts

catrope:
This fixes #936 (again). It was fixed by #1184, but then re-broken
by #1180.


cexbrayat (approved):
Thanks @catrope :+1:

PR #1409, opened 2022-04-08
fix: disable stubs for built-in components on shallow

### LINKED ISSUE #929: Unable to stub transition component
I've tried to stub the Vue Transition component but was unable to do so. I was able to unstub the component with `transition: false` (since it is stubbed by default) but if I want to provide a custom stub, e.g. `transition: {template: '<p></p>'}`, it gets ignored.

Here is a repo with a custom transition stub: https://github.com/Thomasan1999/vue-test-utils-next-transition.

freakzlike:
fix #929 

I came across an equal issue when using `shallow: true` but still want to use the real `teleport`. This ensures to use the real component when using
```js
const wrapper = mount(Comp, {
  shallow: true,
  global: {
    stubs: {
      transition: false,
      teleport: false,
      'transition-group': false
    }
  }
})
```

### DIFF
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -2,6 +2,7 @@ import {
   transformVNodeArgs,
   Transition,
   TransitionGroup,
+  BaseTransition,
   Teleport,
   h,
   defineComponent,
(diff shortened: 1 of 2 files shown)


cexbrayat (approved):
LGTM, I'll let @lmiller1990 take a look and merge

lmiller1990 (approved):
Seems fine - I think it's weird we have so much special code around `shallow` - I generally just use `mount` for this reason :shrug:

freakzlike:
You are right. There are also quiet some open issues related to shallow and stubs. I think some things can be simplified or maybe ( for the future) VTU only offers some API for plugins.

PR #1412, opened 2022-04-11
docs: Add testing composables section

### LINKED ISSUE #713: [Docs] Test usage of provide
It took me quite some time to get an example of how to test a `provide` of a component. I want to test if the component uses the `provide` function correctly and if it does some stuff with the injected data from possible child components. Which is somehow the opposite of the existing option which is already in-build to `vue-test-utils`:

```ts
mount(ChildComponent, {
  global: {  
    provide: { 
      title: 'My title'
    }
  }
})
```

I finally got it working with the following code. Maybe this helps others and can eventually be added to the docs.

I had some interesting errors on my way:
- adding a `name` to the childComponent breaks it
- leaving the `h(childComponent)` away breaks the slot (inside child: `getCurrentInstance().parent` is `null`)

## Sample

```vue
<template>
  <div class="parent">
    <p class="title">{{ title }}</p>
    <slot />
  </div>
</template>

 <script lang="ts">
import {defineComponent} from 'vue';

export defineComponent({
    setup() {
      const title = ref('Parent');
      provide('title', title);
      return { title };
    },
});
</script>
```

```ts
import { h, inject, nextTick, provide, Ref, ref } from 'vue';
import parentComponent from './parentComponent';

describe('Repro', () => {
  it('should update injected title', async () => {
    expect.assertions(1);
    // given

    // the child component works like a mock to simulate missing inject option 
    const childComponent = {
      template: '<div />',

      setup() {
        const title = inject<Ref<string>>('title');
        if (title) {
          title.value = 'Child';
        }
        return {};
      },
    };

    // when
    const wrapper = mount(parentComponent, {
      slots: {
        default: h(childComponent),
      },
    });
    await nextTick();

    // then
    expect(wrapper.find('.title').text()).toStrictEqual(`Child`);
  });
});
```

freakzlike:
fix #713

### DIFF
--- a/docs/guide/advanced/reusability-composition.md
+++ b/docs/guide/advanced/reusability-composition.md
@@ -44,7 +44,7 @@ export function useUser(userId) {
   const user = ref()
   
   function fetchUser(id) {
-    fetch(`users/${id}`)
+    axios.get(`users/${id}`)
       .then(response => (user.value = response.data))
   }
 
@@ -58,6 +58,9 @@ To test this composable, you can create a simple `TestComponent` within the test
 composable the exact same way how the real components would use it.
 
 ```typescript
+// Mock API request
+jest.spyOn(axios, 'get').mockResolvedValue({ data: { id: 1, name: 'User' } })
+
 test('fetch user on mount', async () => {
   const TestComponent = defineComponent({
     props: {
--- a/docs/guide/advanced/reusability-composition.md
+++ b/docs/guide/advanced/reusability-composition.md
@@ -5,3 +5,88 @@ Mostly:
 - `global.provide`.
 - `global.mixins`.
 - `global.directives`.
+
+## Testing composables
+
+When working with the composition API and creating composables, you often want to test only the composable. Let's start
+with a simple example:
+
+```typescript
+export function useCounter() {
+  const counter = ref(0)
+
+  function increase() {
+    counter.value += 1
+  }
+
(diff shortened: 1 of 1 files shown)


cexbrayat:
@freakzlike Let me know when this draft PR is ready to be reviewed 👍

freakzlike:
@cexbrayat Please take a look. I'm not really good at writing documentions, but I hope it will help someone

docs/guide/advanced/reusability-composition.md
```diff
@@ -2,6 +2,214 @@
...
+
+  const wrapper = mount(TestComponent, {
+    props: {
+      userId: 1
+    }
+  })
+
+  expect(wrapper.vm.user).toBeUndefined()
+
+  await flushPromises()
```
cexbrayat, line 87: I suppose here you also need to mock the HTTP call one way or another.
IT's fine by me if we don't show how, but maybe a comment would be nice so readers don't expect this to work magically?

cexbrayat (approved):
Looks great! One minor comment, and I'm OK to merge 👍

docs/guide/advanced/reusability-composition.md
freakzlike: I added mocking the same way like in https://test-utils.vuejs.org/guide/advanced/async-suspense.html#resolving-other-asynchronous-behavior

PR #1418, opened 2022-04-12
chore: bump to vue v3.2.32

cexbrayat:


cexbrayat:
CI is failing because v3.2.32 changed how the root component is used in `createApp` https://github.com/vuejs/core/commit/d121a9bc7e7af59adb2d2803954cfeee95b35270; The root component is now copied before used.

This broke how we stub components, as we used to add the root component to the list of component to not stub.
As the component is now no longer the same, all tests using shallowMount are failing.

The fix is to register the copy of the component created by Vue in `createApp` (which is available as `app._component`)

lmiller1990 (approved):
Argh, shallow mount problems.

PR #1434, opened 2022-04-17
fix: forward Ref prop on mount

### LINKED ISSUE #192: consider about Ref as props when using mount api
I found VTU use this code to make reactive of props when `mount`:
```ts
const props = reactive({
    ...options?.attrs,
    ...options?.props,
    ref: MOUNT_COMPONENT_REF
  })
```
but if I have `const someRef = ref({})` and pass it to `mount`:
```ts
mount(App, {
  props: {
    someRef: someRef
  }
})
```
it will break the `Ref` structure

freakzlike:
fix #192

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -2,7 +2,7 @@ import {
   h,
   createApp,
   defineComponent,
-  reactive,
+  shallowReactive,
   FunctionalComponent,
   ComponentPublicInstance,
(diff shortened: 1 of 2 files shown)


lmiller1990:
Hmm interesting fix! I guess it doesn't break anything, although I have a sneaking suspicious this will break some weird edge cases. We'll see.

Issue #1195, opened 2021-12-30
FindComponent returns WrapperLike with css selector


frisch-raphael:
I'm trying to use findComponent with a css selector. 
However, when I try to use trigger on it, it says that trigger does not exist.
Is there something I'm missing ? 

![image](https://user-images.githubusercontent.com/3451172/147713672-373420b2-235a-4b36-b024-3b1e21de3d47.png)

Here's my package.json:
```
  "dependencies": {
    "@quasar/extras": "^1.0.0",
    "@vue/test-utils": "^2.0.0-rc.18",
    "axios": "^0.21.4",
    "core-js": "^3.6.5",
    "qrcode": "^1.5.0",
    "quasar": "^2.0.0",
    "vuex": "^4.0.2"
  },
  "devDependencies": {
    "@babel/eslint-parser": "^7.13.14",
    "@quasar/app": "^3.0.0",
    "@quasar/quasar-app-extension-testing": "^2.0.1",
    "@quasar/quasar-app-extension-testing-e2e-cypress": "^4.0.0-beta.8",
    "@quasar/quasar-app-extension-testing-unit-jest": "^3.0.0-alpha.7",
    "@types/faker": "^5.5.9",
    "@types/jest": "^26.0.14",
    "@types/node": "^12.20.21",
    "@typescript-eslint/eslint-plugin": "^4.16.1",
    "@typescript-eslint/parser": "^4.16.1",
    "babel-jest": "^26.5.6",
    "eslint": "^7.14.0",
    "eslint-config-prettier": "^8.1.0",
    "eslint-plugin-cypress": "^2.11.3",
    "eslint-plugin-jest": "^25.2.2",
    "eslint-plugin-vue": "^7.0.0",
    "faker": "^5.5.3",
    "jest": "^26.5.6",
    "ts-jest": "^26.5.6",
    "typescript": "^4.3.5",
    "vue-jest": "^5.0.0-alpha.10"
  },
```

freakzlike:
I think this has been recently fixed by https://github.com/vuejs/vue-test-utils-next/pull/1180 but currently not released

freakzlike:
@frisch-raphael can you test this again with rc.19?

yakobe:
I am having this problem in rc.20

freakzlike:
@yakobe You facing the same type error with `trigger` or are you calling another method? It would be good, if you provide some code snippet.

I will add some info to the docs, about the `WrapperLike` type

[...]

PR #1436, opened 2022-04-18
docs: Add type info about WrapperLike on findComponent

freakzlike:
relates #1195

I added some info about the `WrapperLike` type to the docs. I'm not sure whether the Signature section in the API Reference is understandable


lmiller1990:
Thanks. I wonder how useful a typedef like this is in the docs - it's not easy to read.

yakobe:
> Thanks. I wonder how useful a typedef like this is in the docs - it's not easy to read.

@lmiller1990 Very useful 😅. My tests suddenly started failing after an update and this really helped.

PR #1443, opened 2022-04-19
fix: findComponent with ref inside v-for

### LINKED ISSUE #1329: TypeError: Cannot read property 'devtoolsRawSetupState' of undefined
Vue Version: 3.2.31
Test-Utils: 2.0.0-rc.18

When upgrading from vue from 3.2.24 to 3.2.25 all our tests fail with the following error:
```
TypeError: Cannot read property 'devtoolsRawSetupState' of undefined
```

This is thrown when trying to call .findComponent on a mounted component using ref:
`wrapper.findComponent({ref: ".."})`

Seems that we were not alone with this issue:
https://github.com/vuejs/vue-test-utils/issues/1962

freakzlike:
fix #1329

### DIFF
--- a/src/baseWrapper.ts
+++ b/src/baseWrapper.ts
@@ -125,7 +125,13 @@ export default abstract class BaseWrapper<ElementType extends Node>
     }
 
     if (typeof selector === 'object' && 'ref' in selector) {
-      const result = currentComponent.refs[selector.ref]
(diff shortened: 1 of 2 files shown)


cexbrayat (approved):

lmiller1990 (approved):
Nice - not sure how you tracked this down, but great!

Issue #1445, opened 2022-04-21
Bug: Consecutive Mounts Clears Emitted Events Cache


BrettLargent:
**Describe the bug**
The current `attachEmitListener` implementation that runs in the `mount` function clears the entirety of the events cache after a new component is mounted. This prevents a user from mounting multiple components consecutively and checking their `emitted` arrays.

```js
var attachEmitListener = function () {
    events = {};
    // use devtools to capture this "emit"
    Vue.setDevtoolsHook(createDevTools(), {});
};
```

**To Reproduce**
With a test component like this:
`test.vue`
```html
<script setup>
import {onMounted} from "vue";
const emit = defineEmits(["component-mounted"]);
onMounted(() => {
    emit("component-mounted")
})
</script>

<template>
    <h1>Hello, World!</h1>
</template>
```
and a test spec file like this:
`test.spec.js`
```js
import { mount } from "@vue/test-utils";
import { beforeAll, describe, expect, test } from "vitest";
import TestComponent from "./test.vue";

describe("Bug Reproduction", () => {
    let wrapper1;
    beforeAll(() => {
        wrapper1 = mount(TestComponent);
    });

    test("Here's the bug", () => {
        expect(wrapper1.emitted()["component-mounted"].length).toBe(1); // Passes

        const wrapper2 = mount(TestComponent);
        expect(wrapper2.emitted()["component-mounted"].length).toBe(1); // Passes
        expect(wrapper1.emitted()["component-mounted"].length).toBe(1); // TypeError: Cannot read properties of undefined (reading 'length')
    });
});
```
You can run a test suite like vitest or jest to see that the first check of wrapper1's emitted array passes, but a subsequent check after mounting another component throws a TypeError explaining that `wrapper1.emitted()["component-mounted"]` is `undefined`.

**Expected behavior**
I expect to be able to mount any number of components without the emit history of those previously mounted components being cleared unless I deliberately want them to be cleared.

**Related information:**
- `@vue/test-utils` version: 2.0.0-rc.18
- `Vue` version: 3.2.25
- `vitest` version: 0.9.0
- `node` version: v16.13.2
- `npm` version: 8.1.2

lmiller1990:
Nice bug report. This is a pretty big oversight, we should probably generate a unique id each time we mount a component and cache the emitted events like that, something like:

```js
{
  'foo-component-1': { /* events map */ },
  'foo-component-2': { /* events map */ },
}
```

[...]

PR #1449, opened 2022-04-25
Fix/emitted events cache

BrettLargent:
Refactoring attachEmitListener function to not clear entire emitted events history on every call. A new removeEventHistory has been exposed and added to the unmount method logic so that specific emit history will be cleared on wrapper unmount. Unit testing has been added as well to show that the bug has been fixed.

### DIFF
--- a/src/emit.ts
+++ b/src/emit.ts
@@ -11,7 +11,7 @@ const enum DevtoolsHooks {
   COMPONENT_EMIT = 'component:emit'
 }
 
-let events: Events
+let events: Events = {}
 
 export function emitted<T = unknown>(
   vm: ComponentPublicInstance,
@@ -28,7 +28,6 @@ export function emitted<T = unknown>(
 }
 
 export const attachEmitListener = () => {
-  events = {}
   // use devtools to capture this "emit"
   setDevtoolsHook(createDevTools(events), {})
 }
@@ -64,4 +63,9 @@ export const recordEvent = (
 
   // Record the event message sent by the emit
   events[cid][event].push(args)
+}
+
+export const removeEventHistory = (vm: ComponentPublicInstance): void => {
+  const cid = vm.$.uid
+  delete events[cid]
 }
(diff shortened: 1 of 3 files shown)


tests/emit.spec.ts
```diff
@@ -364,4 +364,35 @@ describe('emitted', () => {
...
+              }
+            },
+            'hello world'
+          )
+      }
+    })
+
+    const wrapper1 = mount(Foo)
+    await wrapper1.trigger('click')
+    expect(wrapper1.emitted('foo')).toHaveLength(1)
```
lmiller1990, line 388: Should we also assert 

```js
expect(wrapper2.emitted('foo')).toHaveLength(0)
```

Here as well?

src/emit.ts
```diff
@@ -11,7 +11,7 @@ const enum DevtoolsHooks {
   COMPONENT_EMIT = 'component:emit'
 }
 
-let events: Events
+let events: Events = {};
```
lmiller1990, line 14: Not directly relevant but as a refactor in the future, I wonder if we should use `WeakMap` instead of a record here :thinking:

lmiller1990 (commented):
Just one question about a possible additional assertion.

cexbrayat (approved):
LGTM

The PR needs to be fixed for the linting issues.
I would also squash the commits, and mention that they fix issue #1445 in the commit message.

BrettLargent:
> LGTM
> 
> The PR needs to be fixed for the linting issues. I would also squash the commits, and mention that they fix issue #1445 in the commit message.

commits have been squashed and has a footer mention that it fixes #1445 👍

tests/emit.spec.ts
BrettLargent: We would have to wait until after wrapper2 is mounted. We can do that, but the ultimate point of the test is to ensure that subsequent mounts no longer clear the emit history for previously mounted components.

cexbrayat:
Thanks @BrettLargent !

BrettLargent:
> Thanks @BrettLargent !

No problem! Thank you all for your responsiveness and feedback. Glad we got this squashed quickly 💯

lmiller1990:
I will release this soon! Please wait a few days.

cexbrayat:
Yes it has been released. Are you using VTU v2.0.2?

BrettLargent:
🤦‍♂️ nope, I was updating the wrong dependency. Sorry for that!

Issue #1166, opened 2021-12-21
Error in tests with slots in Mocha


victorborrasdev:
Hi! I think this is more of a question than a Vue Test Utils issue really, but I was hoping any of you could give me any pointer to what to look for.

We use Karma + Mocha, and we're seeing that tests that contain slots throw the following error: `TypeError: Cannot convert undefined or null to object`. 

I have created a reproduction with both Mocha and Jest (see below) and this only happens with the former, and Jest works perfectly, that's why I'm thinking it's not VTU's fault.

The only difference I've noticed is that Mocha tests use `vue-test-utils-esm-bundler.js` and Jest uses `vue-test-utils.cjs.js`. Maybe the problem is there?

I have tried using a function as suggested by @lmiller1990 [here](https://github.com/cypress-io/cypress/issues/18219#issuecomment-950950574), but the problem with this approach is that if we pass HTML, it will be rendered as text. For instance, when passing `() => '<strong>Hello</strong>'` the result of `wrapper.html()` will be `&lt;strong&gt;Hello&lt;/strong&gt;`. This might be acceptable for many use cases, but maybe this is an unexpected behavior that needs to be addressed?

# Reproduction

This is a Vue CLI project in which I've chosen Mocha+Chai as the testing option, and then I've added Jest.

https://github.com/victorborrasdev/vue-test-utils-mocha-slots

Run Jest tests. They should pass:

```
npm run test:jest
```

Run Mocha tests. They should throw the error:

```
npm run test:jest
```

lmiller1990:
Someone looked into this and DM'd me recently, dumping the DM log here for reference when I (or someone) fixes this:

_____

- Tracked this down in VTU in a place where it trys to run a function that is compiled from Vue: https://github.com/vuejs/vue-test-utils-next/blob/b45ff43d1433e581a5d37c332474b2d7f64a5420/src/utils/compileSlots.ts#L31
- Vue seems to return different code from the compile function based on environment. I think if __BROWSER__ is false, it returns code that uses the with (_ctx) syntax that throws the error. For some reason the Cypress CT tests are loading the vue-test-utils.esm.bundler.js , which is rollup outputs if __BROWSER is false.
- When tests are run with Jest, the code returned from compile doesn’t have the with (_ctx) surrounding the code block, and __BROWSER__ is set to true in config

Something to do with our build step.

lmiller1990:
[...]

PR #1462, opened 2022-04-28
fix: string slot null ctx in renderFn

braddialpad:
Fix for #1166. Render function requires ctx as the first param. This was not being passed in so was always null and erroring out. Bit confused why this only seems to be happening for Mocha users. Seems like it would be null for everyone.


lmiller1990 (approved):
Nice. We should add a mocha + webpack test runner combo too, to verify this doesn't regress (and that it actually works) but this seems reasonable.

PR #1463, opened 2022-04-29
fix: wrapper.element with nested multiple roots

### LINKED ISSUE #1461: Bug: wrapper.text() is empty for components that only render a slot
**Describe the bug**
When a component's render function only returns the default slot, `wrapper.text()` for that component is always empty, but `wrapper.html()` contains all the correct markup.

**To Reproduce**

The first two tests here succeed, while the last test fails at `expect(wrapper.text()).toContain('some text');` even though the same assertion with `wrapper.html()` passes.

The problem is with the `Container` component in the final test, which has a render function that only returns the default slot:
```js
    const Container = defineComponent({
      render() {
        return this.$slots.default({});
      },
    });
```

The return values from `Component.render()` (first test) and `Parent.render()` (last test) are almost identical, but VTU treats them very differently.

```js
import { defineComponent, h } from 'vue';

describe('Text', () => {
  test('Component text()', () => {
      render: () => [
        h('div', 'some text'),
        h('div', 'other stuff'),
      ],
    });

    const wrapper = mount(Component);

    expect(wrapper.html()).toContain('some text');
    expect(wrapper.text()).toContain('some text'); // succeeds
  });

  test('Slot in div text()', () => {
    const ContainerWithDiv = defineComponent({
      render() {
        return h('div', undefined, this.$slots.default({}));
      },
    });

    const Parent = defineComponent({
      components: { ContainerWithDiv },
      template: `
    <ContainerWithDiv>
      <div>some text</div>
      <div>other stuff</div>
    </ContainerWithDiv>`,
    });

    const wrapper = mount(Parent);

    expect(wrapper.html()).toContain('some text');
    expect(wrapper.text()).toContain('some text'); // succeeds
  });

  test('Slot root text()', () => {
    const Container = defineComponent({
      render() {
        return this.$slots.default({});
      },
    });

    const Parent = defineComponent({
      components: { Container },
      template: `
    <Container>
      <div>some text</div>
      <div>other stuff</div>
    </Container>`,
    });

    const wrapper = mount(Parent);

    expect(wrapper.html()).toContain('some text');
    expect(wrapper.text()).toContain('some text'); // fails here
  });
});
```

**Expected behavior**

I would expect `wrapper.text()` to return something similar to `wrapper.html()` but without the markup.

**Related information:**
- `@vue/test-utils` version: 2.0.0-rc.21
- `Vue` version: 3.2.33
- `node` version: 3.2.33
- `npm` (or `yarn`) version: 8.1.0

**Additional context**

The component causing trouble for us is effectively a headless component. It takes some props, augments them or transforms them, and makes them available via slot props. We use this as the top level component for a bunch of Input components, and in the tests for Input components we like to use `wrapper.text()` to verify that certain values like labels are present "somewhere" without being too specific.

A workaround is to find an element in the wrapper using `wrapper.find` or `wrapper.get` and then test the `text()` of that, which works successfully.

freakzlike:
fix #1461 

`wrapper.element` did not return the correct element for some edge cases with multiple root nodes.
For example when using another component in template which has multiple root elements:
```vue
<template>
  <WithMultipleRoots />
</template>
```

This has led to issues with `wrapper.text()` (#1461)

I had to differ between stateful and functional components to avoid other failing tests. Maybe not all edge cases covered.

### DIFF
--- a/src/vueWrapper.ts
+++ b/src/vueWrapper.ts
@@ -70,7 +70,7 @@ export class VueWrapper<
     // Recursive check subtree for nested root elements
     // <template>
     //   <WithMultipleRoots />
-    // </template
+    // </template>
     const checkTree = (subTree: VNode): boolean => {
       // if the subtree is an array of children, we have multiple root nodes
       if (subTree.shapeFlag === ShapeFlags.ARRAY_CHILDREN) return true
--- a/src/vueWrapper.ts
+++ b/src/vueWrapper.ts
@@ -21,6 +21,7 @@ import {
   registerFactory,
   WrapperType
 } from './wrapperFactory'
+import { VNode } from '@vue/runtime-core'
 
 export class VueWrapper<
   T extends Omit<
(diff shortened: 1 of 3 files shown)


aethr:
Legend! 🥇

lmiller1990 (approved):
Nice job figuring this out!

PR #1479, opened 2022-05-04
regression: #1476 reactive prop object compare

### LINKED ISSUE #1476: Bug: DOM sometimes not being updated
**Describe the bug**
As of the test-utils version 2.0.0-rc.21, some of our tests started failing seemingly because the DOM was not being updated. For one component I managed to narrow the cause of this down to an equality check between two objects somehow no longer resolving to true, even when they both refer to the same object.
 
(Unfortunately the other components for which tests were broken are relatively old code, so I couldn't immediately verify whether those failures share a similar cause.)
 
 
**To Reproduce**
Here is a small repo in which the issue is reproduced: https://github.com/SkippedTurn/reproduction-recipe-test-utils. The line that seems to be causing the issue is [ln7 in "TestComponent.vue"](https://github.com/SkippedTurn/reproduction-recipe-test-utils/blob/master/src/TestComponent.vue#L7).

In the above project, if the v-if is replaced with `v-if="selectedField?.name === field.name"` so that the check is done on strings instead of Objects, everything does work again.
 
 
**Expected behavior**
The equality check resolves to true, and the associated tests succeed.
Note that, outside of the tests, the component works as intended with the latest version of Vue (3.2.33).
 
 
**Related information:**
- `@vue/test-utils` version: 2.0.0-rc.21
- `Vue` version: 3.2.33
- `node` version: 18.1.0
- `npm` version: 8.8.0

freakzlike:
Regression test and fix for issue #1476 

fix #1476


freakzlike:
I tried some implementation but I was not able to find a good solution which allows both cases (#192 and #1476) to work. Now I have a working solution, but it feels rather uncomfortable.

lmiller1990 (approved):
Nicely done.

PR #1484, opened 2022-05-08
fix: fix module resolution for esm file

### LINKED ISSUE #1483: Bug: Node doesn't resolve ESM correctly
**Describe the bug**

`@vue/test-utils` provides `module` field, that is resolved by bundlers, but if I import library with native node (or vitest, that used native node module resolution), then imported module will be in cjs (`/dist/vue-test-utils.cjs`). Other issue is that all files have the same extension (`js`), which is not supported by Node (For reference: https://github.com/sheremet-va/dual-packaging).

This leads to `@vue/test-utils` importing cjs version of Vue in Vitest, which fails the tests (with `--no-threads`), because it was already `required()` when defining plugin. 

**To Reproduce**

```ts
// leads to cjs.js
```
**Expected behavior**

```ts
// leads to esm.mjs
```

`@vue/test-utils` needs an `exports` field, so Node will correctly resolve paths:
```
  "exports": {
    ".": {
      "types": "./dist/index.d.ts",
      "import": "./dist/vue-test-utils.esm-bundler.mjs",
      "browser": "./dist/vue-test-utils.browser.js",
      "require": "./dist/vue-test-utils.cjs.js",
      "default": "./dist/vue-test-utils.cjs.js"
    }
  },
```
And `mjs` extension on esm files (will require twicking bundling process).

**Related information:**
- `@vue/test-utils` version: any
- `Vue` version: not relevant
- `node` version: not relevant
- `npm` (or `yarn`) version: not relevant

**Additional context**

sheremet-va:
Fixes #1483

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,16 @@
   "main": "dist/vue-test-utils.cjs.js",
   "unpkg": "dist/vue-test-utils.browser.js",
   "types": "dist/index.d.ts",
-  "module": "dist/vue-test-utils.esm-bundler.js",
+  "module": "dist/vue-test-utils.esm-bundler.mjs",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/vue-test-utils.esm-bundler.mjs",
+      "browser": "./dist/vue-test-utils.browser.js",
+      "require": "./dist/vue-test-utils.cjs.js",
+      "default": "./dist/vue-test-utils.cjs.js"
+    }
+  },
   "files": [
     "dist",
     "README.md",


sheremet-va:
Also, this should help with `calling createElement in undefined` in Vitest.

lmiller1990 (approved):
Nice, thanks.

Docs for this is anyone else (like me) was not familiar w/ the `exports` key: https://nodejs.org/api/packages.html#subpath-exports

veitbjarsch:
> Also, this should help with `calling createElement on undefined` in Vitest, because now `vite.config` will call `cjs` Vue version, but `@vue/test-utils` will call new `esm` version, when jsdom will be defined.

How do I ensure that the esm package is called when I use jsdom? I have the problem that it will load the browser package, which search for Vue in globalThis. There is a dirty workaround with
```
testEnvironmentOptions: {
    customExportConditions: ['node', 'node-addons'],
  },
```
But this can't be the soltuion.

lmiller1990:
Vite should be loading the ESM version - I don't think this is an issue in Test Utils. Can you make a minimal reproduction and share it (probably in the Vitest repo)?

veitbjarsch:
It might not relate to Vitest but to jest, because there I receive this problem since the update. I can try to create a repo where I reproduce the issue.

sheremet-va:
> It might not relate to Vitest but to jest, because there I receive this problem since the update. I can try to create a repo where I reproduce the issue.

The error happens, because Jest prioritises "browser" condition, when jsdom is enabled

veitbjarsch:
> > It might not relate to Vitest but to jest, because there I receive this problem since the update. I can try to create a repo where I reproduce the issue.
> 
> The error happens, because Jest prioritises "browser" condition, when jsdom is enabled

Yeah but this should be the case. Because if I use jsdom I wanna test the behavior like I'm in the browser. So in my opinion prioritizing the browser would be correct. But I'm not sure why we need  a browser package for vue test utils which depends on a global registered vue version.

sheremet-va:
> Because if I use jsdom I wanna test the behavior like I'm in the browser

You should not expect browser packages behave the same in Node as they do in browser. These are two different environments, so I wouldn't say that jest does anyone good by including it by default, since it's running inside Node.

lmiller1990:
This kind of thing is why I moved a lot of my tests to Cypress instead of Jest + jsdom... 

If there's action to be taken here to improve the testing experience with Test Utils here, happy to do so, but we definitely need 1) minimal repro and 2) verify with the runner(s) or environments in question that there's actually a bug.

PR #1497, opened 2022-05-16
fix(wrapper): fix wrapper.element for component with slot

xanf:
This PR fixes `wrapper.element` retrieval for components with slot :)

### DIFF
--- a/src/vueWrapper.ts
+++ b/src/vueWrapper.ts
@@ -75,19 +75,11 @@ export class VueWrapper<
       // if the subtree is an array of children, we have multiple root nodes
       if (subTree.shapeFlag === ShapeFlags.ARRAY_CHILDREN) return true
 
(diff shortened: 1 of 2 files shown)


xanf:
@freakzlike this was broken in https://github.com/vuejs/test-utils/pull/1463

It seems like there was an extra conditional added which does not makes sense to me, but maybe I'm missing something

freakzlike:
LGTM. The case where `ShapeFlags.SLOTS_CHILDREN` was set is already covered by the recursive call of `checkTree`. You can also merge the different IFs of `ShapeFlags.STATEFUL_COMPONENT` and `ShapeFlags.FUNCTIONAL_COMPONENT` into one as it is now the same code 👍

xanf:
@freakzlike :white_check_mark:

PR #1510, opened 2022-05-18
chore: bump to vue v3.2.34

cexbrayat:
This is a repro to showcase the regressions we have with vue v3.2.34 (currently beta.1)

As you can see in the build, we are running into some TS issues:
```
FAIL  tests/features/compat.spec.ts
  ● Test suite failed to run

    tests/features/compat.spec.ts:176:23 - error TS2339: Property 'foo' does not exist on type 'ComponentPublicInstance<unknown, {}, {}, {}, {}, {}, unknown, {}, false, ComponentOptionsBase<any, any, any, any, any, any, any, any, any, {}, ComponentProvideOptions>>'.

    176     expect(wrapper.vm.foo).toBe('bar')
                              ~~~

 FAIL  tests/mountingOptions/mocks.spec.ts
  ● Test suite failed to run

    tests/mountingOptions/mocks.spec.ts:50:44 - error TS2339: Property 'id' does not exist on type 'CreateComponentPublicInstance<{ [x: string & `on${string}`]: ((...args: never) => any) | undefined; } | { [x: string & `on${string}`]: ((...args: any[]) => any) | undefined; }, {}, {}, {}, {}, ComponentOptionsMixin, ... 12 more ..., ComponentProvideOptions>'.
      Property 'id' does not exist on type '{ $: ComponentInternalInstance; $data: {}; $props: { [x: string & `on${string}`]: ((...args: never) => any) | undefined; } | { [x: string & `on${string}`]: ((...args: any[]) => any) | undefined; }; ... 10 more ...; $watch(source: string | Function, cb: Function, options?: WatchOptions<...> | undefined): WatchStopHan...'.

    50           this.$router.push(`/posts/${this.id}`)
```

The PR also showcases why we need https://github.com/vuejs/core/pull/5947
and a weird TS inference problem


cexbrayat:
@pikax With your fix and mine, I was able to remove the workarounds to update to 3.2.34
But we still have the type-checking errors in tests. Any ideas?

```
 FAIL  tests/features/compat.spec.ts
  ● Test suite failed to run

    tests/features/compat.spec.ts:176:23 - error TS2339: Property 'foo' does not exist on type 'ComponentPublicInstance<unknown, {}, {}, {}, {}, {}, unknown, {}, false, ComponentOptionsBase<any, any, any, any, any, any, any, any, any, {}, ComponentProvideOptions>>'.

    176     expect(wrapper.vm.foo).toBe('bar')
                              ~~~

 FAIL  tests/mountingOptions/mocks.spec.ts
  ● Test suite failed to run

    tests/mountingOptions/mocks.spec.ts:50:44 - error TS2339: Property 'id' does not exist on type 'CreateComponentPublicInstance<{ [x: string & `on${string}`]: ((...args: never) => any) | undefined; } | { [x: string & `on${string}`]: ((...args: any[]) => any) | undefined; }, {}, {}, {}, {}, ComponentOptionsMixin, ... 12 more ..., ComponentProvideOptions>'.
      Property 'id' does not exist on type '{ $: ComponentInternalInstance; $data: {}; $props: { [x: string & `on${string}`]: ((...args: never) => any) | undefined; } | { [x: string & `on${string}`]: ((...args: any[]) => any) | undefined; }; ... 10 more ...; $watch(source: string | Function, cb: Fun
chore: bump to vue v3.2.34
ction, options?: WatchOptions<...> | undefined): WatchStopHan...'.

    50           this.$router.push(`/posts/${this.id}`)
```

src/baseWrapper.ts
```diff
@@ -115,6 +116,9 @@ export default abstract class BaseWrapper<ElementType extends Node>
   findComponent<T extends ComponentPublicInstance>(
     selector: T | FindComponentSelector
   ): VueWrapper<T>
+  // Find my CatchAll component
+  findComponent<T extends Component>(selector: T): DOMWrapper<Node>
+  findComponent<T extends Component>(selector: string): DOMWrapper<Element>
```
pikax, line 121: This is a more generic way to catch components, I've removed the other `findComponent` overloads and the tests still passed, still not sure if removing the other overloads would cause an issue

tests/mountingOptions/mocks.spec.ts
```diff
@@ -38,10 +38,10 @@ describe('mocks', () => {
         </div>
       `,
       computed: {
-        url() {
+        url(): string {
           return `/posts/${this.$route.params.id}`
         },
-        id() {
+        id(): string | string[] {
           return this.$route.params.id
```
pikax, line 45: Concerned about this change, not 100% sure what's happening here

src/baseWrapper.ts
cexbrayat: Shouldn't these methods return a `VueWrapper<T>`?

pikax: Yes 🤦

tests/mountingOptions/mocks.spec.ts
lmiller1990: `route.params[key]` is typed like this

image

What is your concern?

cexbrayat: The concern is that we shouldn't have to explicitly specify the type: the inference is somehow broken with V3.2.34

pikax: Having to annotate the return type is something is there since V2, but this is not an issue with Vue test utils, is more a concern with Vue 3.2.34 release, all good on test utils IMO

PR #1535, opened 2022-05-24
Update reusability-composition.md

dcrall:
Inject documentation incorrectly uses `global.provides` instead of `global.provide`. You can see `provide` is the correct key at the following locations:

- https://github.com/vuejs/test-utils#mounting-options
- https://github.com/vuejs/test-utils/blob/main/tests/mountingOptions/global.provides.spec.ts#L17


afontcu (approved):
Good catch 👍  also files such as `global.provides.spec.ts` could be renamed 😅

PR #1542, opened 2022-05-27
export package.json from module

snoozbuster:
See https://github.com/nodejs/modules/issues/445 for context. This package doesn't export its own package.json ever since the exports were added, which breaks some tooling I've written to help me with upgrading from VTU 1 to VTU 2 that tries to read the package.json to get the installed version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
       "browser": "./dist/vue-test-utils.browser.js",
       "require": "./dist/vue-test-utils.cjs.js",
       "default": "./dist/vue-test-utils.cjs.js"
-    }
+    },
+    "./package.json": "./package.json"
   },
   "files": [
     "dist",


package.json
```diff
@@ -13,7 +13,8 @@
       "browser": "./dist/vue-test-utils.browser.js",
       "require": "./dist/vue-test-utils.cjs.js",
       "default": "./dist/vue-test-utils.cjs.js"
-    }
+    },
+    "./package.json": "./package.json"
```
cexbrayat, line 17: That should be in the exports section no?

But anyway, I'm not sure that's really useful and that we want to do that: `package.json` is not part of the public API we want to expose.

What are you trying to achieve with this?

snoozbuster: Did I put it in the wrong place? [This comment](https://github.com/nodejs/modules/issues/445#issuecomment-608668879) in the linked issues indicates that it should be a sibling of the `"."` entry.

As for what I'm trying to achieve, like I said in the PR description - I have some tooling I'm working on which is designed to assist with an upgrade from VTU 1 to VTU 2 (and which I would like to soon publish as open-source) which wants reads the installed version of VTU in order to decide how it should act. Since VTU hasn't been explicitly exporting its version number as part of its public interface anywhere I'm aware of I was trying to get it from the `package.json`. 

[This comment](https://github.com/nodejs/modules/issues/445#issuecomment-724202967) suggests that it's not that uncommon for tooling ecosystems to read `package.json`. Indeed, it's what [`eslint-plugin-jest` does](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/detectJestVersion.ts#L28) in order to detect which deprecation warnings it should emit.

lmiller1990: My understanding is this would also go in `exports`, Cypress does something similar:

https://github.com/cypress-io/cypress/blob/develop/cli/package.json#L119-L134.

That said, I am not super clear on **why** we need to list it in `exports` - wouldn't you just do `const pkg = require('@vue/test-utils/package.json')`?

snoozbuster: that is _precisely_ what I can no longer do now that VTU 2 explicitly describes an `exports`, which explicitly excludes the `package.json`. try it yourself:


```
snooz@C02FL4M5Q05N public-dashboard % node
Welcome to Node.js v14.19.1.
Type ".help" for more information.
> require('@vue/test-utils/package.json')
Uncaught:
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './package.json' is not defined by "exports" in /SNIP/node_modules/@vue/test-utils/package.json
    at new NodeError (internal/errors.js:322:7)
    at throwExportsNotFound (internal/modules/esm/resolve.js:332:9)
    at packageExportsResolve (internal/modules/esm/resolve.js:565:3)
    at resolveExports (internal/modules/cjs/loader.js:450:36)
    at Function.Module._findPath (internal/modules/cjs/loader.js:490:31)
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:888:27)
    at Function.Module._load (internal/modules/cjs/loader.js:746:27)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:101:18)
    at REPL9:1:1 {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}
> 
```

lmiller1990: Weird! I swear this used to work... I didn't even know Node 14 supported `exports`.

I just tested, I think it should be:

```json
{
  "exports": {
    ".": {
      "types": "./dist/index.d.ts",
      "import": "./dist/vue-test-utils.esm-bundler.mjs",
      "browser": "./dist/vue-test-utils.browser.js",
      "require": "./dist/vue-test-utils.cjs.js",
      "default": "./dist/vue-test-utils.cjs.js"
    },
    "./package.json": "./package.json"
  },
```

lmiller1990: Edit: that's exactly what you did, LGTM!

lmiller1990 (approved):

package.json
snoozbuster: man, I was gonna say... am I hallucinating? I _did_ make this edit in github's UI but I triple checked it so many times 😆

PR #1548, opened 2022-05-28
fix(types): Export BaseWrapper from index.ts

snoozbuster:
BaseWrapper isn't exported like VueWrapper and DOMWrapper, even though it is exported from baseWrapper.ts. This means that any consumers who want to use the type have to import it from `@vue/test-utils/dist/baseWrapper` instead of `@vue/test-utils` like the others.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,6 @@
 import { DOMWrapper } from './domWrapper'
 import { VueWrapper } from './vueWrapper'
-import { BaseWrapper } from './baseWrapper'
+import BaseWrapper from './baseWrapper'
 import { mount, shallowMount } from './mount'
 import { MountingOptions } from './types'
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -1,5 +1,6 @@
 import { DOMWrapper } from './domWrapper'
 import { VueWrapper } from './vueWrapper'
+import { BaseWrapper } from './baseWrapper'
```
freakzlike, line 3: `BaseWrapper` is only exported with default and not named.

```suggestion
import BaseWrapper from './baseWrapper'
```

freakzlike, line 3: `BaseWrapper` is only exported with default and not named.

```suggestion
import BaseWrapper from './baseWrapper'
```

PR #1565, opened 2022-06-02
fix(compat): respect v-model in vue 2 with compat build

### LINKED ISSUE #1550: Bug: `shallowMount`ed components that use `v-model` under `@vue/compat` don't respect legacy custom `model` option
**Describe the bug**

Consider a test like the following:

```js
import { configureCompat } from 'vue'; // with @vue/compat
import { BFormInput } from 'bootstrap-vue';

configureCompat({ MODE: 2 });

const TestComponent = {
  emits: ['update'],
  components: {
    BFormInput
  },
  setup(_, { emit }) {
    const notes = ref('');
    watch(notes, () => emit('update', notes.value));

    return {
      notes
    };
  },
  template: `
    <BFormInput v-model="notes" />
  `
}

it('should emit "update" event when "notes" change', async () => {
  // arrange
  const wrapper = shallowMount(AppDetailsContractsUpdateContractOptions);

  // act
  await wrapper.findComponent(BFormInput).vm.$emit('update', 'user note');

  // assert
  expect(wrapper.emitted('update')[0][0]).toEqual('user note');
});
```

One would expect this test to pass - because of `configureCompat({ MODE: 2 })` the `v-model` usage is translated into the old syntax that BFormInput is expecting, and shallow mounted components have [historically respected `v-model`](https://github.com/vuejs/vue-test-utils/issues/965) in VTU 1. However, in this test the watcher for `notes` is never triggered and the test fails because `wrapper.emitted('update')` is undefined.

**To Reproduce**

See test code above.

**Expected behavior**

Shallow mounted components using old v-model syntax with correct configureCompat settings should use the underlying component's Vue 2 `model` option.

**Related information:**
- `@vue/test-utils` version: 2.0.0
- `Vue` version: 3.2.36
- `@vue/compat`: yes
- `node` version: 14.x
- `npm` (or `yarn`) version: 7.x

**Additional context**

This test does pass if BFormInput is not stubbed (eg, by using `mount` or with `stubs: { BFormInput }`). 

I suspect that this could be fixed by passing along `model` in [createStub](https://github.com/vuejs/test-utils/blob/aa45c60bf4c357f53e8ef2266f9d0deb36a8a85b/src/stubs.ts#L91-L92). not sure if the fix really needs to be any fancier than that.

snoozbuster:
Fixes #1550. 

I tried to write a test for this but it didn't pass. I think it's either because [vue-test-utils-compat](https://github.com/xanf/vue-test-utils-compat) is an additional step to make this work, or vue needs to be aliased to @vue/compat in the entire test suite, which I was a little hesitant to do. I saw that @vue/compat was already a dev dependency - not sure if we've tried to write tests like this in the past.

### DIFF
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -91,7 +91,7 @@ export const createStub = ({
     name: name || anonName,
     props: componentOptions.props || {},
     // fix #1550 - respect old-style v-model for shallow mounted components with @vue/compat
-    // @ts-ignore
+    // @ts-expect-error
     model: componentOptions.model,
     setup(props, { slots }) {
       return () => {
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -83,13 +83,16 @@ export const createStub = ({
   const anonName = 'anonymous-stub'
   const tag = name ? `${hyphenate(name)}-stub` : anonName
 
-  const propsDeclaration = type
-    ? unwrapLegacyVueExtendComponent(type).props || {}
+  const componentOptions = type
+    ? unwrapLegacyVueExtendComponent(type) || {}
     : {}
 
   return defineComponent({
     name: name || anonName,
-    props: propsDeclaration,
+    props: componentOptions.props || {},
+    // fix #1550 - respect old-style v-model for shallow mounted components with @vue/compat
+    // @ts-ignore
+    model: componentOptions.model,
     setup(props, { slots }) {
       return () => {
         // https://github.com/vuejs/test-utils/issues/1076


src/stubs.ts
```diff
@@ -83,13 +83,16 @@ export const createStub = ({
...
+    ? unwrapLegacyVueExtendComponent(type) || {}
     : {}
 
   return defineComponent({
     name: name || anonName,
-    props: propsDeclaration,
+    props: componentOptions.props || {},
+    // fix #1550 - respect old-style v-model for shallow mounted components with @vue/compat
+    // @ts-ignore
+    model: componentOptions.model,
```
lmiller1990, line 95: Did not know `model` was a thing - what is the TS error you get without `ts-ignore`? Ideally we don't ignore things, although maybe it's unavoidable. You could use `// @ts-expect-error` perhaps, since this is something we expect, not necessary an incorrect type, as such.

snoozbuster: `componentOptions` no longer includes the type for the `model` property, but it is there if you're using a Vue 2 component in Vue 2 mode. I have a patch locally using patch-package which basically does the same thing:

image

I can change it to expect-error if you would prefer

lmiller1990: I think that might be good - just to communicate why we have it a little better. Then we can merge!

PR #1569, opened 2022-06-03
chore(find): extend tests with deep nested multiple roots

freakzlike:
Extends some test for `findAll` with deep nested multiple root nodes. Created to test #1546. Feel free to merge or close if not necessary.


cexbrayat (approved):

lmiller1990:
Neat. In this case do you think we should do a beta for the next release @cexbrayat @freakzlike, or can we just push out 2.0.1? If so, I can do that now (today).

cexbrayat:
Let's do a release and see then!

Issue #1509, opened 2022-05-18
Bug: emitted does not include fallthrough events


KarKarbI4:
**Describe the bug**
`wrapper.emitted` does not include events, which are proxied by component to child component via context.attrs

**To Reproduce**
reproduction repo: https://github.com/KarKarbI4/vtu-emitted-fallthrough-bug

```js
// CoolInput.vue proxies BaseInput.vue
<template>
  <base-input v-bind="{ ...props, ...attrs }" />
</template>

<script>
import { defineComponent } from "vue";
import BaseInput from "./BaseInput.vue";

export default defineComponent({
  name: "CoolInput",
  components: {
    BaseInput,
  },
  setup(props, { attrs }) {
    return {
      props,
      attrs,
    };
  },
});
</script>
```

```js
// BaseInput.vue
<template>
  <input :value="props.modelValue" @input="onInput" />
</template>

<script>
import { defineComponent } from "vue";

export default defineComponent({
  name: "BaseInput",
  props: {
    modelValue: {
      type: String,
      required: true,
    },
  },
  setup(props, { emit }) {
    function onInput(event) {
      emit("update:modelValue", event.target.value);
    }

    return {
      props,
      onInput,
    };
  },
});
</script>
```

```js
// CoolInput.test.js
import CoolInput from './CoolInput.vue';
import { it, expect } from "vitest";
import { mount } from "@vue/test-utils";

it('Test CoolInput', async () => {
  const wrapper = mount(CoolInput);
  
  await wrapper.find('input').setValue('testvalue');
  
  const emitted = wrapper.emitted('update:modelValue') || [];
  expect(emitted.length).toBe(1);  // FAILS: recieved "0", expected "1"
  expect(emitted).toMatchObject([ ['testvalue'] ]);
})
```

meanwhile in application code I use
```
<CoolInput v-model="text" />
```
and text is correctly updated

**Expected behavior**
I expect wrapper.emitted to include fallthrough events

**Observed behavior**
wrapper.emitted does not include fallthrough events

**Related information:**
- `@vue/test-utils` version: 2.0.0-rc.21
- `Vue` version: 3.2.25
- `node` version: v16.14.2
- `pnpm` version: 7.0.0-rc.3

**Additional context**
Maybe bug is releated to issue https://github.com/vuejs/test-utils/issues/352

lmiller1990:
[...]

PR #1574, opened 2022-06-06
chore(emitted): Add test case for emitted pass-through events

freakzlike:
Test for #1509

This is only an issue, when the root component uses the pass-through (`v-bind="$attrs"`). Passing a function as prop `onUpdate:modelValue` does call the function but not record the emitted event. It seems that the event is not catched by  devtools here: https://github.com/vuejs/test-utils/blob/3cc95f9a34656522806ee8431ccde80b2d626676/src/emit.ts#L35-L45

### DIFF
--- a/tests/emit.spec.ts
+++ b/tests/emit.spec.ts
@@ -365,6 +365,35 @@ describe('emitted', () => {
     expect(wrapper.emitted().bar[1]).toEqual(['click'])
   })
 
+  it('captures pass through events', async () => {
+    const BaseInput = defineComponent({
+      props: {
+        modelValue: String
+      },
+      emits: ['update:modelValue'],
+      template: `<input :value="modelValue" @input="$emit('update:modelValue', $event.target.value)" />`
+    })
+
+    const CustomInput = defineComponent({
+      components: { BaseInput },
+      template: `<BaseInput v-bind="{ ...$attrs }"/>`
+    })
+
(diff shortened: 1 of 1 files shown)


freakzlike:
Unfortunally I was still not able to find a solution for this. Anyone who wants to help with this can do so 😃

cexbrayat:
I looked into it with the help of Guillaume (author of the devtools), and this is normal behavior, as there is no event emitted in that case.

So I think the way to test it would be to pass a listener in the props of `mount`:
```ts
it('captures pass through events', async () => {
    const BaseInput = defineComponent({
      props: {
        modelValue: String
      },
      emits: ['update:modelValue'],
      template: `<input :value="modelValue" @input="$emit('update:modelValue', $event.target.value)" />`
    })

    const CustomInput = defineComponent({
      components: { BaseInput },
      template: `<BaseInput />`
    })

    let emitted = ''
    const wrapper = mount(CustomInput, {
      props: {
        modelValue: 'Text',
        'onUpdate:modelValue': (e) => (emitted = e)
      }
    })

    const input = wrapper.get('input')
    await input.setValue('New Text')

    const baseInput = wrapper.findComponent(BaseInput)
    expect(baseInput.emitted('update:modelValue')).toEqual([['New Text']])
    expect(emitted).toBe('New Text')
  })
```
I'll close this and the related issue

PR #1601, opened 2022-06-15
repro: type-checking issue with script setup in v2.0.1

cexbrayat:
There is a type-checking issue with VTU v2.0.1 (not with v2.0.0) when using the latest vue-tsc and components with script setup. This is reproducible in the repo if we bump t the latest vue-tsc (which necessitates to work around https://github.com/johnsoncodehk/volar/issues/1452).

We should fix and release, as people using TS and script setup will most likely encounter the issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "vue-class-component": "^8.0.0-rc.1",
     "vue-jest": "^5.0.0-alpha.10",
     "vue-router": "^4.0.16",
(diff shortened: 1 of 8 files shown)


cexbrayat:
@pikax Did you had time to take another look at this? It would be great to unblock and cut another release with the fix. No worries if you haven't, I just wanted to make sure it's still on your radar 😉

PR #1608, opened 2022-06-17
fix: fix typo

liuzhuan:


docs/guide/essentials/forms.md
```diff
@@ -366,7 +366,7 @@ Assume we have a form that uses the Vuetify textarea:
 <template>
   <form>
     <v-textarea v-model="description" ref="description" />
-    <button @click="handleClick">Send</button>
+    <button class="submit" @click="handleClick">Send</button>
```
cexbrayat, line 369: Good catch, but it would be better to use `type="submit"`, and change the test to `wrapper.find('button[type=submit]')`

Anyway, as I was reading this example again, I think it is not a very good one. It should handle the submit event on the form instead of the click on the button, and the emitted event should have a different name than `submit`.

Could you update your PR with:

```
<template>
  <form @submit.prevent="handleSubmit">
    <v-textarea v-model="description" ref="description" />
    <button type="submit">Send</button>
  </form>
</template>
<script>
export default {
  name: 'CustomTextarea',
  data() {
    return {
      description: ''
    }
  },
  methods: {
    handleSubmit() {
      this.$emit('submitted', this.description)
    }
  }
}
</script>
```

and update the test below to use the `submitted` event instead of the `submit` one?

cexbrayat (changes requested):

docs/guide/essentials/forms.md
liuzhuan: Good idea, I'm on it.

liuzhuan: @cexbrayat The `click` event of `<button>` is changed into `submit` event of `<form>`. If not, a new container has to be created and **attachTo**-ed, according to [Native form submission](https://test-utils.vuejs.org/guide/essentials/forms.html#triggering-complex-event-listeners), which might cause distractions to `findComponent()`.

cexbrayat (approved):

cexbrayat:
Thanks @liuzhuan 👍

PR #1629, opened 2022-06-28
fix: check if component unmounted in Wrapper.exists()

Djaler:


cexbrayat (approved):
LGTM, thanks fot the PR 👍

xanf:
@Djaler thank you for catching this issue!

I've addead a bit alternate approach (just removing conditional and adding logic in child class)

@cexbrayat thank you for merging!

Djaler:
Thx for the quick approve. Any chance to release it soon, so I can use the fix in my project? 😅

cexbrayat:
@Djaler we need to fix another issue first, so the release is blocked by #1601 for now
Once that gets sorted out and merged, we'll cut a new release. It might take some time though

Issue #1512, opened 2022-05-19
Bug: doc website search bar not work


arnauddoub:
**Describe the bug**
On the site test-utils.vuejs.org when I open the search bar I get an error
`POST https://undefined-dsn.algolia.net/1/indexes/*/queries?x-algolia-agent=Algolia%20for%20JavaScript%20(4.8.5)%3B%20Browser%20(lite)%3B%20docsearch%20(3.0.0)%3B%20docsearch-react%20(3.0.0)%3B%20docsearch.js%20(3.0.0)&x-algolia-api-key=ee1b8516c9e5a5be9b6c25684eafc42f&x-algolia-application-id=undefined net::ERR_NAME_NOT_RESOLVED`

**To Reproduce**
I go to website test-utils.vuejs.org, i click on the search bar, i fill input and nothing happens.

**Expected behavior**
I want to use the search bar.

**Related information:**
- Website doc production

**Additional context**
<img width="1427" alt="Screenshot 2022-05-19 at 10 20 03" src="https://user-images.githubusercontent.com/14277978/169248601-465ddb96-83ef-4ea0-be49-076845f21072.png">

cexbrayat:
Thanks for the report, we'll look into it 👍

lmiller1990:
Algolia bug? I've not worked with the Algolia integration before.

Looks like the application id is undefined

```
x-algolia-api-key=ee1b8516c9e5a5be9b6c25684eafc42f&x-algolia-application-id=undefined
```

Not sure what's going on. I am also not sure who set up Algolia or how we can configure it, maybe @afontcu?

afontcu:
hey! yeah I set it up a while ago, and I'd say it worked as expected. Probably something went south during the site transition for vue 2 -> vue 3. I remember we needed to define some tags to split the V1 docs and the V2 ones, but now that might work different as domains are different as well.

For instance, [current config](https://github.com/vuejs/test-utils/blob/main/docs/.vitepress/config.ts#L20-L26) is probably aiming at the old setup (notice the `tags:next` filters). I recall contacting to algolia open source support team, they were really helpful to set up the appropriate config

luckylooke:
Please, searching docs is quite important feature of libraries, thanks 🙏 🙏 🙏

lmiller1990:
[...]

PR #1636, opened 2022-07-01
fix: algolia appId

okxiaoliang4:
reference: https://vuepress.vuejs.org/theme/default-theme-config.html#algolia-search

close: #1512


afontcu:
Looks like an improvement as searching through API, migrating from VTU 1, and Installation work, but cannot search content from Guide 🤔 anyway I'm merging this as it's def better than it was!

lmiller1990:
Nicely done!

PR #1640, opened 2022-07-01
chore(types): improve typing of emitted

xanf:
I find pretty annoying usage of emitted with typescript:

```ts
    expect(wrapper.emitted<unknown[]>('input')[0][0]).toEqual(testItems)
```

Requirement to specify `<unknown[]>` is pretty sad - since emitted events is always an array of arrays

This PR updates typings to simplify such use case


lmiller1990 (approved):
Nice.

This won't be able to go out until https://github.com/vuejs/test-utils/pull/1601 unfortunately.

PR #1644, opened 2022-07-04
chore: switch to pnpm

cexbrayat:
As vuejs/core, vuejs/router, vitepress and others have switched to pnpm, this PR does the same for VTU


lmiller1990 (approved):
Sure. I am not entirely sure of any obvious benefits in switch (`yarn` seems great to me - or even move back to the default `npm`?) but happy to go with this.

PR #1649, opened 2022-07-05
bugfix: updated `mocks` params level

jofaval:
---
name: Feature request
about: Update the router mock testing documentation
title: 'Unnested mock param:'
labels: ''
assignees: ''

---

When testing a mounted component, the `global` prop does not exist anymore, `mocks` can be at the same level than `props`.

To remove the global key so that `mocks` is at the same level than `props`

Specifying a change in versions with some sort of disclaimer so that newer users know where to put the key.

**Additional context**
I was trying to mock a router param and when refering to the documentation typescript suggested that global does not exist but `mocks` does, without it being nested inside `global`.

### DIFF
--- a/docs/guide/advanced/vue-router.md
+++ b/docs/guide/advanced/vue-router.md
@@ -52,11 +52,9 @@ test('allows authenticated user to edit a post', async () => {
     props: {
       isAuthenticated: true
     },
-    global: {
-      mocks: {
-        $route: mockRoute,
-        $router: mockRouter
-      }
+    mocks: {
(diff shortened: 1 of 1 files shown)


cexbrayat:
Hi @jofaval 

Thank you for opening a PR, but `mocks` is still a property of `global` and should not be used at the same level as `props`.
If TS complains about it, there is probably something weird going on. If you think this is a VTU issue, please open an issue with a minimal reproduction that shows the TS error, and we'll investigate.

PR #1729, opened 2022-08-19
chore: Add version info to bug report template

freakzlike:
Add version of to issue template to inform users about different repos for v1 and v2


cexbrayat (approved):
🙏

PR #1744, opened 2022-08-31
chore: improve type for mount with TS v4.8

cexbrayat:
I'm not 100% sure why we need these with TS v4.8, probably a change needed by https://devblogs.microsoft.com/typescript/announcing-typescript-4-8/#improved-intersection-reduction-union-compatibility-and-narrowing

TS compiler is happy with these changes


cexbrayat:
Merging this to unblock the bump to TS v4.8

PR #1759, opened 2022-09-06
Update baseWrapper.ts

Ken-vdE:
The disabled attribute can be set to 'false', which means it is not disabled. 
The current isDisabled function does not check for that.

I'm aware the spec does not allow for false values (as per https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes), 
however when used in this situation e.g. 
 ```vue
<template>
    <component :is="type" :disabled="index > 1" @click="$emit('something')" data-testid="button"/>
</template>
```

And running a test like this:
```javascript
const wrapper = mount(Component, { props: { type: 'button' } })
const button = wrapper.find('[data-testid="button"]')
await button.trigger('click');
expect(wrapper.emitted()).toHaveProperty('something')
```
Does not work because the `this.attributes().disabled` returns `'false'` instead of `undefined`...

### DIFF
--- a/src/baseWrapper.ts
+++ b/src/baseWrapper.ts
@@ -329,12 +329,9 @@ export default abstract class BaseWrapper<ElementType extends Node>
       'TEXTAREA',
       'INPUT'
     ]
-    const hasDisabledAttribute = this.attributes().disabled !== undefined
-    const elementCanBeDisabled =
-      isElement(this.element) &&
-      validTagsToBeDisabled.includes(this.element.tagName)
-
-    return hasDisabledAttribute && elementCanBeDisabled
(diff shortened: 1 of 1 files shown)


cexbrayat:
Maybe it returns false and not undefined because you have a dynamic component no?
I think we already have test cases with disabled in the code base, it should be working fine.

Can you please try to add test cases to `attributes.spec.ts` to:
- make sure your change is working as expected
- make sure  your change is really necessary

I think that would help us understand what you're trying to achieve

cexbrayat:
Yeah I suspected this was not really an issue on VTU.

happy-dom has several issues sadly, that's why we use jsdom

PR #1767, opened 2022-09-13
fix: Add props to stubs for built-in components

wolfgangwalther:
The stubs for `<transition>`, `<transition-group>` and `<teleport>` don't have any props defined, which currently makes it impossible to test for those to be set correctly.

Took the props to add from https://vuejs.org/api/built-in-components.html.

### DIFF
--- a/src/stubs.ts
+++ b/src/stubs.ts
@@ -171,9 +171,10 @@ export function stubComponents(
       return [
         createStub({
           name: 'transition',
+          type: type as any,
           renderStubDefaultSlot: true
         }),
-        undefined,
+        props,
         children
       ]
     }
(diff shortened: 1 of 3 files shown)


src/stubs.ts
```diff
@@ -171,9 +171,28 @@ export function stubComponents(
       return [
         createStub({
           name: 'transition',
+          type: {
```
freakzlike, line 174: Maybe you can pass the real transition component here? So we don't have all props defined which might change in the future

cexbrayat (approved):
LGTM 👍

freakzlike (approved):
Good work. Thank you 👍

cexbrayat:
@wolfgangwalther The build is failing. Ping me when you managed to fix it, and I'll merge the PR

cexbrayat:
I'll push an additional commit to slighly refactor. thanks @wolfgangwalther !

PR #1776, opened 2022-09-19
chore: refactor and remove old files

lmiller1990:
- reorder mount to keep types next to each other
- remove yarn.lock (we are using pnpm)


cexbrayat (commented):
I think you re-added the yarn lockfile in the first commit 😉

PR #1779, opened 2022-09-20
wip: support extending mounting options

lmiller1990:
Some libraries like Cypress and Playwright and Testing Library would like to extend Test Utils by adding some new mounting options, whilst maintaining the type safety.

Testing Library example: we need to use `any`: https://github.com/testing-library/vue-testing-library/blob/6dd48477002760262652dd047de0897762bd60c4/types/index.d.ts#L51

Eg: maybe you want to add a special `foo` mounting option.

```ts
mount(Comp, {
  foo: 'bar'
})
```

I'm not sure how to extend our types to do this. I took a stab but it's not really working.

I think the ideal API might be:

```ts
export const myMount: typeof mount<{ foo: string }> = (comp, opts) => {
  // do something with foo
  console.log('You passed foo!', opts.foo)
  return mount(comp, opts)
}
```

Basically I'd like support "wrapping" mount and extending it. Is this something we can do?

Edit: I tried something a little different but more simple, still no luck: https://github.com/vuejs/test-utils/pull/1780

Edit edit: got some tips, see this comment: https://github.com/cypress-io/cypress/issues/23653#issuecomment-1251792959

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -294,9 +294,9 @@ export function mount<
 >
 
 // implementation
-export function mount(
+export function mount<Ext = {}>(
   inputComponent: any,
-  options?: MountingOptions<any> & Record<string, any>
+  options?: MountingOptions<any> & Ext
 ): VueWrapper<any> {
   // normalize the incoming component
   const originalComponent = unwrapLegacyVueExtendComponent(inputComponent)
@@ -561,3 +561,24 @@ export function mount(
 export const shallowMount: typeof mount = (component: any, options?: any) => {
   return mount(component, { ...options, shallow: true })
 }
+
+const C = defineComponent({
+  render () {
+    return h('div')
+  } 
+})
+
+interface Foo {
+  /**
+   * it's foo
+   */
+  foo: string
+}
+
+const myMount: typeof mount<Foo> = (c: any, o?: any) => {
+  return mount(c, o)
+}
+
+myMount(C, {
+  foo: 'string'
+})


lmiller1990:
More minimal example...
```ts
type Comp<P> = (props: P) => string

interface Props {
  /**
   * foooooo
   */
  foo: string
}

const comp: Comp<Props> = (props) => {
  return `${props.foo}!`
}

interface Opts<P> {
  props: P
}

function render<E, T>(comp: Comp<T>, options?: Opts<T> & E) {
  return 'anything'
}

// Infer the first type somehow?
render(comp, {
  props: {
    foo: 'bar'    
  },
  n: 1
})
```

I want `render` to infer `foo` as a string` as well as type the additional `n` arg. But you cannot do `mount<{ n: number }>` since it expects a second arg for `T` - which I want to infer. I don't know the correct terminology but I want a kind of partial type inference :thinking:

lmiller1990:
Ok, this is almost there...

```ts
type MountFn<ExtraOptions, MountResult> = (component: any, options?: MountingOptions<any> & ExtraOptions) => MountResult

export const myMount: MountFn<{ blah: string }, VueWrapper> = (component, options) => {
  return mount(component, options)
}

const d = defineComponent({ 
  props: { 
    bar: { 
      type: String, 
      required: false 
    }, 
  },
})

myMount(d, {
  props: {},
  blah: ''
})
```

Only problem is we do `MountingOptions<any>`, so we lose the type safety for `data` and `props`.

PR #1783, opened 2022-09-24
fix: return the correct  name when stubbing a `script setup` component

joeyhuang0235:
When stubbing a component using the `<script setup>` syntax test-utils can't return its name correctly, so it skips the stub and the stub fails.There's an [issue ](https://github.com/antfu/unplugin-vue-components/issues/429) that describes the problem.
This is because `getComponentName()` function doesn't return its `__name` property.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "rollup-plugin-typescript2": "0.34.0",
     "tslib": "2.4.0",
     "typescript": "4.8.3",
+    "unplugin-vue-components": "^0.22.7",
     "vite": "3.1.3",
     "vitepress": "0.22.4",
     "vitest": "0.22.1",
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "rollup-plugin-typescript2": "0.34.0",
     "tslib": "2.4.0",
     "typescript": "4.8.3",
-    "unplugin-vue-components": "^0.22.7",
+    "unplugin-vue-components": "0.22.7",
     "vite": "3.1.3",
     "vitepress": "0.22.4",
     "vitest": "0.22.1",
--- a/src/utils/componentName.ts
+++ b/src/utils/componentName.ts
@@ -39,6 +39,9 @@ export const getComponentName = (
 ): string => {
   if (isObjectComponent(type)) {
     return (
+      // If the component we stub is a script setup component and is automatically
+      // imported by unplugin-vue-components we can only get its name through
+      // the `__name` property.
       getComponentNameInSetup(instance, type) || type.name || type.__name || ''
     )
   }
--- a/src/utils/componentName.ts
+++ b/src/utils/componentName.ts
@@ -38,7 +38,9 @@ export const getComponentName = (
   type: VNodeTypes
 ): string => {
   if (isObjectComponent(type)) {
-    return getComponentNameInSetup(instance, type) || type.name || ''
+    return (
+      getComponentNameInSetup(instance, type) || type.name || type.__name || ''
+    )
   }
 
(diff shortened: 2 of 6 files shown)


cexbrayat (changes requested):
Thanks @joeyhuang0235 

Can you add a test that checks this change please?

joeyhuang0235:
OK, I'll do it later

joeyhuang0235:
@cexbrayat If I want this test case to run expectedly I have to install `unplugin-vue-vomponents` and add it to Vitest configration file like this:

joeyhuang0235:
@cexbrayat If I want the test case to run expectedly I have to install `unplugin-vue-components` and add it to Vitest configration file like this:
image
Is it OK?

cexbrayat:
I guess we can, if that solves a specific issue for this plugin.

Is it possible to limit the application of this plugin only to the test you are adding?

joeyhuang0235:
I think I can't. As long as a component is imported, the plugin will take effect, so I can't limit the scope of this plugin to a test, but only control the target it transforms. So I specified that it can only transform the AutoImportScriptSetup component

cexbrayat:
Ok, it should be good enough. Push your test and we'll take a look 👍

package.json
```diff
@@ -52,6 +52,7 @@
     "rollup-plugin-typescript2": "0.34.0",
     "tslib": "2.4.0",
     "typescript": "4.8.3",
+    "unplugin-vue-components": "^0.22.7",
```
cexbrayat, line 55: nit: can you use an exact dependency here please?

src/utils/componentName.ts
```diff
@@ -38,7 +38,9 @@ export const getComponentName = (
   type: VNodeTypes
 ): string => {
   if (isObjectComponent(type)) {
-    return getComponentNameInSetup(instance, type) || type.name || ''
+    return (
+      getComponentNameInSetup(instance, type) || type.name || type.__name || ''
```
cexbrayat, line 42: A comment could be nice to explain why we need `__name`

vitest.config.ts
```diff
@@ -2,9 +2,18 @@ import path from 'path'
...
 import vue from '@vitejs/plugin-vue'
 import jsx from '@vitejs/plugin-vue-jsx'
+import Components from 'unplugin-vue-components/vite'
 
 export default defineConfig({
-  plugins: [vue(), jsx()],
+  plugins: [
+    vue(),
+    jsx(),
+    Components({
```
cexbrayat, line 11: same here: a short comment explaining why we have this plugin would be great

cexbrayat (changes requested):
Just a few nits, and we're good to merge 👍

package.json
joeyhuang0235: OK, I'm doing it

vitest.config.ts
joeyhuang0235: Is it okay like this?
image

src/utils/componentName.ts
joeyhuang0235: Is it OK?
image

cexbrayat: Perfect

vitest.config.ts
cexbrayat: 👍

cexbrayat:
You need to run `pnpm i` again to update the lockfile, and we should be ready to merge

joeyhuang0235:
Ok, sorry I forgot

cexbrayat:
Awesome, thanks @joeyhuang0235

PR #1792, opened 2022-10-04
Added findAllComponentsByProps() and findComponentByProps()

mrcwbr:
To improved the efficiency of getting a specific instance of a component, I introduces to functions of the `wrapper` class:
* findAllComponentsByProps()
* findComponentByProps()

therefore it's pretty easy to get access a specific component in a test. 
Example
```vue
<MyButton title="save" />
<MyButton title="cancel" />
```

Now I'm able to get the cancel button in tests easy by calling:
```js
const cancelButton = wrapper.findComponentByProps(MyButton, {title: 'cancel'})
```

inspired by `react test renderer`'s `testInstance.findByProps()` https://reactjs.org/docs/test-renderer.html#testinstancefindbyprops


cexbrayat:
@mrcwbr Thanks for the PR

I'm not sure this is really worth adding and maintaining though.

@lmiller1990 @freakzlike @xanf What do you think?

xanf:
I'm not sure this should make it to the core. Considering we have working plugins system - it feels like a perfect case for plugin

freakzlike:
As @xanf pointed out: That's exactly the use case for plugins as descriped in docs: https://test-utils.vuejs.org/guide/extending-vtu/plugins.html#writing-a-plugin

mrcwbr:
> I think we all agree.
> 
> Thanks @mrcwbr and let us know if you publish it as a plugin!

@cexbrayat as you mentioned I published a plugin to npmjs https://github.com/mrcwbr/vue-test-utils-plugin-find-by-props

PR #1793, opened 2022-10-04
fix: cleanup event listeners on unmount

### LINKED ISSUE #1787: Bug:
**Describe the bug**

I have a project with 8,000+ tests using mocha, many of which create a real Vuex store.  Recently we have migrated to Vue3 and test-utils v2.  After the upgrade a memory leak is causing the test suite to fail as it runs out of heap space.  At this time it seems to be adding a plugin (store) when mounting causes the issue, though not fully sure if there is additional cleanup I must be doing that isn't mentioned in docs.

**To Reproduce**

I have created the HelloWorld project from vue-cli (webpack, vuex, mocha), and modified the one and only test in that project:
1. Looped single test many times
1. created a store and added to `global.plugins` for each test
   1.  I used [this example ](https://test-utils.vuejs.org/guide/advanced/vuex.html#testing-with-a-real-vuex-store )to create/add the store

```
describe('HelloWorld.vue', () => {
  for (let i = 0; i < 10000; i++) {
    it(`${i}`, () => {
      const store = createStore({
        state () {
          return {
            count: 0
          }
        },
        mutations: {
          increment (state) {
            state.count += 1
          }
        }
      })

      const msg = 'new message'
      const wrapper = shallowMount(HelloWorld, {
        props: { msg },
        global: {
          plugins: [store]
        }
      })
      expect(wrapper.text()).to.include(msg)
    })
  }
})
```
When I inspect this while running I can observe each test is leaking memory.  Simply, not creating `store` and not pass in `plugins`, I no longer see the leak.

**Expected behavior**

Run entire test suite without getting out of heap exception.

**Related information:**
- `@vue/test-utils` version: 2.1.0 
- `Vue` version: 3.2.39
- `node` version: 14.18.1 and 16.17.1
- `vuex` version 4.0.2
- `@vue/cli-plugin-*` version ~4.5.19

**Additional context**

A few other things I have tested:
- creating one store for all tests and setting with `config.global.plugins = [store]`, still saw leak
- created `store` but did not actually use in `config.global.plugins` or in `shallowMount`, did not see leak

freakzlike:
fix #1787

This is only a first implementation. There might be some more to optimize

### DIFF
--- a/src/vueWrapper.ts
+++ b/src/vueWrapper.ts
@@ -35,6 +35,7 @@ export class VueWrapper<
   private rootVM: ComponentPublicInstance | undefined | null
   private __app: App | null
   private __setProps: ((props: Record<string, unknown>) => void) | undefined
+  private cleanUpCallbacks: Array<() => void> = []
 
   constructor(
     app: App | null,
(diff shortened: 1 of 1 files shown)


cexbrayat (approved):
Awesome, thanks for taking a look at this issue

PR #1797, opened 2022-10-06
fix(config): Do not use config.renderStubDefaultSlot

xanf:
Right now I've spent 30 minutes figuring why my default slots in stubs were not rendered, just to figure out, that our docs mention `global.renderStubDefaultSlot` as the way to control stub behavior, however we have **both** `config.renderStubDefaultSlot` and `config.global.renderStubDefaultSlot` with first being a preference

While **technically** this is a breaking change, I find an ability to define same thing in two places extremely confusing and fix for affected codebases will be trivial (swap `config.renderStubDefaultSlot` with `config.global.renderStubDefaultSlot`)

Otherwise, if we decide that we do not want to have such breaking change, I suggest swapping priority, so `config.global.renderStubDefaultSlot` will be taken into account before `config.renderStubDefaultSlot` and mark `config.renderStubDefaultSlot` as deprecated

### DIFF
--- a/docs/guide/advanced/stubs-shallow-mount.md
+++ b/docs/guide/advanced/stubs-shallow-mount.md
@@ -275,11 +275,11 @@ For this use case, you can use `config.renderStubDefaultSlot`, which will render
 import { config, mount } from '@vue/test-utils'
 
 beforeAll(() => {
-  config.renderStubDefaultSlot = true
+  config.global.renderStubDefaultSlot = true
 })
 
 afterAll(() => {
-  config.renderStubDefaultSlot = false
+  config.global.renderStubDefaultSlot = false
 })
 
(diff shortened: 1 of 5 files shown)


lmiller1990 (approved):
I think this is fine - I don't think that many people are using this, except people who know VTU really well. Also, we can just make it a minor and tell them they can set `config.global.renderStubDefaultSlot = true` if they needed the previous behavior. WDYT?

xanf:
@lmiller1990 @cexbrayat I've updated PR with your suggestions:

* `config.renderStubDefaultSlot` is marked as deprecated in typescript
* We will trigger console warning when `config.renderStubDefaultSlot` is set to `true` even if user is not using TypeScript

cexbrayat (approved):
👍

PR #1811, opened 2022-10-12
chore: use @rollup/plugin-typescript

cexbrayat:
Replaces `rollup-plugin-typescript2` by the official core plugin for TS, which supports Rollup v3.

To make sure we don't have a regression, I made a repo containing the result of the build on main (formatted with prettier): https://github.com/cexbrayat/test-utils-dist

Then I made a PR with the result of the build using `@rollup/plugin-typescript`: https://github.com/cexbrayat/test-utils-dist/pull/1/files

As you can see the result is pretty much the same.


cexbrayat:
I'm going to go ahead and merge it, to unblock the renovate PR

PR #1812, opened 2022-10-12
chore(scripts): add prepare script

xanf:
This is extremely simple, yet useful (in my opinion) change.

> NOTE: If a package being installed through git contains a prepare script, its dependencies and devDependencies will be installed, and the prepare script will be run, before the package is packaged and installed.

Since npm v5 npm executes `prepare` script when dependency is installed from git. By including `prepare` script it will allow our users use `@vue/test-utils` from `main` (or any other branch) even if release was not published to npm yet

However, this has a downside - after doing install locally (without include `@vue/test-utils` as dependency) it will be also executed. However, I feel here benefits outweight the downside

WDYT?

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     "test:build": "vitest --mode test-build",
     "tsd": "tsc -p test-dts/tsconfig.tsd.json",
     "build": "rollup -c rollup.config.js",
+    "prepare": "rollup -c rollup.config.js",
     "lint": "eslint --ext .ts src/ tests/",
     "lint:fix": "pnpm run lint --fix",
     "docs:dev": "vitepress dev docs",


package.json
```diff
@@ -76,6 +76,7 @@
     "test:build": "vitest --mode test-build",
     "tsd": "tsc -p test-dts/tsconfig.tsd.json",
     "build": "rollup -c rollup.config.js",
+    "prepare": "rollup -c rollup.config.js",
```
xanf, line 79: We're duplicating `build` script instead of `pnpm run build` - we do not want to rely on `pnpm`, and having `build` invoked as `pnpm run prepare` feels very wrong to me

cexbrayat:
I don't have a strong opinion about this one, I don't think I ever used `prepare`.

@lmiller1990 @freakzlike WDYT?

xanf:
> I don't have a strong opinion about this one, I don't think I ever used `prepare`.

The great thing is it used automatically and our users will be able to use main branch if needed

freakzlike:
I think this is good. Not sure if that will affect the release process

xanf:
@freakzlike not at all. We have extra build triggered inside pipeline (you can see even in this PR that `prepare` was called during `npm install`) but it doesn't affect released package at all

cexbrayat:
Let's go ahead and try it then

PR #1819, opened 2022-10-15
fix(stubs): Do not render function body in stub

xanf:
`@vue/test-utils` v1 had special handing for functions in props when rendering stubs. 

This PR makes `@vue/test-utils` v2 behave in the same way, even reusing same placeholder from https://github.com/vuejs/vue-test-utils/blob/97fdf18cd96f2ce19036d497ec98f2f76fc7b066/packages/create-instance/create-component-stubs.js#L21

(this allows us to run snaphot tests of vue2 and `@vue/compat` and to verify snapshots identity)

### DIFF
--- a/src/vnodeTransformers/stubComponentsTransformer.ts
+++ b/src/vnodeTransformers/stubComponentsTransformer.ts
@@ -63,12 +63,15 @@ const shouldNotStub = (type: ConcreteComponent) => doNotStubComponents.has(type)
 export const addToDoNotStubComponents = (type: ConcreteComponent) =>
   doNotStubComponents.add(type)
 
-const stringifySymbols = (props: ComponentPropsOptions) => {
+const normalizeStubProps = (props: ComponentPropsOptions) => {
   // props are always normalized to object syntax
   const $props = props as unknown as ComponentObjectPropsOptions
   return Object.keys($props).reduce((acc, key) => {
     if (typeof $props[key] === 'symbol') {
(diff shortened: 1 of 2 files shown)


freakzlike:
I think this will break using a function with `.props()`.

```ts
const someFunction = wrapper.props('someFunction')
someFunction()
```

Maybe we can implement this logic into the `wrapper.html()` method?

xanf:
> I think this will break using a function with `.props()`.

@freakzlike  I don't think so

image

<details>
<summary>Code for copy-paste</summary>

```js
  it('does not have problems on stubbing functions', () => {
    const ChildCmp = { template: '<div>child</div>', props: ['fn'] }
    const Root = {
      components: {
        ChildCmp
      },
      props: ['fn'],
      template: '<child-cmp :fn="fn">'
    }

    const fn = vi.fn()
    const wrapper = shallowMount(Root, { props: { fn } })
    expect(wrapper.findComponent(ChildCmp).props('fn')).toBe(fn)
    expect(wrapper.findComponent(ChildCmp).html()).toBe(
      '<child-cmp-stub fn="[Function]"></child-cmp-stub>'
    )
  })
```

</details>

When we're retrieving `props` - we are getting them on stub component. It get's replaced only for rendering child dom element

cexbrayat (approved):

freakzlike:
Well, then LGTM 👍

PR #1825, opened 2022-10-20
feat(setValue): allow array for multiselect .setValue() method

### LINKED ISSUE #1768: Support multiple values in wrapper<select[multiple]>.setValue
I want to simulate a user selection on a multi-select element (a `select` element which have the `mulitple` attribute).

```html
<select multiple @change="doSomething">
    <option value="1">value 1</option>
    <option value="2">value 2</option>
    <option value="3">value 3</option>
</select>
```

Currently, if I do this in my test:
`await wrapper.find('select').setValue('1');`

My `doSomething` function receives an event, whose `target.selectedOptions` contains the selection of a single element.

But when I do this: 
`await wrapper.find('select').setValue(['1', '2']);`

My `doSomething` function receives an event, whose `target.selectedOptions` is **an empty array**.

It's like `setValue` accepts an array as argument, but does not process it properly. I found no way to simulate a multiple selection.

**EDIT**:
Checking [the source code](https://github.com/vuejs/test-utils/blob/41f77b347bc600d9bce10e0a23f06908706957a5/src/domWrapper.ts#L96), the multiple select is not handled at all:
image

It has been done in test-utils v1 though : https://github.com/vuejs/vue-test-utils/pull/1554/files 

I'd like that, when I gave several values to `setValue`, all the matching options are selected inside the `select` element.

Looking at what has been done in test-utils v1, I created this Typescript utility function that does the trick:
```typescript
function setSelected(...values: string[]): Promise<void> {
  const select = wrapper.find('select');
  Array.from((select.element as HTMLSelectElement).options).forEach((option: HTMLOptionElement) => {
    option.selected = values.includes(option.value);
  });
  return select.trigger('change');
}
```

I can use it like this: `await setSelected('1', '2');` and my `doSomething` function receives the right selected options.

freakzlike:
Most taken from https://github.com/vuejs/vue-test-utils/pull/1554

closes #1768

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -1633,6 +1633,12 @@ setValue(value: unknown, prop?: string): Promise<void>
 
   <input type="checkbox" v-model="checked" />
   <div v-if="checked">The input has been checked!</div>
+
+  <select v-model="multiselectValue" multiple>
+    <option value="value1"></option>
+    <option value="value2"></option>
+    <option value="value3"></option>
+  </select>
 </template>
 
 <script>
@@ -1640,7 +1646,8 @@ export default {
   data() {
     return {
       text: '',
-      checked: false
+      checked: false,
+      multiselectValue: []
     }
   }
 }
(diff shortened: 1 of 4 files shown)


cexbrayat (approved):
👍

PR #1833, opened 2022-10-26
fix(stubs): teleport stub children as a function

### LINKED ISSUE #1829: Bug: [Vue warn] Non-function value encountered for default slot on stubbed teleport
**Describe the bug**
After update from 2.1.0 to 2.2.0, tests with `stubs: { teleport: true }` produce a warning:
```
[Vue warn]: Non-function value encountered for default slot. Prefer function slots for better performance.
  at <Teleport to="body" >
  at <Anonymous ref="VTU_COMPONENT" >
  at <VTUROOT>         
```

**To Reproduce**
Able to reproduce within VTU unit test
https://github.com/vuejs/test-utils/blob/6371349bd97e010169ed773b24a222ac3dc584e1/tests/mountingOptions/global.stubs.spec.ts#L514-L531

**Expected behavior**
Not print a warning

**Related information:**
- `@vue/test-utils` version: 2.2.0 
- `Vue` version: 3.2.41
- `node` version: 16.18.0
- `npm` (or `yarn`) version: pnpm@7.14.0

**Additional context**
I think this might come from the stubs refactoring and some previous special handling on the teleport child nodes, where they have been explicit passed as function:
https://github.com/vuejs/test-utils/blob/30e9c0c90ccfa5c7a90340ee29efce4935ed0d4d/src/stubs.ts#L208

cexbrayat:
Fixes #1829

I'm not super convinced by this fix, but it indeed looks like @freakzlike intuition is right: we need to return the children as a function for Teleport to avoid the warning.

Maybe @xanf you have a better idea on how to handle that?

### DIFF
--- a/src/vnodeTransformers/util.ts
+++ b/src/vnodeTransformers/util.ts
@@ -20,6 +20,8 @@ export type VTUVNodeTypeTransformer = (
   instance: InstanceArgsType
 ) => VNodeTransformerInputComponentType
 
+const isTeleport = (type: any): boolean => type.__isTeleport
+
 export const createVNodeTransformer = ({
   transformers
 }: {
@@ -31,15 +33,20 @@ export const createVNodeTransformer = ({
   > = new WeakMap()
 
   return (args: VNodeTransformerArgsType, instance: InstanceArgsType) => {
-    const [originalType, ...restVNodeArgs] = args
+    const [originalType, props, children, ...restVNodeArgs] = args
 
     if (!isComponent(originalType)) {
-      return [originalType, ...restVNodeArgs]
+      return [originalType, props, children, ...restVNodeArgs]
     }
 
     const cachedTransformation = transformationCache.get(originalType)
     if (cachedTransformation) {
-      return [cachedTransformation, ...restVNodeArgs]
+      // https://github.com/vuejs/test-utils/issues/1829
+      // Teleport should return child nodes as a function
+      if (isTeleport(originalType)) {
+        return [cachedTransformation, props, () => children, ...restVNodeArgs]
+      }
+      return [cachedTransformation, props, children, ...restVNodeArgs]
     }
 
     const componentType: VNodeTransformerInputComponentType = originalType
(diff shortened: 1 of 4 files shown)


src/vnodeTransformers/util.ts
```diff
@@ -53,8 +53,12 @@ export const createVNodeTransformer = ({
       transformationCache.set(originalType, transformedType)
 
       registerStub({ source: originalType, stub: transformedType })
+      // https://github.com/vuejs/test-utils/issues/1829
+      // Teleport should return child nodes as a function
+      if (transformedType.name === 'teleport') {
```
freakzlike, line 58: It might be better to check for the `Teleport` component instead of the name
```suggestion
      if (originalType === Teleport) {
```

src/vnodeTransformers/util.ts
```diff
@@ -31,15 +31,15 @@ export const createVNodeTransformer = ({
...
 
     if (!isComponent(originalType)) {
-      return [originalType, ...restVNodeArgs]
+      return [originalType, props, children, ...restVNodeArgs]
     }
 
     const cachedTransformation = transformationCache.get(originalType)
     if (cachedTransformation) {
-      return [cachedTransformation, ...restVNodeArgs]
+      return [cachedTransformation, props, children, ...restVNodeArgs]
```
freakzlike, line 42: Do we need to make this also as function for teleport?

freakzlike (commented):
I have no better solution for this.
Can you add a spy on `console.warn` on one of the teleport tests?

src/vnodeTransformers/util.ts
cexbrayat: 👍 but sadly this does not compile: I ended up borrowing the check from core: `const isTeleport = (type: any): boolean => type.__isTeleport`

cexbrayat:
@freakzlike I added a test to check that the warning is not logged anymore, and added a check in other tests when it _is_ logged, but when that we expect it.

Maybe we should do the same for the warnings logged by vue compact (either mock them or fix them if needed) to avoid polluting the test logs, and missing a regression like this one in the future

Anyway, this should fix the issue, unless someone comes up with a better idea. I'll let you give a last check and merge

lmiller1990 (approved):

PR #1834, opened 2022-10-26
chore: update to js-beautify v1.14.7

cexbrayat:
The following commit in js-beautify introduced a breaking change for us, as custom element are now treated as inline elements https://github.com/beautify-web/js-beautify/commit/aa82eb9ce664abadbcb454fd7a3a4d8de53dddec

This means a bunch of our tests, and snapshots of our users, break with this new version.

What do you think we should do?
- lock the version of js-beautify for ever to v1.14.6
- bump to v1.1.4.7, update our tests, and warn our users about the breaking change
- talk about this with the js-beautify team and see if that should be configurable maybe?

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "eslint-config-prettier": "8.5.0",
     "eslint-plugin-prettier": "4.2.1",
     "husky": "8.0.1",
-    "js-beautify": "1.14.6",
+    "js-beautify": "1.14.7",
     "jsdom": "20.0.1",
     "jsdom-global": "3.0.2",
     "lint-staged": "13.0.3",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -22,7 +22,7 @@ specifiers:
   eslint-config-prettier: 8.5.0
   eslint-plugin-prettier: 4.2.1
   husky: 8.0.1
-  js-beautify: 1.14.6
+  js-beautify: 1.14.7
   jsdom: 20.0.1
   jsdom-global: 3.0.2
   lint-staged: 13.0.3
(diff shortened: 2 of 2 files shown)


freakzlike:
I think we should talk with them to make it configurable. As it seems that all stubbed components are treat as custom elements

cexbrayat:
@freakzlike I started a conversation https://github.com/beautify-web/js-beautify/issues/2113

lmiller1990:
Hi all, sorry about the slow replies, was on leave for a while and not coding much.

I agree we should try make it configurable. Snapshots are a bit unreliable (since implementation details like this can break them) but I'd really prefer to avoid breaking changes where possible. Looks like we've already got some movement in the JS Beautify repo, thanks @cexbrayat.

lmiller1990:
For now, should we just lock `v1.14.6` and close this? The PR you made in beautify-web doesn't seem to have much movement.

cexbrayat:
@lmiller1990 We're locked on v1.14.6 (and I disabled Renovate for this package). I'll keep the PR around to have a reminder of this ongoing task, and hopefully we'll have feedback from the js-beautifier team soon (if not, then we'll close and stay on v1.14.6 forever...)

JounQin:
Just curious, why `js-beautify` was chosen instead of `prettier`?

lmiller1990:
> Just curious, why `js-beautify` was chosen instead of `prettier`?

We use it to format the output of `html()`. It's much smaller - prettier is a big dependency (it's like 15mb) which we don't need for what we are doing.

nathanaelytj:
Hello, is this advisory related to this PR and issue?
https://github.com/advisories/GHSA-c2qf-rxjj-qqgw

Thank you

cexbrayat:
@nathanaelytj Not directly. We can't update to the latest js-beautify version because of a breaking change.
Regarding the vulnerability, if you only use js-beautify via test-utils, then you have nothing to worry about.

cexbrayat:
Superseeded by #2131

PR #1839, opened 2022-10-27
docs(v-model): add v-model section

pikax:
Simple how to handle components with `v-model`

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -86,6 +86,7 @@ export default defineConfig({
             text: 'Reusability and Composition',
             link: '/guide/advanced/reusability-composition'
           },
+          { text: 'Testing v-model', link: '/guide/advanced/v-model' },
           { text: 'Testing Vuex', link: '/guide/advanced/vuex' },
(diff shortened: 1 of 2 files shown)


cexbrayat (approved):
Thanks for taking the time to write this ♥️

lmiller1990 (approved):

pikax:
> Replace the value and @input props in Editor with v-model and it won't work

That's because you must call the correct event on model change, if you use the [defineModel](https://blog.vuejs.org/posts/vue-3-3#definemodel) or [useVModel](https://vueuse.org/core/useVModel/) it will work

nelisbijl:
That makes no difference. Problem is not the way the custom modelValue is defined, problem is the use of Vue's built-in v-model on the <input tag.
Probably related to this warning:
[Vue warn]: withDirectives can only be used inside render functions.
No idea, how to resolve that (if at all possible)

pikax:
@nelisbijl if you find a bug, please report on the correct project issue page not on a year's old merged PR. 

For questions please use forums, discord, etc.

PR #1844, opened 2022-11-03
chore(mount): simplify mount operation

xanf:
Found this approach in main `@vuejs/core` repo

https://github.com/vuejs/core/blob/main/packages/runtime-core/__tests__/rendererTemplateRef.spec.ts#L23

I think this might be a good approach to simplify our codebase too!

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -25,7 +25,8 @@ import {
   ComponentOptions,
   ConcreteComponent,
   Prop,
-  transformVNodeArgs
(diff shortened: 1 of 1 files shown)


cexbrayat (approved):
LGTM, just an unused import to remove to make CI happy

xanf:
@cexbrayat that was fast, I've even was unable to push another version (full one), since my internet connection got flaky (I'm commuting between two cities via train :train: )

Now this one is a final state :)

PR #1853, opened 2022-11-09
chore: update to vue v3.2.43

cexbrayat:


cexbrayat:
This is currently failing because this commit in core  https://github.com/vuejs/core/commit/506a42a9a5782b6cd297a8f4b9679db51b7aacff is breaking our type-checking.

Our codebase includes a `testing.d.ts` file that defines:
```ts
import type { Router } from 'vue-router'

declare module '@vue/runtime-core' {
  interface ComponentCustomProperties {
    $router: Router
  }
}
```

As the change in core makes the type of `globalProperties` `ComponentCustomProperties & Record<string, any>` (when it used to be only `Record<string, any>`), we now have type-checking errors for example in `config.spec.ts` where a test uses:
```ts
config.global.config.globalProperties = {
  myProp: 1
}
```

which produces:
```
tests/config.spec.ts:27:7 - error TS2322: Type '{ myProp: number; }' is not assignable to type 'ComponentCustomProperties & Record<string, any>'.
  Type '{ myProp: number; }' is missing the following properties from type 'ComponentCustomProperties': $route, $router

27       config.global.config.globalProperties = {
```

I'm not sure how this should be fixed. @pikax maybe you have an idea?

pikax:
The issue there is the return type of [mergeAppConfig](https://github.com/vuejs/test-utils/blob/main/src/utils.ts#L28) is not correct, because both `configGlobalConfig` and `mountGlobalConfig` are **optional**, but we are returning the `Required` version of it, meaning most likely we don't have all the correct types in as expected. 

I would say this error can be "ignored" by adding
```ts
return {
  ...configGlobalConfig,
  ...mountGlobalConfig,
  globalProperties: {
    ...configGlobalConfig?.globalProperties,
    ...mountGlobalConfig?.globalProperties
  } as Required<GlobalMountOptions>['config']['globalProperties']
}
``` 

Just to have the `globalProperties` with the expected type

PR #1856, opened 2022-11-11
fix(emit): force event handling even with fake timers

### LINKED ISSUE #1854: Bug: event handlers not invoked when using `jest.useFakeTimers` or `vi.useFakeTimers`
**Describe the bug**
We are experiencing an issue in our unit and integration tests that utilise `useFakeTimers` when using:
- `vue@3.2.41` on `jest@27.5.1`
- `vue@3.2.40` on `vitest@0.25.1`

Components with event bindings on outer elements are not having their handlers invoked when using `useFakeTimers` in jest or vitest.

**To Reproduce**
https://codesandbox.io/s/vue3-vue-test-utils-vitest-5yex3q?file=/src/useFakeTimers.test.js

**Expected behavior**
We expect all relevant event handlers to be invoked when `userFakeTimers` is used.

**Related information:**
- `@vue/test-utils` version: 2.2.1
- `Vue` version: 3.2.40+
- `vitest` version: 0.25.1

Related issue in `vitest`: https://github.com/vitest-dev/vitest/issues/649

**Additional context**
This isn't actually a bug in VTU, it's a side-effect of a recent edge-case change to `vue`, visible here: https://github.com/vuejs/core/blob/5ee40532a63e0b792e0c1eccf3cf68546a4e23e9/packages/runtime-dom/src/modules/events.ts#L100-L104

Vue is trying to prevent an edge case where an inner event handler causes a state change that attaches new event handlers to outer elements, where the new event handlers would catch the original event as it bubbles outwards. This code won't invoke any event handlers that were attached in the same tick when the event was created.

Using `useFakeTimers` mocks the `Date` object, and causes all `Date.now()` to return the same value, which means that handlers and events all get the same timestamp, so the check in the code linked above prevents event handlers from firing.

A simple workaround is to advance the timers after the component is mounted, but before the event is invoked (see the last unit test in the codesandbox).

I've created a simple plugin that will advance timers by 1ms before triggering any event if `useFakeTimers` is in use:
```js
const triggerWithFakeTimers = (wrapper) => {
  const vtuTrigger = wrapper.trigger;

  return {
    trigger: async (...args) => {
      if (vi && vi._timers._fakingTime) {
        vi.advanceTimersByTime(1);
      }
      return vtuTrigger.bind(wrapper)(...args);
    },
  };
};

config.plugins.DOMWrapper.install(triggerWithFakeTimers)
```

cexbrayat:
Fixes #1854

I'm not 100% convinced this is the proper way to fix it, but I don't have a better idea at the moment.
Setting `_vts` to Date.now() + 1 tricks Vue into thinking it has to handle the event even if `Date.now()` is faked and always returns the same value https://github.com/vuejs/core/blob/5ee40532a63e0b792e0c1eccf3cf68546a4e23e9/packages/runtime-dom/src/modules/events.ts#L100-L104

I hope this doesn't introduce other subtle issues. It would be great to only do this if the test is running with a fake Date, but I don't know how we can find out reliably for jest or vitest or even a manually mocked Date object.

### DIFF
--- a/tests/trigger.spec.ts
+++ b/tests/trigger.spec.ts
@@ -364,4 +364,31 @@ describe('trigger', () => {
 
     expect(wrapper.emitted().enter).toHaveLength(1)
   })
+
+  // https://github.com/vuejs/test-utils/issues/1854
+  it('dispatches events even with fakeTimers', async () => {
+    const handlerSpy = vi.fn()
+
+    const Component = defineComponent({
+      setup() {
+        return { handlerSpy }
+      },
+      template: `
+      <div>
+        <a @click="handlerSpy">
+          <span @click="() => {}">Ok</span>
+        </a>
+      </div>
+      `
+    })
+
+    vi.useFakeTimers()
+    const wrapper = mount(Component)
+
+    expect(handlerSpy).not.toHaveBeenCalled()
+
+    await wrapper.get('span').trigger('click')
+
+    expect(handlerSpy).toHaveBeenCalled()
+  })
 })
--- a/src/baseWrapper.ts
+++ b/src/baseWrapper.ts
@@ -369,6 +369,13 @@ export default abstract class BaseWrapper<ElementType extends Node>
 
     if (this.element && !this.isDisabled()) {
       const event = createDOMEvent(eventString, options)
+      // see https://github.com/vuejs/test-utils/issues/1854
+      // fakeTimers provoke an issue as Date.now() always return the same value
+      // and Vue relies on it to determine if the handler should be invoked
+      // see https://github.com/vuejs/core/blob/5ee40532a63e0b792e0c1eccf3cf68546a4e23e9/packages/runtime-dom/src/modules/events.ts#L100-L104
+      // we workaround this issue by manually setting _vts to Date.now() + 1
+      // thus making sure the event handler is invoked
+      event._vts = Date.now() + 1
       this.element.dispatchEvent(event)
     }
 


tests/trigger.spec.ts
```diff
@@ -364,4 +364,33 @@ describe('trigger', () => {
...
+      `
+    })
+
+    vi.useFakeTimers()
+    const wrapper = mount(Component)
+
+    expect(handlerSpy).not.toHaveBeenCalled()
+
+    wrapper.setProps({ label: 'foo' })
+
```
aethr, line 391: This line with `wrapper.setProps` isn't needed, I forgot to remove it from an earlier version when I was tinkering with the repro.

aethr (approved):
I actually think this is a clever solution that should prevent others from ever having to worry about the fake timers issue, and everything will "just work" as expected. This was very similar to my first fix in my project's codebase when I discovered the issue, but I had to add it to every event using the second arg to `trigger`.

Although it's tied to behavior in vue internals that could change, I think that's fine since this is the official testing library for vue. The comment does a good job of explaining what the workaround is for, and if vue changes implementation it should be easy to find the code that needs to be understood to fix it if it breaks.

Nice work!

freakzlike (approved):
Looks ok to me

lmiller1990 (commented):
Wow, what a hack. I don't know of a better way to do this but this feels pretty dirty. 🤷 If it works, it works, nice fix.

PR #1858, opened 2022-11-11
refactor: use a proxy for wrapper.vm

cexbrayat:
This fixes the issues with Vue v3.2.45 (see https://github.com/vuejs/core/issues/7103)


freakzlike (approved):
I think this is a good solution. Well done 👍

PR #1861, opened 2022-11-14
repro: "'set' on proxy: trap returned falsish for property" error with Vue v3.2.45

cexbrayat:
This comes from the fact that mocks are set via a mixin (Options API) and we run into the more strict behavior of Vue v3.2.45 introduced in https://github.com/vuejs/core/commit/f73925d76a76ee259749b8b48cb68895f539a00f#diff-ea4d1ddabb7e22e17e80ada458eef70679af4005df2a1a6b73418fec897603ceR404

First spotted by @Akryum and reported on internal discord

The test succeeds with Vue v3.2.44 and not with Vue v3.2.45


cexbrayat:
This one is trickier.

Until now, we used a mixin with a beforeCreate to populate the mocks (the code is probably inherited from VTU v1). But the recent changes in Vue makes such a modification to a script setup component impossible if they come from a component written with the Options API. See https://github.com/vuejs/core/commit/f73925d76a76ee259749b8b48cb68895f539a00f#diff-ea4d1ddabb7e22e17e80ada458eef70679af4005df2a1a6b73418fec897603ceR404

I have not found a way to make it work in all cases. If someone thinks about a potential fix, feel free to take a stab.

cexbrayat:
I now have a potential fix, by still using a mixin, but updating `setupState` directly in case we are in a script setup component. This hopefully does the trick.

lmiller1990 (approved):

PR #1871, opened 2022-11-16
fix: allow to mock global function in script setup components

### LINKED ISSUE #1869: Bug: Mocking not working when setup() is defined
**Describe the bug**
I'm trying to mock the vuei18n $t. But this is not working when using the setup composition api. 
When disabling the setup, it works as expected. I also tried with other function names like $test. 
Strange thing is, when using a name without the dollar sign, it also works just normal. 

**To Reproduce**

index.vue
```
<template>
    <div> {{ $t('hello') }}</div>
</template>
<script>
  export default {
    setup(){
        return {} // will not work
    },
  }
</script>
```

test file
```
import { config, mount } from '@vue/test-utils'
import Index from './index'

config.global = {
    mocks: {
        $t: (msg) => msg
    }
}

describe ('Index Tests', () => {
     it ('should mount without crashing', () => {
         const wrapper = mount(Index)
         expect(wrapper).toBeTruthy();
     })
});

````

**Expected behavior**
Should also works when using the composition api.

**Related information:**
- `@vue/test-utils` version: 2.2.3
- `Vue` version: 3.2.45
- `node` version: v18.2.0
- `npm` (or `yarn`) version: 8.10.0

### PR DESCRIPTION
**Pull request by cexbrayat**
Fixes #1869

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -478,16 +478,35 @@ export function mount(
   if (global?.mocks) {
     const mixin = defineComponent({
       beforeCreate() {
-        for (const [k, v] of Object.entries(
-          global.mocks as { [key: string]: any }
-        )) {
-          // we need to differentiate components that are or not not `script setup`
-          // otherwise we run into a proxy set error
-          // due to https://github.com/vuejs/core/commit/f73925d76a76ee259749b8b48cb68895f539a00f#diff-ea4d1ddabb7e22e17e80ada458eef70679af4005df2a1a6b73418fec897603ceR404
-          // introduced in Vue v3.2.45
-          if (hasSetupState(this)) {
-            this.$.setupState[k] = v
-          } else {
+        // we need to differentiate components that are or not not `script setup`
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/mount.ts [hunk 1]
```diff
@@ -478,16 +478,35 @@ export function mount(
...
-          global.mocks as { [key: string]: any }
-        )) {
-          // we need to differentiate components that are or not not `script setup`
-          // otherwise we run into a proxy set error
-          // due to https://github.com/vuejs/core/commit/f73925d76a76ee259749b8b48cb68895f539a00f#diff-ea4d1ddabb7e22e17e80ada458eef70679af4005df2a1a6b73418fec897603ceR404
-          // introduced in Vue v3.2.45
-          if (hasSetupState(this)) {
-            this.$.setupState[k] = v
-          } else {
+        // we need to differentiate components that are or not not `script setup`
```
lmiller1990, line 481: Test Utils is getting very complex :sweat_smile:

lmiller1990 (approved):

Issue #1879, opened 2022-11-21
Bug: data callback on mount/shallowmount does not work for composition / class component


AWoelfel:
**Describe the bug**
i use `<script setup/>` composite component api and want to use the `data : () => Record<string, unknown>` callback in mount/shallowmount

**To Reproduce**
following test fails if used with the component below
```
it('should handle setData on composition api component', async () => {
    const wrapper = mount(SimpleData)
    expect(wrapper.html()).toEqual(
      '<div>sArray:[\n  "initialA",\n  "initialB"\n  ]|s:initialC</div>'
    )

    await wrapper.setData({
      dataStringArray: ['setA', 'setB'],
      dataString: 'setC'
    })

    expect(wrapper.html()).toEqual(
      '<div>sArray:[\n  "setA",\n  "setB"\n  ]|s:setC</div>'
    )
  })
```

Component `SimpleData.vue`
```
<template>
  <div>sArray:{{dataStringArray}}|s:{{dataString}}</div>
</template>


<script setup lang="ts">

import {ref} from "vue";

const dataStringArray = ref<string[]>(["initialA","initialB"])
const dataString = ref<string>("initialC")

</script>
```

while this test (using an inline declaration will not fail)

```
  it('should handle setData on inline component component', async () => {
    const Component = {
      template: `<div>sArray:{{dataStringArray}}|s:{{dataString}}</div>`,
      data: () => ({
        dataStringArray: ['initialA', 'initialB'],
        dataString: 'initialC'
      })
    }
    const wrapper = mount(Component, {
      data: () => {
        return {
          dataStringArray: ['setA', 'setB'],
          dataString: 'setC'
        }
      }
    })

    expect(wrapper.html()).toEqual(
      '<div>sArray:[\n  "setA",\n  "setB"\n  ]|s:setC</div>'
    )
  })
```

**Expected behavior**
both tests should not fail

**Related information:**
- `@vue/test-utils` version: 2.2.4 
- `Vue` version: 3.2.33
- `node` version: 16.18.0
- `npm` version: 8.19.2

cexbrayat:
Hi @AWoelfel 

This is expected (and mentioned in the docs, see https://test-utils.vuejs.org/api/#setdata)

If you want to update the internal state of a script setup component, you can use `wrapper.vm.dataString = 'hello'; await nextTick();`. `setData` is only working for components declared with the options API

cexbrayat:
setData is still in VTU v2 for components written with the options API. 
[...]

Issue #1851, opened 2022-11-09
Bug: setData with classes/Object prototypes does not work correctly


jacksongross:
**Describe the bug**
When testing a component that calls `setData` with a non-primitive class/object, the object gets set into the component VM without the prototype properties which breaks what should be a straightforward test. If you set the data property on the VM property directly with `wrapper.vm.foo = bar` it works as expected.

**To Reproduce**
Made a very basic example that illustrates the bug relating to the [mergeDeep](https://github.com/vuejs/test-utils/blob/2700e88351fe9853288f974b9d0469f6084380d7/src/utils.ts#L47) behaviour when calling setData
https://stackblitz.com/edit/vue3-script-setup-with-vite-prcxxz?file=src%2Fcomponents%2FHelloWorld.spec.ts,src%2Fcomponents%2FMyObject.js,src%2Fcomponents%2FHelloWorld.vue,src%2FApp.vue

**Expected behavior**
I expect that calling `setData` will simply set the property on the component's VM without modification.

**Related information:**
- `@vue/test-utils` version: 2.0.0 
- `Vue` version: 3.2.37
- `node` version: 16.15.0
- `npm` (or `yarn`) version: 8.10.0

I can see a fix was applied [specifically for Date objects](https://github.com/vuejs/test-utils/issues/1717) but feel the real problem isn't solved.

cexbrayat:
Hi @jacksonGross 

Thanks for the repo! This indeed looks like an issue.

`deepMerge` probably needs to handle this edge case. Our codebase mentions a SO post that was the source of the algorithm https://stackoverflow.com/questions/27936772/how-to-deep-merge-instead-of-shallow-merge/48218209#48218209

It looks like the answer was updated with a reference to https://jhildenbiddle.github.io/mergician/#/ I wonder if this lib handles this use-case? Maybe we could use it, or at least improve `mergeDeep` with their findings.

Would you like to try and open a PR to solve it? Any help would be gladly appreciated!

jacksonGross:
Yeah I might take a stab at this when I have some extra time in the next week or so.

lmiller1990:
Just to check you said 

> @vue/test-utils version: 2.0.0

We are on 2.2.2 now - there has been many releases since 2.0.0, might be worth updating? That said, I don't think we've changed `deepMerge` since then.

jacksonGross:
> Just to check you said
> 
> > @vue/test-utils version: 2.0.0
> 
[...]

Issue #1058, opened 2021-10-16
Class Component issues


navidshad:
I use typescript and vue-class-component on vue3. when I want to change the component data I will get an error:
```ts
await wrapper.setData({ showNextButton: true });
// TypeError: Cannot add property showNextButton, object is not extensible
```
This is the component:
```ts
import { Vue, Options, prop } from "vue-class-component";
import { QuizMode } from "@/common/models/config";

import MarkdownComponent from "@/common/components/layout/Markdown.vue";
import ButtonCheckComponent from "./ButtonCheck.vue";

type SubmitFunction = (configNumber?: number) => Promise<boolean>;

class Props {
  text = prop<string>({ required: true });
  subtext = prop<string>({ default: "" });
  type = prop<"multiple_choice" | "info">({ required: true });
  submitFunction = prop<SubmitFunction>({ required: true });
}

@Options({
  components: {
    MarkdownComponent,
    ButtonCheckComponent,
  },
  inject: ["quiz_mode", "staging", "staging_message"],
  emits: ["next"],
})
export default class ConfigQuestion extends Vue.with(Props) {
  quiz_mode!: QuizMode;
  staging!: boolean;
  staging_message!: string;

  checking = false;
  disabledCheckButton = false;
  disabledNextButton = false;

  status: string | null = null;
  showNextButton = false;

  check(configNumber?: number) {
    this.checking = true;
    this.disabledCheckButton = true;

    this.submitFunction(configNumber)
      .then((success) => {
        this.status = success ? "correct" : "wrong";
        this.showNextButton = true;
        console.log('showNextButton', this.showNextButton);
        
      })
      .finally(() => {
        this.checking = false;
      });
  }

  next() {
    if (this.disabledNextButton) return;

    this.disabledNextButton = true;
    this.$emit("next");
  }
}
```

And this is the test script:
```ts
import { expect } from 'chai';
import { shallowMount, mount, VueWrapper, flushPromises } from '@vue/test-utils';

import ConfigQuestionComponent from '@/modules/lab/components/ConfigQuestion.vue';
import { ComponentPublicInstance, nextTick } from 'vue';

describe("Lab Config Component", () => {

	let is_submitFUnctionRun = false;
	let wrapper: VueWrapper<ComponentPublicInstance>;

	before(() => {

		wrapper = shallowMount(ConfigQuestionComponent, {
			props: {
				text: "This is a config question",
				subtext: "this is a subline",
				type: "info",
[...]

PR #1880, opened 2022-11-21
initial data & setData for composite & class components

AWoelfel:
I want to be able to use `setData` and `mount/shallowmount` `data: () => {}` callbacks to modify data in my composite and class components.

This should handle: 
- #1879 
- #1058 
and maby #1851 as well

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -516,6 +516,24 @@ export function mount(
     app.mixin(mixin)
   }
 
+  // mixin for initial data of script setup composition components
+  if (options?.data) {
+    const providedData = options.data()
+    const mixin = defineComponent({
+      beforeCreate() {
+        if (hasSetupState(this)) {
+          for (const [k, v] of Object.entries(
+            providedData as { [key: string]: any }
+          )) {
+            this.$.setupState[k] = v
+          }
+        }
+      }
+    })
+
+    app.mixin(mixin)
+  }
+
   // AppConfig
   if (global.config) {
     for (const [k, v] of Object.entries(global.config) as [
--- a/src/vueWrapper.ts
+++ b/src/vueWrapper.ts
@@ -242,10 +242,16 @@ export class VueWrapper<
   }
 
   setData(data: Record<string, unknown>): Promise<void> {
-    mergeDeep(this.componentVM.$data, data)
+    //this.componentVM.$data is not defined on all composition api components
+    //script setup components store data in root node instead of $data
+    if (hasSetupState(this.vm)) {
+      const target = this.componentVM as Record<string, unknown>
+      mergeDeep(target, data)
+    } else {
+      mergeDeep(this.componentVM.$data, data)
+    }
     return nextTick()
   }
-
   setProps(props: Record<string, unknown>): Promise<void> {
     // if this VM's parent is not the root or if setProps does not exist, error out
     if (this.vm.$parent !== this.rootVM || !this.__setProps) {
--- a/tests/components/SimpleData.vue
+++ b/tests/components/SimpleData.vue
@@ -1,0 +1,13 @@
+<template>
+  <div>sArray:{{dataStringArray}}|s:{{dataString}}</div>
+</template>
+
+
+<script setup lang="ts">
+
+import {ref} from "vue";
+
+const dataStringArray = ref<string[]>(['initialA','initialB'])
+const dataString = ref<string>('initialC')
+
+</script>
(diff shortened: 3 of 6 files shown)


cexbrayat (commented):
As I mentioned in #1879 , I don't think supporting `setData`/initial data on components written with Composition API really makes sense, as the `data` field no longer exists in these components.

But I don't think strongly about it. @lmiller1990 @freakzlike what do you think ?

The same thing can be achieved via `wrapper.vm.prop = value`

src/mount.ts
```diff
@@ -516,6 +516,24 @@ export function mount(
...
+          for (const [k, v] of Object.entries(
+            providedData as { [key: string]: any }
+          )) {
+            this.$.setupState[k] = v
+          }
+        }
+      }
+    })
+
+    app.mixin(mixin)
```
freakzlike, line 534: Wouldn't this be called for every component? Not only the component which has been passed to `mount`

freakzlike (commented):
I don't feel comfortable using `data`/`setData` for Composition API. Maybe using a different name/functions for this?

src/mount.ts
AWoelfel: you are right, as i added this as a mixin it will be called for every component in this App.
I will have a look if there is another convenient way to set the data fields.

lmiller1990:
I agree with @cexbrayat, `setData` is `data` as in the Options API `data` function. This will be confusing.

Is there a case where you can't test your component without a `setSetupData` function? One of the explicit goals of the Script Setup is [closed by default](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0040-script-setup.md#exposing-components-public-interface). The way Vue recommends exposing those is via `defineExposes`:

> To explicitly expose properties in a <script setup> component, use the defineExpose compiler macro

So, if we support this, would it makes sense to call it `setExposes` or `setExposedData`? You'd need to call `defineExposes` on whatever variables you'd like to modify in your tests, though.

lmiller1990:
Left a comment ^

Interested in this but would like to at least discuss a tiny bit more - any new feature is more complexity and maintenance so we should carefully consider based on use cases, edge cases, back/forward compatibility, etc.

merkle-rking:
> Is there a case where you can't test your component without a `setSetupData` function? I haven't encountered a compelling use case so far - I usually just set the state how a user does, by triggering events, etc.

We've recently migrated from Vue 2 to Vue 3 and are embracing [knowing what to test](https://v1.test-utils.vuejs.org/guides/#common-tips) given the changes when moving to the Composition API. It's been quite a mentality switch from our 100% line coverage approach in Vue 2.

Today, I may have run into such a case where having access to the underlying data in the setup() would be helpful, if not necessary. I'm trying to write tests around a file input component. The file input's value cannot be directly modified for security reasons:

`InvalidStateError: This input element accepts a filename, which may only be programmatically set to the empty string.`

I'm having trouble figuring out how to get the input populated with a File() object in our tests so I can test various things (validation, the file being passed through to an API call upon form submission, etc.). Any thoughts on how to bridge this gap would be appreciated, but it may be the use-case you're looking for.

I should also mention that while we have been able to adapt to the idea that we have to emulate user actions to set values in controls, i.e. element.trigger('click'), element.setValue('foo'), etc, there's a noticeable friction in this testing when compared to the old methodology. As an example, Vuetify 3.1 recently made some changes to some of their controls and layouts that are really unimportant to us from an implementation standpoint. They changed some of the control activation events from "click" to "mousedown". These have the same effect in a browser and are almost entirely transparent to the user. For us developers, this means our tests failed upon upgrade and we had to refactor our emulated user actions. In the Options API, we would have just set the data directly with the expectation that our controls' internals are unimportant. Locating elements used in Teleports is another such example. Instead of finding an array of elements and clicking on all of them, we would've just set the data to an array in our tests.

Admittedly, the new approach gives us the stronger assertion that our controls are correctly tied to data elements because we're forced to interact with the control to engage the data.

merkle-rking:
Coincidentally, I may have stumbled on a second use-case. We have a component that allows for updating a resource server-side. When that resource is updated, the change is logged. On the front-end, we have a View that renders both that resource and the log of the changes to the resource. When the resource is changed, we also want to have the log events on page get refreshed. The log events rendering is abstracted into its own separate component. To facilitate the forced re-render, we're using they key-changing technique. Within our tests for the main View, the log component is stubbed. The testing for the log component is separated. Our snapshots in the main View contain the stubbed representation of our log component, but because there is no representation of the `:key` in the stub, we're unable to test that when a user does an action in the main View, it forces a re-render of the log component. One alternative solution to exposing setup() data would be to have `:key` be reflected in the stub so we can assert against the value incrementing in our snapshots when we initiate user clicks. We could of course mount the log component directly in the main View's tests, but it's much cleaner to separate the concerns.

cexbrayat:
Superseeded by https://github.com/vuejs/test-utils/pull/2166

PR #1924, opened 2022-12-27
[docs]: improve slot scope auto exposing "params" behaviour

renatodeleao:
Hello again 👋 
I had a bunch of specs that rely on the previous VTU1 behaviour, where slot scope was auto-exposed as a `props` object for string templates and realised that in VTU2 they are exposed as `params`

https://github.com/vuejs/test-utils/blob/39d86ad5abe71bac8d76c5b82f11f30225cf8673/src/utils/compileSlots.ts#L9-L11

While VTU1 docs [mention](https://v1.test-utils.vuejs.org/api/options.html#scopedslots) that behaviour, VTU2 docs [didn't **explicitly** do it](https://test-utils.vuejs.org/guide/advanced/slots.html#scoped-slots) as the only example provided used an `<template #scope="params">` which doesn't document the implicit behaviour if not providing a wrapping `<template>` tag.

Update both main docs and migration guide, since I believe this is a common _bump in the road_ while migrating codebases.

As always feel free to edit the writing style and examples to suite the rest of the docs.
Cheers ✌️

### DIFF
--- a/docs/guide/advanced/slots.md
+++ b/docs/guide/advanced/slots.md
@@ -139,7 +139,7 @@ test('layout full page layout', () => {
 
 ## Scoped Slots
 
-[Scoped slots](https://v3.vuejs.org/guide/component-slots.html#scoped-slots) and bindings are also supported.
+[Scoped slots](https://v3.vuejs.org/guide/component-slots.html#scoped-slots) and bindings are also supported. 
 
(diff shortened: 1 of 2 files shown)


cexbrayat (approved):
Thanks @renatodeleao 👍

PR #1926, opened 2022-12-29
[docs]: fix broken  slots documentation

miguel-tzab:
The last change broke the conclusion section.


cexbrayat (approved):
Good catch thanks!

PR #1950, opened 2023-01-26
chore: add proper dependencies

cexbrayat:
We were only declaring `devDependencies`, so this refactors the `package.json` to use `dependencies` when it makes sense. `@vue/runtime-core` was also used in some imports, where we could use `vue` directly, so this commit changes this as well.


lmiller1990:
Does this make sense? What if people have a different version of Vue 3 to the one we specified? I am not sure if Test Utils should have a hard dependency on a specific version of Vue 3. Vue Router, for example, does not: https://github.com/vuejs/vue-router/blob/dev/package.json

cexbrayat:
I'm not sure about what should be the best practice TBH. @Jinjiang pointed that out on Discord, but maybe this should stay as dev deps and all 3 deps should be specified in peer deps? Or maybe we specify a range for the deps?

freakzlike:
`peerDependencies` might be the correct on for `@vue/compiler-dom` and `vue`. I would still set `js-beautify` as dependency and pin to `1.14.6` until #1834 is fixed

cexbrayat:
I updated the PR with:
- js-beautify as a (set) dependency
- compiler-dom as a peer dep

Please take another look @lmiller1990 @freakzlike @Jinjiang when you have some time and let me know what you think.

(I wonder if we should have `^3.0.1` for the Vue peer deps, I'm not sure this is correct, as we probably depend on stuff introduced in way more recent versions... but this can be changed in another PR)

freakzlike (approved):

Jinjiang (approved):

PR #1952, opened 2023-01-28
fix: use attachTo in failing tests with jsdom v21.1

cexbrayat:
`attachTo` can be removed when https://github.com/jsdom/jsdom/issues/3502 is fixed
 With jsdom v21.1, `getComputedStyle` is cached and not updated for elements not in the DOM
 so some our tests fail (because the elements are checked a first time)

We'll probably see issues in Vue projects using VTU and jsdom as well, and we can point to this PR if that happens

### DIFF
--- a/tests/isVisible.spec.ts
+++ b/tests/isVisible.spec.ts
@@ -50,7 +50,10 @@ describe('isVisible', () => {
         }
       }
     }
-    const wrapper = mount(Comp)
+    // TODO: attachTo can be removed when https://github.com/jsdom/jsdom/issues/3502 is fixed
+    // with jsdom v21.1, getComputedStyle is cached and not updated for elements not in the DOM
+    // so the test fails (because the element is checked a first time)
+    const wrapper = mount(Comp, { attachTo: document.body })
 
     expect(wrapper.find('span').isVisible()).toBe(true)
     await wrapper.find('button').trigger('click')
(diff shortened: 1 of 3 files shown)


tests/isVisible.spec.ts
```diff
@@ -50,7 +50,10 @@ describe('isVisible', () => {
         }
       }
     }
-    const wrapper = mount(Comp)
+    // TODO: attachTo can be removed when https://github.com/jsdom/jsdom/issues/3502 is fixed
+    // with jsdom v21.1, getComputedStyle is cached and not updated for elements not in the DOM
+    // so the test fails (because the element is checked a first time)
+    const wrapper = mount(Comp, { attachTo: document.body })
```
lmiller1990, line 56: Nice. I almost wonder if attachTo: document.body should be the default behavior. I wonder if that'd be a breaking change?

lmiller1990 (approved):
Fine either way, but any reason we need to update to this version of jsdom asap? Should we just wait for the release w/ the patch?

cexbrayat:
@lmiller1990 We don't have to, I'm fine either way 😉

It's just that:
- I was happy to have a workaround and to share it with you
- I'm not even sure jsdom is going to fix it
- I wanted to have something to point out if someone opens an issue (we had the same issues in our projects of course)
- It unblocks Renovate

tests/isVisible.spec.ts
cexbrayat: Maybe we should, I don't know what are the downsides (most testing libs in other ecosystems do attach components right?). As you have more context, do you know why this is not the case in VTU?

lmiller1990:
That's fair - happy to move forward with this.

Once we merge up all the deps PRs, we can do another release - good to have the latest code in the wild, even if it's just dev dependencies updates that don't make a difference - good to keep a fast release cadence.

PR #1962, opened 2023-02-02
fix: TS workaround for Vue v3.2.47 

cexbrayat:
A change of types in core (https://github.com/vuejs/core/pull/2818) introduced a TS compilation error when bumping to Vue v3.2.47.
I couldn't figure out a proper way to fix, but using `EmitOptions` instead of `E` (which extends `EmitOptions`...) fixes the issue.
It may be good enough to allow to update to the latest Vue releases, and I think it should have minimal impact on our users (or none at all).


lmiller1990 (approved):

lmiller1990:
Doing this as a patch, hope it won't be a breaking change for some versions 🤔

lmiller1990:
It's out 2.2.10

PR #1964, opened 2023-02-03
feat: add support for input with max-length prop

OXXXRA:
Inherited problem from v1, - https://github.com/vuejs/vue-test-utils/issues/1864
This PR solve this issue and add support for max-length prop for input include tests

### DIFF
--- a/src/domWrapper.ts
+++ b/src/domWrapper.ts
@@ -117,6 +117,10 @@ export class DOMWrapper<NodeType extends Node> extends BaseWrapper<NodeType> {
       }
       return this.trigger('change')
     } else if (tagName === 'INPUT' || tagName === 'TEXTAREA') {
+      if (
+        element.hasAttribute('maxlength') &&
+        element.maxLength < value
+      ) value = value.slice(0, element.maxLength);
       element.value = value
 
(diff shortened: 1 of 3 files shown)


OXXXRA:
@cexbrayat, i fixed lint now it should work!

cexbrayat:
Hi @OXXXRA 

Thanks for the PR!

I'm personally not convinced this is something we should do in VTU though. I don't really see the value of handling this specific use-case. And if we do it, we would need to support all the potential attributes like `minlength`, `max`, `min`, `pattern`, etc to be coherent.

So I think we should drop this. But let's hear from other members of the team what they think

cc @lmiller1990 @freakzlike

lmiller1990:
The PR and code itself is 10/10 - clean fix, test added, good description. Unfortunately, I am inclined to agree that this isn't something Test Utils should be supporting.  Handling `maxlength` would be something that browsers handle case-by-case - we definitely don't want to re-implement all these rules. 

I suspect that using JS in *any* browser will ignore `maxlength`, anyway - until you submit the form, anyway, in which case the default browser validation should prevent it from submitting. I wonder how this works in jsdom and happydom 🤔

PR #1966, opened 2023-02-06
docs: change type for unableAutoUnmount

voskresla:
I think this change for `enableAutoUnmount` type makes it easier to understand.

If ok, i can change types in this PR too )

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -1886,7 +1886,7 @@ function shallowMount(Component, options?: MountingOptions): VueWrapper
 **Signature:**
 
 ```ts
-enableAutoUnmount(hook: (callback: Function) => void));
+enableAutoUnmount(hook: (callback: () => void) => void);
 disableAutoUnmount(): void;
 ```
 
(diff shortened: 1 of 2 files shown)


docs/api/index.md
```diff
@@ -1886,7 +1886,7 @@ function shallowMount(Component, options?: MountingOptions): VueWrapper
 **Signature:**
 
 ```ts
-enableAutoUnmount(hook: Function));
+enableAutoUnmount(hook: (callback: Function) => void));
```
cexbrayat, line 1889: Maybe even `hook: (callback: () => void) => void`?

cexbrayat (commented):
I think you can update the types at the same time 👍

docs/api/index.md
voskresla: I will update types, thx.

lmiller1990 (approved):

docs/api/index.md
cexbrayat: To be coherent, maybe change the type here in the docs as well to be the same as the TS types?

PR #1971, opened 2023-02-10
feat: add renderToString function

cexbrayat:
It is now possible to use `renderToString` to test the SSR rendering of a component.

To do so, use `renderToString(MyComponent)`.
`renderToString` returns a `Promise<string>` with the HTML rendered.

```ts
  it('returns correct html with pre-fetched data on server', async () => {
      template: '<div>{{ text }}</div>',
      setup() {
        const text = ref<string | null>(null)
        onServerPrefetch(async () => {
          text.value = await fakeFetch('onServerPrefetch')
        })
        return { text }
      }
    })

    const contents = await renderToString(Component)

    expect(contents).toBe('<div>onServerPrefetch</div>')
  })
```

### DIFF
--- a/src/renderToString.ts
+++ b/src/renderToString.ts
@@ -1,6 +1,6 @@
+import { renderToString as baseRenderToString } from '@vue/server-renderer'
 import {
   FunctionalComponent,
-  ComponentPublicInstance,
   ComponentOptionsWithObjectProps,
   ComponentOptionsWithArrayProps,
   ComponentOptionsWithoutProps,
@@ -17,10 +17,8 @@ import {
   ComponentPropsOptions,
   Prop
 } from 'vue'
+
 import { MountingOptions } from './types'
-import { VueWrapper } from './vueWrapper'
-import { trackInstance } from './utils/autoUnmount'
-import { createVueWrapper } from './wrapperFactory'
 import { createInstance } from './createInstance'
 
 // NOTE this should come from `vue`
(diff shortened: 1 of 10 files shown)


cexbrayat:
@lmiller1990 @freazlike This is a rebased/tidied up version of #1572 
(@wobsoriano is still marked as the author of the commit 😉)

cexbrayat:
I forgot to mention I added types tests and a bit of documentation

freakzlike (approved):
Looks good and I think it's a good starting point. 👍

src/renderToString.ts
```diff
@@ -0,0 +1,217 @@
...
+    Defaults
+  >,
+  options?: MountingOptions<
+    Partial<Defaults> & Omit<Props & PublicProps, keyof Defaults>,
+    D
+  > &
+    Record<string, any>
+): Promise<string>
+
+// component declared by vue-tsc ScriptSetup
```
lmiller1990, line 125: Did we copy paste this entire thing? Is it possible to share it somehow?

lmiller1990 (approved):
Let's ship it and start to get feedback! Thanks @cexbrayat and @websoriano

src/renderToString.ts
cexbrayat: @lmiller1990 Yes, it's nearly the same as mount, but returns a `Promise<string>`. I tried to extract it with a mapped type, but sadly all the tricks I found did not work (you can map one signature, but the overloads aren't mapped)😞

Issue #1973, opened 2023-02-11
"vue-tsc" fails with "error TS2769: No overload matches this call" in "strict" mode


tinobino:
**Describe the bug**
`vue-tsc` fails with "error TS2769: No overload matches this call" in when "strict" mode is configured in "tsconfig.json".

If the `defineEmits` part in the component is removed `vue-tsc` doesn't complain. Seems that this issue is somehow related with Vue version 3.2.46.

**To Reproduce**
https://stackblitz.com/edit/vitest-dev-vitest-nwzj9x?file=components%2FSlotComponent.vue,test%2FSlotComponent.test.ts,package.json&initialPath=__vitest

**Expected behavior**
`vue-tsc` does not fail for the example given.

cexbrayat:
Hi @tinobino and thanks for the repro

After looking at your repro, it does look like there is an issue, but I don't know if that should be fixed in Vue core, VTU or Volar.

The error goes away if:
- Vue v3.2.45 is used
- if `defineEmits` is removed
- More surprising, if `<slot />` is removed from the template

Let me ping @johnsoncodehk as he may have some idea of what's going on

johnsoncodehk:
It seems due to `DefineComponent` type updates. It should fix if update this line: https://github.com/vuejs/test-utils/blob/f68ceee8675e3274eb80e6b1f5a99cfe093be4ce/src/mount.ts#L204

```diff
- export declare function mount<T extends DefineComponent<any, any, any, any, any>>(component: T, options?: ComponentMountingOptions<T>): VueWrapper<InstanceType<T>>;
+ export declare function mount<T extends DefineComponent<any, any, any, any, any, any, any, any, any, any>>(component: T, options?: ComponentMountingOptions<T>): VueWrapper<InstanceType<T>>;
```

cexbrayat:
@johnsoncodehk Thanks for taking a look! Sadly, it doesn't look like it's fixing the issue
See https://github.com/vuejs/test-utils/pull/1974

tinobino:
Hi guys! Any updates on this? This issue prevents us from upgrade to latest vue version.

cexbrayat:
@johnsoncodehk kindly offered (on another channel) to refactor the types when he has time, so this issue will get fixed, but we can't give you a timeline.

In the meantime, you can still upgrade to the latest vue version, by adding an explication cast `mount(SlotComponent as any, {...})`. We understand this is not ideal, but that should unblock you to update

[...]

PR #1974, opened 2023-02-11
repro: #1973

cexbrayat:


johnsoncodehk:
Add more 3 `any` fixed it. 😅

`T extends DefineComponent<any, any, any, any, any, any, any, any, any, any, any, any>`

PR #1979, opened 2023-02-15
fix: Stub instance of the same component

### LINKED ISSUE #1162: shallowMount doesn't stub instances of the same component
# Expected behavior

When using `shallowMount` on a component that contains itself, nested instances should be stubbed, so the output for the reproduction in the shared repo should be something like:

```
    <ul>
      <li>Option 1
        <!--v-if-->
      </li>
      <li>Option 2 <list-stub></list-stub>
      </li>
      <li>Option 3
        <!--v-if-->
      </li>
    </ul>
```

# Current behavior

Nested instances are being mounted as if we were using `mount`. Current output is:

```
    <ul>
      <li>Option 1
        <!--v-if-->
      </li>
      <li>Option 2 <ul>
          <li>Option 2.1
            <!--v-if-->
          </li>
          <li>Option 2.2
            <!--v-if-->
          </li>
        </ul>
      </li>
      <li>Option 3
        <!--v-if-->
      </li>
    </ul>
```

# Reproduction:

https://github.com/victorborrasdev/vue-test-utils-nested-stubs

Please let me know if you need any additional info. Thanks!

freakzlike:
Fixes #1162 

I'm not really happy with the wording, because there is a lot of double negation. However, the code does the job

### DIFF
--- a/src/vnodeTransformers/stubComponentsTransformer.ts
+++ b/src/vnodeTransformers/stubComponentsTransformer.ts
@@ -36,10 +36,14 @@ interface StubOptions {
   renderStubDefaultSlot?: boolean
 }
 
-const doNotStubComponents: WeakSet<ConcreteComponent> = new WeakSet()
+const doNotStubComponents = new WeakMap<ConcreteComponent, boolean>()
 const shouldNotStub = (type: ConcreteComponent) => doNotStubComponents.has(type)
-export const addToDoNotStubComponents = (type: ConcreteComponent) =>
-  doNotStubComponents.add(type)
+const shouldNotStubRoot = (type: ConcreteComponent) =>
+  !!doNotStubComponents.get(type)
+export const addToDoNotStubComponents = (
+  type: ConcreteComponent,
+  onlyRoot?: boolean
+) => doNotStubComponents.set(type, !!onlyRoot)
 
 const normalizeStubProps = (props: ComponentPropsOptions) => {
   // props are always normalized to object syntax
@@ -163,7 +167,10 @@ export function createStubComponentsTransformer({
     }
 
     if (shouldNotStub(type)) {
-      return type
+      // Either don't stub everytime or only on root level
+      if (!instance?.parent || !shouldNotStubRoot(type)) {
+        return type
+      }
     }
 
     const registeredName = getComponentRegisteredName(instance, type)
--- a/src/vnodeTransformers/stubComponentsTransformer.ts
+++ b/src/vnodeTransformers/stubComponentsTransformer.ts
@@ -36,15 +36,6 @@ interface StubOptions {
   renderStubDefaultSlot?: boolean
 }
 
-const doNotStubComponents = new WeakMap<ConcreteComponent, boolean>()
-const shouldNotStub = (type: ConcreteComponent) => doNotStubComponents.has(type)
-const shouldNotStubRoot = (type: ConcreteComponent) =>
-  !!doNotStubComponents.get(type)
-export const addToDoNotStubComponents = (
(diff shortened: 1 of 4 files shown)


src/vnodeTransformers/stubComponentsTransformer.ts
```diff
@@ -36,10 +36,14 @@ interface StubOptions {
...
+const doNotStubComponents = new WeakMap<ConcreteComponent, boolean>()
 const shouldNotStub = (type: ConcreteComponent) => doNotStubComponents.has(type)
-export const addToDoNotStubComponents = (type: ConcreteComponent) =>
-  doNotStubComponents.add(type)
+const shouldNotStubRoot = (type: ConcreteComponent) =>
+  !!doNotStubComponents.get(type)
+export const addToDoNotStubComponents = (
+  type: ConcreteComponent,
+  onlyRoot?: boolean
+) => doNotStubComponents.set(type, !!onlyRoot)
```
cexbrayat, line 46: maybe we could have two distinct functions, with different names to make it easier to understand?
like `addToDoNotStubComponentsButOnlyInRoot`? (I admit it's not a great name ^^)

cexbrayat (approved):
Thanks for taking a stab at this @freakzlike 
Code looks good, even if I agree it's hard to understand all the negations of negations

xanf:
I'm iterating on this one, and while I'm working on suggestion, would like to share my thoughts:

* we are already tracking "root" instances for auto-destroy purposes, so what I'm thinking might be is drop `addToDoNotStubComponents` logic for root but instead when making stub logic check it from registry of "root" instances (and move `trackInstance` logic and reigstry to separate f ile

freakzlike:
> I'm iterating on this one, and while I'm working on suggestion, would like to share my thoughts:
> 
> * we are already tracking "root" instances for auto-destroy purposes, so what I'm thinking might be is drop `addToDoNotStubComponents` logic for root but instead when making stub logic check it from registry of "root" instances (and move `trackInstance` logic and reigstry to separate f ile

You are right. It's only used for the root components and it might be a better way

freakzlike:
I tried to retrieve or check for the root component without the need of `rootComponents`, but didn't manage to do so for all cases. The logic has not changed now, but it's a bit easier to read I think

lmiller1990:
Need a rebase, since we just released the `renderToString` PR.

Glad @xanf is running this on a real code base - will wait back to hear from him before merging this one up.

src/vnodeTransformers/stubComponentsTransformer.ts
```diff
@@ -102,13 +97,18 @@ const resolveComponentStubByName = (
   }
 }
 
-interface CreateStubComponentsTransformerConfig {
+export interface CreateStubComponentsTransformerConfig {
+  rootComponents: {
+    legacy?: Component
+    component?: Component
+  }
```
cexbrayat, line 104: Would you mind adding a few comments on what each field does?

cexbrayat (approved):
LGTM

This needs to be rebased, as some logic moved to `src/createInstance` now that `renderToString` landed (sorry)

lmiller1990:
Shall we merge or did we want more testing from @xanf? Happy to move forward as-is.

xanf:
@freakzlike @lmiller1990 I'm going to finish testing no later than tomorrow, sorry for delay - our codebase uses v-once, and I need to use vuejs with this PR https://github.com/vuejs/core/pull/7730 merged to test it, cause compiler blows up

lmiller1990:
I think we should merge and release this one. I'll give it another 24h then ship.

PR #1996, opened 2023-03-06
fix: return wrapper error from DOMWrapper ctor if element is nullish

### LINKED ISSUE #1984: Feature request: Make it easier to wrap an arbitrary element/selector
When testing components that Teleport a node to somewhere else, I need to make a wrapper for that teleported HTML element. I can't use .findComponent, because the teleported node might not be a component, and finding by selector doesn't work because the teleported HTML is not in the DOM tree of my root mounted component.

Instead, I have to make the following utility:

```js
export function getWrapper(selector) {
  const el = document.querySelector(selector);
  if (el) return new DOMWrapper(el);
  else return createWrapperError("DOMWrapper");
}
```

This works, but there's a major footgun here because if you don't know about `createWrapperError`, the intuitive thing is to just always return a `DOMWrapper`. But, `DOMWrapper.exists()` always returns true, regardless of whether `element` is nullish or not.

`DOMWrapper.exists` returns false if its element doesn't exist. 

OR export a function similar to the one I wrote above from vue-test-utils that includes this error handling.

Keep writing this wrapper function myself.

**Additional context**

If there's a better way of testing Teleport when it only teleports pure DOM nodes (not components), I'd love to see it added to https://test-utils.vuejs.org/guide/advanced/teleport.html.

ascott18:
fixes #1984 

Rationale: Rather than overriding `exists` on DomWrapper, I elected to instead return a wrapper error from DomWrapper's ctor if the provided element is nullish. 

This allows the ctor of DomWrapper to be changed to allow `null | undefined` without having to propagate `null | undefined` downwards into `BaseWrapper.constructor`, `BaseWrapper.wrapperElement`, `BaseWrapper.element`, and `WrapperLike.element`, a change that would be significantly breaking and cause consumers of test-utils to have to add non-null assertions all over the place.

This also simplifies some other uses cases, like for example the one I changed in plugins.md.

### DIFF
--- a/docs/guide/extending-vtu/plugins.md
+++ b/docs/guide/extending-vtu/plugins.md
@@ -109,11 +109,7 @@ const DataTestIdPlugin = (wrapper) => {
   function findByTestId(selector) {
     const dataSelector = `[data-testid='${selector}']`
     const element = wrapper.element.querySelector(dataSelector)
-    if (element) {
-      return new DOMWrapper(element)
-    }
-
-    return createWrapperError('DOMWrapper')
+    return new DOMWrapper(element)
   }
 
   return {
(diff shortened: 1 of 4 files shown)


cexbrayat (approved):
LGTM 👍

lmiller1990 (approved):

PR #2023, opened 2023-04-08
fix(renderToString): Remove document usage on SSR render

### LINKED ISSUE #2020: Bug: document calls should fail during ssr test
**Describe the bug**
Correct me if I'm wrong, but during `ssr` any access to `document` should fail. I want to test my component in a `node` environment, but `vue test utils` throws an error.

It works if I change environment to `JSDOM`, but then my component will not throw an error.

```
ReferenceError: document is not defined
 ❯ createInstance ../../../node_modules/.pnpm/@vue+test-utils@2.3.1/node_modules/@vue/test-utils/dist/vue-test-utils.esm-bundler.mjs:8020:14
 ❯ Module.renderToString ../../../node_modules/.pnpm/@vue+test-utils@2.3.1/node_modules/@vue/test-utils/dist/vue-test-utils.esm-bundler.mjs:8281:15
```

I think we shouldn't use `document` inside `createInstance` function

https://github.com/vuejs/test-utils/blob/main/src/createInstance.ts#L114
https://github.com/vuejs/test-utils/blob/main/src/createInstance.ts#L119

**To Reproduce**
https://stackblitz.com/edit/vitest-dev-vitest-nnsmqh?initialPath=__vitest__

**Expected behavior**
Tests work in `node` environment

**Related information:**
```
System:
    OS: Windows 10 10.0.19044
    CPU: (16) x64 11th Gen Intel(R) Core(TM) i7-11700KF @ 3.60GHz
    Memory: 3.58 GB / 15.83 GB
  npmPackages:
    vitest: ^0.29.2 => 0.29.2
```

freakzlike:
`renderToString` unit test now runs in `node` environment and not `jsdom`.

The `attachTo` option has no usage on `renderToString`. Should we print a warning, when it is used?

Fix #2020

### DIFF
--- a/src/createInstance.ts
+++ b/src/createInstance.ts
@@ -111,24 +111,6 @@ export function createInstance(
   // Let's register it as a stub so user can find it
   registerStub({ source: originalComponent, stub: component })
 
-  const el = document.createElement('div')
-
-  if (options?.attachTo) {
-    let to: Element | null
-    if (typeof options.attachTo === 'string') {
-      to = document.querySelector(options.attachTo)
-      if (!to) {
-        throw new Error(
-          `Unable to find the element matching the selector ${options.attachTo} given as the \`attachTo\` option`
-        )
-      }
-    } else {
-      to = options.attachTo
-    }
-
-    to.appendChild(el)
-  }
-
   function slotToFunction(slot: Slot) {
     switch (typeof slot) {
       case 'function':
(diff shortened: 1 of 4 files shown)


cexbrayat (approved):
👍 Maybe we could:
- add a warning if `attachTo` is used in renderToString
- exclude it from the mounting options types for renderToString

PR #2024, opened 2023-04-09
fix(renderToString): Remove attachTo from options and print warning on usage

freakzlike:
Prints a warning when using `attachTo` option on `renderToString`.

I had to use `attachTo?: never` because the options are typed with `RenderMountingOptions<any> & Record<string, any>` (See https://github.com/vuejs/test-utils/pull/1079)

Related to #2023


cexbrayat (approved):
👌

PR #2040, opened 2023-04-30
fix: allow finding async component stubs by definition

xanf:
This MR introduces an ability to find async component stubs by definition of **underlying** component, not just by wrapper itself

It also mimics `@vue/test-utils@1.x` behavior (well, kinda) - `@vue/test-utils@1.x` was unable to stub async components due to limitations in our patch approach, so "finding" always worked cause there were no stubs

### DIFF
--- a/src/vnodeTransformers/stubComponentsTransformer.ts
+++ b/src/vnodeTransformers/stubComponentsTransformer.ts
@@ -11,8 +11,8 @@ import {
   ConcreteComponent,
   ComponentPropsOptions,
   ComponentObjectPropsOptions,
-  DefineComponent,
-  Component
+  Component,
+  ComponentOptions
 } from 'vue'
 import { hyphenate } from '../utils/vueShared'
 import { matchName } from '../utils/matchName'
@@ -28,6 +28,7 @@ import { registerStub } from '../stubs'
 export type CustomCreateStub = (params: {
   name: string
   component: ConcreteComponent
+  registerStub: (config: { source: Component; stub: Component }) => void
 }) => ConcreteComponent
 
 interface StubOptions {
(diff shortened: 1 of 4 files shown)


src/vnodeTransformers/stubComponentsTransformer.ts
```diff
@@ -230,7 +243,8 @@ export function createStubComponentsTransformer({
       return (
         config.plugins.createStubs?.({
           name: stubName,
-          component: type
+          component: type,
+          registerStub
```
xanf, line 247: We want to provide an ability to mess with stubs (including async componentS) to the plugin system, so our plugins will be as capable as original `createStub`

xanf:
A bit of off-topic: happy to be back to active development here! We have a Vue.js 3 migration working group established in GitLab, so I will be happy to dedicate more time polishing things here and adding new cool features

tests/mountingOptions/global.stubs.spec.ts
```diff
@@ -679,16 +679,16 @@ describe('mounting options: stubs', () => {
...
-      name: 'AsyncComponent',
-      template: '<span>AsyncComponent</span>'
-    }))
-
     const AsyncComponentWithoutName = defineAsyncComponent(async () => ({
       template: '<span>AsyncComponent</span>'
     }))
 
     it('stubs async component with name', async () => {
+      const AsyncComponent = defineAsyncComponent(async () => ({
```
xanf, line 687: We need to move component definition inside test, because otherwise for second test `__asyncResolved` will be already defined from previous test

lmiller1990 (commented):
🎉 

Will give a bit of time for other reviewers to chime in, but tests look good.

lmiller1990 (approved):

cexbrayat (approved):

PR #2043, opened 2023-05-11
feat: update to vue 3.3

dacoto97:


cexbrayat:
Ha I missed your PR, and let Renovate do the update, sorry. It should be done on `main` now that Renovate did its job.
We'll need a release, but it will probably be an alpha release first, as we are testing new type signatures and we don't want to introduce regressions.

PR #2052, opened 2023-05-12
Update index.md

DespertaWeb:
Make sure Composition API Users are aware of it!!!!

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -1526,7 +1526,9 @@ setData(data: Record<string, any>): Promise<void>
 
 `setData` does not allow setting new properties that are not defined in the component.
 
+::: warning
 Also, notice that `setData` does not modify composition API `setup()` data.
+:::
 
 `Component.vue`:
 


docs/api/index.md
```diff
@@ -1526,7 +1526,9 @@ setData(data: Record<string, any>): Promise<void>
 
 `setData` does not allow setting new properties that are not defined in the component.
 
+::: warning
 Also, notice that `setData` does not modify composition API `setup()` data.
+:::
```
cexbrayat, line 1531: Is that a special syntax that Github supports? It does not look like it does in the rich diff.

cexbrayat (changes requested):

docs/api/index.md
cexbrayat: Ha my bad, I thought it was in the readme, not in the docs. Looks good then

cexbrayat (approved):

PR #2066, opened 2023-05-19
chore: use a stackblitz repro with create-vue for issue template

cexbrayat:
As we do receive quite a few issues about types, I think it would be nice to use repros that come with vue-tsc (and vitest of course) out of the box.

Note that this link can point to any of the snapshots listed here https://github.com/vuejs/create-vue-templates/tree/main but I suppose a minimal one by default is better


freakzlike (approved):
Yes, makes sense. I would prefer to have only one vue component instead of all these, but this would require a separate template so I think we can take this 👍

lmiller1990 (approved):

PR #2067, opened 2023-05-19
fix(stub): rename prefix prop to avoid warning on render

### LINKED ISSUE #2035: Bug: Failed setting prop "prefix" on [component]
**Describe the bug**
There is a set-up with Vitest, Vue 3, and Vuetify 3. When running tests for components that contain Vuetify components there appears a warning about the `prefix` prop. The full warning looks like this:

```
[Vue warn]: Failed setting prop "prefix" on <v-text-field-stub>: value undefined is invalid. TypeError: Cannot set property prefix of #<Element> which has only a getter
    at patchDOMProp (/vuetify-project/node_modules/@vue/runtime-dom/dist/runtime-dom.cjs.js:269:17)
    at patchProp (/vuetify-project/node_modules/@vue/runtime-dom/dist/runtime-dom.cjs.js:387:9)
    at mountElement (/vuetify-project/node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:5164:21)
    at processElement (/vuetify-project/node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:5136:13)
    at patch (vuetify-project/node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:5056:21)
    at ReactiveEffect.componentUpdateFn [as fn] (/vuetify-project/node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:5594:21)
    at ReactiveEffect.run (/vuetify-project/node_modules/@vue/reactivity/dist/reactivity.cjs.js:194:25)
    at instance.update (/vuetify-project/node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:5701:56)
    at setupRenderEffect (/vuetify-project/node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:5715:9)
    at mountComponent (/vuetify-project/node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:5497:9) 
  at <VTextField> 
  at <HelloWorld ref="VTU_COMPONENT" > 
  at <VTUROOT>
```

**To Reproduce**
1. Clone the [repo](https://github.com/Oleksii14/vuetify-test-utils-issue)
2. `yarn install`
3. `yarn test:unit`

**Expected behavior**
No warnings appear

**Related information:**
Created [issue](https://github.com/vuetifyjs/vuetify/issues/17185) in Vuetify repo

freakzlike:
My only solution currently would be to rename the `prefix` property on render to `_prefix` for example. Not nice but this would only affect the `.html()` output `.props('prefix')` would still work.

In case we decide to use this solution, then I would add something to the docs

Fix #2035

### DIFF
--- a/src/vnodeTransformers/stubComponentsTransformer.ts
+++ b/src/vnodeTransformers/stubComponentsTransformer.ts
@@ -37,17 +37,28 @@ interface StubOptions {
   renderStubDefaultSlot?: boolean
 }
 
+const mapStubPropValue = (value: unknown) => {
+  if (typeof value === 'symbol') {
+    return value?.toString()
+  } else if (typeof value === 'function') {
+    return '[Function]'
+  }
+
+  return value
+}
 const normalizeStubProps = (props: ComponentPropsOptions) => {
   // props are always normalized to object syntax
   const $props = props as unknown as ComponentObjectPropsOptions
   return Object.keys($props).reduce((acc, key) => {
-    if (typeof $props[key] === 'symbol') {
-      return { ...acc, [key]: $props[key]?.toString() }
+    let propName = key
+
+    // Rename `prefix` prop because it's shared with Element prop and will output a warning on render
+    // https://developer.mozilla.org/en-US/docs/Web/API/Element/prefix
+    if (propName === 'prefix') {
+      propName = '_prefix'
     }
-    if (typeof $props[key] === 'function') {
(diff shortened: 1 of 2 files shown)


freakzlike:
I just saw this issue https://github.com/vuejs/test-utils/issues/1995 with is exactly the same one like this with `attributes`. Therefore we can also drop this, but I would add some things to docs

PR #2100, opened 2023-06-20
docs: note `mount`'s dependence on system time

aentwist:
Closes https://github.com/vuejs/test-utils/issues/2074

### DIFF
--- a/docs/guide/faq/index.md
+++ b/docs/guide/faq/index.md
@@ -2,6 +2,15 @@
 
 [[toc]]
 
+## Mocking Dates and Timers with Vitest
+
+Vue's scheduler depends on the system time. Make sure to mount components
+*after* calling `vi.setSystemTime`, since Vue depends on its side effects.
+Mounting components before calling `vi.setSystemTime` may cause breaks in
+reactivity.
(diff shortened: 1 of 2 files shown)


docs/guide/faq/index.md
```diff
@@ -2,6 +2,14 @@
 
 [[toc]]
 
+## Mocking Dates and Timers with Vitest
+
+`mount` depends on the system time in some way \[needs clarification\]. For
```
lmiller1990, line 7: Do you intend to clarify? If not, we could make this more generic. For example:

"Some edge cases are known to exist when using Vitest and the `vi.setSystemTime` that can interfere with Vue's scheduler, causing unexpected behaviors. If you are using `setSystemTime`, make sure you call it *before* calling `mount` to avoid any unexpected behavior".

aentwist: Is that better? I'm not clarifying because I don't know anything and don't plan to find out. But you pretty much just clarified. And I paraphrased it.

cexbrayat (approved):
Thanks @aentwist 👍

PR #2103, opened 2023-06-22
chore: use node v18 on netlify

cexbrayat:
Currently Netlify builds using node v16, which uses pnpm v7 and fails the build. According to https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-and-javascript adding a `.node-version` file should force Netlify to build with node v18 (and pnpm v8)


cexbrayat:
Still v7, but I think I have another idea

PR #2120, opened 2023-07-03
feat: include the mocks in component globalThis

### LINKED ISSUE #2119: Feature request: global.mocks to also mock into the script setup, not just the template
I am coding with Nuxt and would like to component test following component:

```
<script setup lang="ts">
import { ref } from 'vue'
const hello = ref('hello')

function clicked() {
  //@ts-ignore For example Nuxt will inject these kinds of global functions
  $fetch('http://www.example.com').then((response : any) => {
    hello.value = response.data
  })
}

</script>

<template>
  <button @click="clicked">{{ hello }}</button>
  <!-- this emulates components that use a global function like $t for i18n -->
  <!-- this function can be mocked using global.mocks -->
</template>

```

The problem is the `$fetch` call that is with Nuxt freely available to use in the component. I would like to mock the `$fetch` method.

The test would then be like this:
```
  it('mocks a global function used in a script setup', async () => {
    const wrapper = mount(ScriptSetupWithGlobalFunction, {
      global: {
        mocks: {
          $fetch: async (url) => ({ data: 'mocked' })
        }
      }
    })
    expect(wrapper.text()).toContain('hello')
    await wrapper.find('button').trigger('click')
    expect(wrapper.text()).toContain('mocked')
  })
```

The solution would be that when I mock the `$fetch` it will become available _also_ in the script setup. At the moment it is only available to use in the template code.

jrutila:
In the components beforeCreate we include all the mocks also in `globalThis`. That way the mocks are usable also in the `script setup` part of the component.

This PR contains the unit test and the solution.

Fixes #2119

### DIFF
--- a/src/createInstance.ts
+++ b/src/createInstance.ts
@@ -228,6 +228,8 @@ export function createInstance(
               this.$.setupState[k] = v
               // eslint-disable-next-line no-empty
             } catch (e) {}
+            //@ts-ignore Setting an unknown global variable for the component instance
+            globalThis[k] = v
           }
           // also intercept the proxy calls to make the mocks available on the instance
           // (useful when a template access a global function like $t and the developer wants to mock it)
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -505,6 +505,15 @@ export default {
   }
 }
 </script>
+
+// Or if you are using script setup
+
+<script setup>
+function onClick() {
+  $store.dispatch('click')
+}
+</script>
+
 ```
 
 `Component.spec.js`:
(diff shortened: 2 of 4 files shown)


src/createInstance.ts
```diff
@@ -228,6 +228,8 @@ export function createInstance(
               this.$.setupState[k] = v
               // eslint-disable-next-line no-empty
             } catch (e) {}
+            //@ts-ignore Setting an unknown global variable for the component instance
+            globalThis[k] = v
```
lmiller1990, line 232: Did this work in Test Utils v1? I actually am surprised this doesn't "just work".

lmiller1990 (commented):
Any chance we can add a few more tests around the various globals from Nuxt? I believe they also inject components and a few other things? cc @jrutila

jrutila:
> Any chance we can add a few more tests around the various globals from Nuxt? I believe they also inject components and a few other things? cc @jrutila

Good point. Nuxt injects through auto imports components. I see if I manage to add that case, too.

PR #2122, opened 2023-07-05
docs: add warning about `isVisible()`

alecgibson:
This change updates the documentation to reflect that `isVisible()` will only work correctly [when attached to the DOM][1] because of an upstream [`jsdom` issue][2].

[1]: https://github.com/vuejs/test-utils/issues/2016
[2]: https://github.com/jsdom/jsdom/issues/3502

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -1433,6 +1433,10 @@ isVisible(): boolean
 
 **Details:**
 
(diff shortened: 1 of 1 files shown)


cexbrayat (approved):
Awesome, thanks @alecgibson !

PR #2137, opened 2023-07-25
fix: stricter props types

alecgibson:
At the moment, `mount()` offers [strong typing][1] of props, but this strong typing is lost when dealing with the `VueWrapper` through either the `props()` or `setProps()` methods.

This change strengthens the typing of these methods to help raise compile-time errors when trying to get or set incorrect props.

[1]: https://github.com/vuejs/test-utils/blob/11b34745e8e66fc747881dfb1ce94cef537c455e/src/types.ts#L44

### DIFF
--- a/tests/props.spec.ts
+++ b/tests/props.spec.ts
@@ -20,6 +20,7 @@ describe('props', () => {
 
   it('returns undefined if props does not exist', () => {
     const wrapper = mount(WithProps, { props: { msg: 'ABC' } })
+    // @ts-expect-error :: non-existent prop
     expect(wrapper.props('foo')).toEqual(undefined)
   })
 
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -47,7 +47,13 @@ export function mount<
 >(
   originalComponent: T,
   options?: ComponentMountingOptions<C>
-): VueWrapper<ComponentExposed<C> & ComponentProps<C> & ComponentData<C>>
+): VueWrapper<
+  ComponentProps<C> & ComponentData<C> & ComponentExposed<C>,
+  ComponentPublicInstance<
+    ComponentProps<C>,
+    ComponentData<C> & ComponentExposed<C>
+  >
+>
 
 // implementation
 export function mount(
--- a/src/vueWrapper.ts
+++ b/src/vueWrapper.ts
@@ -215,10 +215,14 @@ export class VueWrapper<
     return this.componentVM
   }
 
-  props(): { [key: string]: any }
-  props(selector: string): any
-  props(selector?: string): { [key: string]: any } | any {
-    const props = this.componentVM.$props as { [key: string]: any }
+  props(): T['$props']
+  props<Selector extends keyof T['$props']>(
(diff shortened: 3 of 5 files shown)


tests/props.spec.ts
```diff
@@ -342,7 +343,7 @@ describe('props', () => {
     })
 
     it('should get props from functional component', async () => {
-      const wrapper = mount(Title, {
+      const wrapper: VueWrapper<any> = mount(Title, {
```
alecgibson, line 346: I had to add this to make the compilation pass; I'm not sure I understand why this isn't working properly with the functional component?

cexbrayat: I'm not sure.

@johnsoncodehk Would you mind taking a look at this PR please? Is this the proper way to type props?

cexbrayat (commented):
I think we need @johnsoncodehk insight

tests/props.spec.ts
lmiller1990: If we add something that requires users to add code to their code base (like this) I think this would be a breaking change, which I'm not sure we'd want to do.

Not sure if Johnson will have time to 👀 , we may need to have a look. Vue types are too complex 😓

lmiller1990:
We may need to just find some wild code bases to validate if this breaks anything. Maybe GitLab (not sure if they are on Vue 3 yet) or Cypress (cypress/vue uses Test Utils)

pikax:
@alecgibson if you change the `mount` to:
```ts
export function mount<
  T,
  C = T extends ((...args: any) => any) | (new (...args: any) => any)
    ? T
    : T extends { props?: infer Props }
    ? DefineComponent<
        Props extends Readonly<(infer PropNames)[]> | (infer PropNames)[]
          ? { [key in PropNames extends string ? PropNames : string]?: any }
          : Props
      >
    : DefineComponent
>(
  originalComponent: T,
  options?: ComponentMountingOptions<C>
): VueWrapper<
  ComponentProps<C> & ComponentData<C> & ComponentExposed<C>,
  ComponentPublicInstance<
    ComponentProps<C>,
    ComponentData<C> & ComponentExposed<C>
  >
>
```

it should work without needing to cast on functional components
```ts
const wrapper: VueWrapper<any> = mount(Title, {
```

tests/props.spec.ts
alecgibson: Okay have applied @pikax 's suggestion and the build is green without this cast

cexbrayat (approved):
LGTM 👍

johnsoncodehk:
LGTM 👍 (Sorry for the delay!)

johnsoncodehk (approved):

DanielleHuisman:
This PR (released in 2.4.2) breaks existing TypeScript projects, because `setProps` is too strict. I see this has already been addressed in https://github.com/vuejs/test-utils/commit/e9fe51fb10d265e4bcb25d5d4a0ea8474f244fc4 by making it `Partial`. When can we expect this fix to be released?

PR #2161, opened 2023-08-17
docs: fix spell Exemple to Example

yasuaki640:
I found what appears to be a spelling error in `docs/api/index.md`.

In French, Ex**e**mple is correct, but in English, Ex**a**mple would be a more appropriate spelling.


cexbrayat (approved):
Good catch, thanks!

PR #2166, opened 2023-08-22
Fixing prototype methods being discarded when using `setData`

rory-instil:
Currently when calling `setData` on a wrapper it will not retain methods on any constructed objects passed in. These methods reside in the prototype which the current logic do not persist across. These changes will ensure that any prototype properties / methods are copied over to the proxied data object.

fixes https://github.com/vuejs/test-utils/issues/1851

### DIFF
--- a/tests/setData.spec.ts
+++ b/tests/setData.spec.ts
@@ -218,7 +218,8 @@ describe('setData', () => {
   it('should retain prototype methods for constructed objects when calling setData', async () => {
     const expectedResult = 'success!'
     class TestClass {
-      getResult() {
+      constructor(readonly name: string) {}
+      getResult(): string {
         return expectedResult
       }
     }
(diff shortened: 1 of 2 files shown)


cexbrayat (approved):
LGTM, thanks @rory-instil 👍

tests/setData.spec.ts
```diff
@@ -214,4 +214,35 @@ describe('setData', () => {
...
+      getResult() {
+        return expectedResult
+      }
+    }
+
+    const wrapper = mount(
+      defineComponent({
+        template: '<div/>',
+        data() {
+          return { value: new TestClass() }
```
cexbrayat, line 230: Maybe the initial value should be something else to see if `setData` properly updates it no?

cexbrayat (changes requested):

tests/setData.spec.ts
rory-instil: As in a different class altogether? 
Or just with some properties on the class that get changed?

rory-instil: for example:
```
it('should retain prototype methods for constructed objects when calling setData', async () => {
  const expectedResult = 'success!'
  class TestClass {
    name: string
    constructor(name: string) {
      this.name = name
    }
    getResult(): string {
      return expectedResult
    }
  }

  const wrapper = mount(
    defineComponent({
      template: '<div/>',
      data() {
        return { value: new TestClass('test1') }
      },
      methods: {
        getResult() {
          return `${this.value.name}: ${this.value.getResult()}`
        }
      }
    })
  )

  expect(wrapper.vm.getResult()).toStrictEqual(`test1: ${expectedResult}`)

  await wrapper.setData({
    value: new TestClass('test2')
  })

  expect(wrapper.vm.getResult()).toStrictEqual(`test2: ${expectedResult}`)
})
```

cexbrayat (approved):

cexbrayat:
Thanks for the help @rory-instil !

aleksi-magner:
И как теперь с этими изменениями проверять соответствие?
изображение

PR #2180, opened 2023-09-02
Fix: calling element method instead of trigger DOM event

### LINKED ISSUE #2176: Bug: focus active
**Describe the bug**

when i test the focs event, the document.activeElement is the full `body`, like this:
image

**To Reproduce**

https://stackblitz.com/edit/github-3auzsz?file=src%2Fcomponents%2F__tests__%2FHelloWorld.spec.ts,src%2Fcomponents%2FHelloWorld.vue

**Expected behavior**

the `document.activeElement` is `wrapper.element`

**Related information:**

"@vue/test-utils": "^2.4.1"
"jsdom": "^22.1.0"
"vitest": "^0.34.3"

**Additional context**

freakzlike:
Calls `element.click()`, `element.focus()`, etc. on `trigger('click')` instead of dispatching a DOM event.

I was able to get it working, but I'm not really convinced of the implementation.

Fix #2176

### DIFF
--- a/src/baseWrapper.ts
+++ b/src/baseWrapper.ts
@@ -376,7 +376,8 @@ export default abstract class BaseWrapper<ElementType extends Node>
       if (typeof element[eventString] === 'function') {
         element.addEventListener(
           eventString,
-          (event) => {
+          (nativeEvent) => {
+            const event = nativeEvent as Event & { _vts: number }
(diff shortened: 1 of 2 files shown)


cexbrayat (commented):
Even if I appreciate the creative way you solved the problem (👏), I'm not sure it's worth the added complexity. I'm fine if we merge it or not.

PR #2189, opened 2023-09-19
Add WebdriverIO to list of supporting frameworks

christian-bromann:
WebdriverIO has a browser testrunner allowing to run component tests on real browser revisions and not just browser engines. We added docs on how to use `@vue/test-utils` and wonder if it makes sense to add WebdriverIO to the list of supporting frameworks.


cexbrayat (approved):

cexbrayat:
Thanks @christian-bromann 👍

PR #2197, opened 2023-09-25
fix: don't require optional peerDependencies

thebanjomatic:
`@vue/server-renderer` and `@vue/compiler-dom` were both required in the
sense that they will throw regardless of whether or not your are using
functionality from those libraries if they weren't installed. I don't
believe this was intentional, so I have made them optional and they will
now only throw if you try to use them and the package isn't available,
and will print a more helpful error message when doing so.

### DIFF
--- a/src/utils/requireOptionalPeer.ts
+++ b/src/utils/requireOptionalPeer.ts
@@ -1,0 +1,27 @@
+import { createRequire } from 'module'
+
+// @ts-ignore - cjs build will complain about import.meta not being allowed, but its not actually used as __filename will be defined for cjs
+const file = typeof __filename !== 'undefined' ? __filename : import.meta.url
+const require = createRequire(file)
+
+export function requireOptional(packageName: string): unknown {
+  try {
+    return require(packageName)
+  } catch {
+    throw new Error(
+      `The optional peer-dependency ${packageName} is needed for this test and was not found. Please ensure that it has been installed as a dev-dependency and retry.`
+    )
+  }
+}
+
+export function requireVueCompilerDom() {
+  return requireOptional(
+    '@vue/compiler-dom'
+  ) as typeof import('@vue/compiler-dom')
+}
+
+export function requireVueServerRenderer() {
+  return requireOptional(
+    '@vue/server-renderer'
+  ) as typeof import('@vue/server-renderer')
+}
--- a/package.json
+++ b/package.json
@@ -65,10 +65,14 @@
     "vuex": "4.1.0"
   },
   "peerDependencies": {
+    "@vue/compiler-dom": "^3.0.1",
     "@vue/server-renderer": "^3.0.1",
     "vue": "^3.0.1"
   },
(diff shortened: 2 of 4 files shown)


src/utils/requireOptionalPeer.ts
```diff
@@ -0,0 +1,27 @@
+import { createRequire } from 'module'
+
+// @ts-ignore - cjs build will complain about import.meta not being allowed, but its not actually used as __filename will be defined for cjs
+const file = typeof __filename !== 'undefined' ? __filename : import.meta.url
+const require = createRequire(file)
```
thebanjomatic, line 5: I opted for using a require here as its behavior is synchronous and I believe the functions where these were being called from were also synchronous and not returning promises. Ideally, we would use a dynamic `import` for the esm code, and a `require` for the cjs code, but introducing the dynamic import would change the signatures of these functions. Let me know if you would like me to pursue this.

lmiller1990: I am not sure what the best way to proceed here is, but this seems really heavy handed to make a breaking change and add all this complexity to support one specific package manager.

lmiller1990:
How can I reproduce and test the issue this supposedly fixes? I am not comfortable merging this as it - I don't know what it fixes or how to test it.

thebanjomatic:
I'll see if I can create a small reproducer.

src/utils/requireOptionalPeer.ts
thebanjomatic: Functionally the require method being used should be fine and match existing behavior.

thebanjomatic:
I created a test repo that demonstrates 3 separate issues across yarn, pnpm and npm that are related to the issue in this PR. I was really hoping to just see the same problem manifest itself in the same way 3 times, but funny enough each package manager seems broken in slightly different ways.

https://github.com/thebanjomatic/test-utils-repro/

dschmidt:
Today (after updating test-utils) I ran into the issue fixed by this PR.
pnpm has stricter package isolation than classic npm or yarn, so when unconditionally importing the module without having it as a dependency, it will always fail.

For now my workaround is to add
```json
{
 …
 "pnpm": {
    "packageExtensions": {
      "@vue/test-utils": {
        "dependencies": {
          "@vue/server-renderer": "3.4.15"
        }
      }
    }
  }
  …
}
```

to my `package.json`.

nazarepiedady:
@thebanjomatic, do you believe that your pull request still necessary?

@lmiller1990, @cexbrayat, could you say if the issue this pull request is trying to solve still valid?

If the answer is not, I think we should close it to keep the repository clean with only real problems.

lmiller1990:
I am not sure. I am not deep enough in the package management ecosystem to know, and I have never had an issue using Vue / Test Utils with any of the projects I'm working on.

Since it looks like there isn't a pressing need for this and no-one else has raised issues, let's just close it for now.

PR #2227, opened 2023-10-27
chore(deps): Set range for dependencies

freakzlike:
The `dependencies` packages are pinned to a fixed version and updated through renovate. However without a release it still uses the old versions on projects. I've changed them to use a version range and changed renovate config to don't pin the depdencies


cexbrayat (approved):
Makes sense 👍

PR #2231, opened 2023-11-03
fix: wrapper.text method

harunari0928:
### Summary

This change fixes wrapper.text to handle suspense components with multiple elements in a slot.
When suspense component had multiple elements in a slot, wrapper.text returned empty.


lmiller1990 (approved):
Great!

PR #2234, opened 2023-11-08
chore: use experimentalVmThreads to run unit tests

cexbrayat:
Tests are roughly twice faster with this option.

Before:
```
Duration  7.58s (transform 859ms, setup 441ms, collect 8.55s, tests 2.61s, environment 15.28s, prepare 3.19s)
```

After:
```
Duration  3.06s (transform 1.03s, setup 344ms, collect 7.36s, tests 2.08s, environment 1.58s, prepare 4.25s)
```

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -24,6 +24,7 @@ export default defineConfig({
   },
   test: {
     environment: 'jsdom',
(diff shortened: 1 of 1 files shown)


cexbrayat:
CI builds are twice faster as well

Issue #2251, opened 2023-11-21
Bug: Wrapper.Text function is trimming elements


occitaneUbald:
**Describe the bug**
Text function is trimming elements when multiple elements are concatenated

**To Reproduce**
This example works in 2.4.1 but does not work in 2.4.2
```
it('does not work', () => {
    const wrapper = shallowMount({
        template: `<span>Trimmed </span><span>Example</span>`
    })

    expect(wrapper.text()).toBe('Trimmed Example')
})
```

**Expected behavior**
This test should pass as in 2.4.1 version

cexbrayat:
Hi @occitaneUbald 

This is because of this fix in VTU https://github.com/vuejs/test-utils/pull/2231
But I think the new behavior is probably better.

You'll have to update your test to either:
- check with `TrimmedExample`
- check both span elements individually with`element.textContent` if you want to check if the space is there (because `text()` does trim the content).

occitaneUbald:
You should at least mention this in your changelog as it is a breaking change in your API

cexbrayat:
Maybe we should have, but I don't think it was spotted by the contributor who did this fix.

Feel free to open a PR to open the changelog, we'll gladly review it.

occitaneUbald:
And we should probably add an automated test to avoid further undetected regressions in the future on this API

PR #2253, opened 2023-11-22
Do not trim text content of the element

occitaneUbald:
This is a fix for the regression found in #2251 

I removed the trim that was used inside the textContent function. I think it is more understandable for people as it is not doing unexpected things and better matches the description of the documentation : https://test-utils.vuejs.org/api/#text
I also added a test reproducing the regression mentionned in #2251 so that it avoid coming back in the future.

Obviously this pull request is a breaking change for people that were relying on this trim logic (that was not documented anyway)

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -128,9 +128,7 @@ export function isObjectComponent(
 export function textContent(element: Node): string {
   // we check if the element is a comment first
   // to return an empty string in that case, instead of the comment content
-  return element.nodeType !== Node.COMMENT_NODE
-    ? element.textContent?.trim() ?? ''
-    : ''
+  return element.nodeType !== Node.COMMENT_NODE ? element.textContent ?? '' : ''
 }
 
 export function hasOwnProperty<O extends {}, P extends PropertyKey>(
--- a/tests/components/FunctionComponent.tsx
+++ b/tests/components/FunctionComponent.tsx
@@ -4,7 +4,7 @@ interface Props {
   title: string
 }
 const Title: FunctionalComponent<Props> = ({ title }) => {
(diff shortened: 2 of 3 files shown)


cexbrayat (changes requested):
Thanks for the PR.

I don't think we can remove `trim` here: it has been there since the very beginning of VTU v2 and it was there in VTU v1 since the very beginning as well if I remember correctly.

So it would be too breaking to change this.

What I was suggesting in the discussion of https://github.com/vuejs/test-utils/issues/2251 was to open a PR to add a mention in the changelog that when calling `text()` on several nodes, they are now all trimmed and that some tests may need to be updated when bumping to VTU v2.4.2 (sorry if I wasn't clear)

occitaneUbald:
I don't know how to update the changelog because it is not available in the codebase, it is only in the release declaration. Btw it would be nice to update the documentation if this is a wanted behavior but I don't understand why it would be wanted.
I was expecting a lot more tests to fail related to this change, as it had little to no impact, I thought it was the best solution. 
I'm also not really fan of doing unexpected things in the back of users of the library.

I'm okay with not taking this change (because it is a breaking change that could annoy a lot of people), but in this case, we should change :
* the documentation to explain the trim behavior
* the changelog to mention the regression
* consider changing the name of the function from `text()` to `trimmedText()` (verbose but explicit)

cexbrayat:
I'll update the release notes then.

If you feel like it, you can open a PR to add a mention of the trim behavior in the documentation for the text method.

I don't think it's worth renaming the method: the text content has been trimmed for the past 7 years, so it wouldn't make much sense to rename it now as it would force thousands of projects to rename it in every test 😉.

Issue #2257, opened 2023-11-24
Bug: `setData()` incorrectly sets object prototype methods as keys


MetRonnie:
**Describe the bug**

Using `setData()` with an object is causing Object prototype methods to be set as keys on the object.

**To Reproduce**

Stackblitz isn't loading for me. But here is hopefully a clear example:

```js
const wrapper = mountFunction()
expect(wrapper.vm.filters).toEqual({
  'workflow state': [],
  'task state': [],
}) // ✔
expect(Object.keys(wrapper.vm.filters)).toEqual([
  'workflow state',
  'task state',
]) // ✔

await wrapper.setData({
  filters: {
    'workflow state': [],
    'task state': [],
  }
})
expect(Object.keys(wrapper.vm.filters)).toEqual([
  'workflow state',
  'task state',
]) // ✖
```

```diff
- Expected
+ Received

  Array [
    "workflow state",
    "task state",
+   "constructor",
+   "__defineGetter__",
+   "__defineSetter__",
+   "hasOwnProperty",
+   "__lookupGetter__",
+   "__lookupSetter__",
+   "isPrototypeOf",
+   "propertyIsEnumerable",
+   "toString",
+   "valueOf",
+   "toLocaleString",
```

**Expected behavior**

Object prototype methods should not be set as keys on objects.

**Related information:**

```
    "vue": "3.3.4",
    "@vue/test-utils": "2.4.2",
    "vitest": "0.34.6"
```

**Additional context**

I think this was introduced by #2166

cexbrayat:
Hi @MetRonnie 

Interesting. You're probably right that this is a regression from #2166

Would you like to open a PR to fix this? We would gladly review and merge it.

MetRonnie:
I'm afraid I don't have a lot of time on my hands

cexbrayat:
No worries, maybe someone will pick it up.

Haberkamp:
I'll create a PR for that issue during next days.

PR #2265, opened 2023-11-30
fix: ignore prototype methods when using setData on objects

Haberkamp:
This change ignores the prototype methods of an object when using `setData`

resolves: #2257

### DIFF
--- a/tests/setData.spec.ts
+++ b/tests/setData.spec.ts
@@ -247,7 +247,8 @@ describe('setData', () => {
     expect(wrapper.vm.getResult()).toStrictEqual(`test2: ${expectedResult}`)
   })
 
-  it('should be possible to replace a primitive value with another', async () => {
+  // https://github.com/vuejs/test-utils/issues/2257
+  it('should ignore prototype methods when using setData on objects', async () => {
     const wrapper = mount(
       defineComponent({
         template: '<div />',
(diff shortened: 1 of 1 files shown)


tests/utils.spec.ts
```diff
@@ -0,0 +1,105 @@
+/**
+ * @vitest-environment node
+ */
+
+import { describe, expect, it } from 'vitest'
+import { mergeDeep } from '../src/utils'
+
+describe('utils', () => {
+  it('should be possible to replace a primitive value with another', () => {
+    // ARRANGE
```
cexbrayat, line 10: nit: you can remove the `// ARRANGE`, `// ACT`, `// ASSERT` comments as we don't use these in other files.

tests/utils.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+      bar: []
+    })
+
+    // ASSERT
+    expect(result).toStrictEqual({
+      foo: [],
+      bar: []
+    })
+  })
+})
```
cexbrayat, line 105: I think there should be more tests to check if there are no issues with classes, prototype methods, etc.

Maybe you can add them here, or in the `setData.spec.ts` file.

A test that reproduces #2257 would be especially useful

cexbrayat (changes requested):

tests/utils.spec.ts
Haberkamp: I've removed the `utils.spec.ts` file I've created and added a test case that tests the bug of the issue (#2257) and put that into the `setData.spec.ts` file. That file also contains a tests case for checking prototype inheritance.

tests/setData.spec.ts
```diff
@@ -246,4 +246,33 @@ describe('setData', () => {
...
+            firstArray: [],
+            secondArray: []
+          }
+        }
+      })
+    )
+
+    await wrapper.setData({
+      firstArray: [],
+      secondArray: []
```
cexbrayat, line 265: Maybe use different values than the original ones, like `firstArray: [1, 2], secondArray: [3, 4]`

cexbrayat (changes requested):

cexbrayat (approved):
Awesome, thanks @Haberkamp 👍

doutatsu:
When will the new version be released, with this fix? Need this fix, as I use a lot of `Object.keys` and as a result my tests are failing

PR #2269, opened 2023-12-05
fix: always load cjs bundle in node environment

danielroe:
at the moment the subpath import condition for `import` loads a 'bundler' version which expects a browser environment (it references `document`). But node running in a native ESM context will also resolve this condition.

To fix simply, we can inject a `node` condition in front of it which will force node to use the CJS version.

Longer term, I would recommend _also_ producing an ESM version that does not rely on a browser context.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
   "exports": {
     ".": {
       "types": "./dist/index.d.ts",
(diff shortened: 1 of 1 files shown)


cexbrayat (approved):
These packaging quirks are pure madness, thanks for the PR @danielroe

PR #2327, opened 2024-02-04
fix(findComponent): Use correct vm for stubbed component with selector

### LINKED ISSUE #2300: Bug: findComponent doesn't work properly if child component is root of parent component
**Describe the bug**
This is known issue in vue test utils v1, but the previous workaround doesn't work anymore

```
<template>
  <WelcomeItem :test="2"> </WelcomeItem>
</template>
```
vs
```
<template>
  <div>
    <WelcomeItem :test="2"> </WelcomeItem>
  </div>
</template>
```

**To Reproduce**

https://stackblitz.com/edit/github-orekbd?file=src%2Fcomponents%2F__tests__%2FTheWelcome.spec.ts

run `yarn test:utils TheWelcome`

**Expected behavior**

`findComponent.props()` returns props of child component

**Related information:**

In vue test utils v1 there was workaround: 
```
const childWrapper = createWrapper(wrapper.vm.$children[0])
```
Now there is no `wrapper.vm.$children`

freakzlike:
It seems that using the vm from `currentComponent.subTree.component.proxy` is more accurate in that case instead of `currentComponent.proxy`

fix #2300

### DIFF
--- a/src/baseWrapper.ts
+++ b/src/baseWrapper.ts
@@ -175,7 +175,12 @@ export default abstract class BaseWrapper<ElementType extends Node>
       matches(currentComponent.vnode, selector) &&
       this.element.contains(currentComponent.vnode.el as Node)
     ) {
-      return createVueWrapper(null, currentComponent.proxy!)
+      return createVueWrapper(
+        null,
+        currentComponent.subTree.component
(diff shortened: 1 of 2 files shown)


cexbrayat (approved):
Thanks for looking into this @freakzlike 👍

PR #2340, opened 2024-02-13
fix(vue-router.md): update highlights when mocking real router

dmoyadev:
It seems like the highlighted parts of the code in the [Testing Vue Router](https://test-utils.vuejs.org/guide/advanced/vue-router.html#Using-a-Real-Router) page may have slightly move upwards.

This will fix all the highlights in the code blocks of this section to match what is said in the documentation.

### Before:
image

### After:
image


cexbrayat (commented):
LGTM, thanks.

Would you mind updating it in the french version as well?

cexbrayat (approved):

PR #2356, opened 2024-02-29
Allow access to nested computed values (fix #2196)

### LINKED ISSUE #2196: Computed ref unwrapping is different than in vue
### Subject of the issue

Hey, I'm working on a codebase where we use computed refs outside of vue components.
A store class has a computed ref (`store.count `in the stackblitz environment), an instance of this class is passed as prop to a vue component.

In Vue itself, `store.count` is a ComputedRefImpl, where you still need .value to access the value.
In Vue-test-utils `store.count` already is the value itself, and `store.count.value` will be undinefed

### Steps to reproduce

I have created [a stackblitz environment](https://stackblitz.com/edit/vitejs-vite-hqwxfz?file=src%2FApp.spec.js,src%2FApp.vue)
When viewing the ui, App.storeValue is `3`, and store.count is a ComputedRefImpl
If you kill the devserver and `npm test` App.storeValue is `NaN`, because store.count is the value itself, so store.count.value is undefined

### Expected behaviour

store.count should be a computedRefImpl which still needs `.value` to access the value

### Actual behaviour

store.count immediately contains the value, unlike in vue itself

### Possible Solution

?

Evobaso-J:
fix #2196


lmiller1990:
Thanks!

PR #2364, opened 2024-03-09
chore(deps): Remove peerDependencies

qmonmert:


lmiller1990 (approved):

Akryum:
Hello, this breaks installation in a big monorepo if there are multiple vue versions for different parts of the monorepo.

For example here using pnpm, vue test utils is incorrectly installed in a `@vue+test-utils@2.4.6` folder:

image

This causes the `require('vue')` to resolve to a potentially incorrect version of vue and can lead to multiple versions of vue being evaluated in vitest:
image

Pinning vue test utils to 2.4.4 solves this as pnpm is able to install the correct versions of vue-test-utils and vue so they are resolved correctly:

image

Which makes the test pass (no duplicate vue):

image

PR #2375, opened 2024-03-20
fix: update attachTo type in MountingOptions interface

### LINKED ISSUE #2194: Feature request: Attach to SVG elements

I was testing a component that renders an SVG element (e.g. `<rect>`), but when I tried to attach it to another SVG element during mount I noticed that my typescript type checker claimed that `attatchTo` must be either a `HTMLElement` or a `string`. Reading the code I noticed that `mount` will always create a `<div>` wrapper element, even when I ignore my type checker and provide an SVG element to the `attacheTo` option.

Here is an example of a mounted SVG element which goes into an HTML element that is inside another SVG element, resulting in this weird structure.

```js
const svg = document.createElementNS("http://www.w3.org/2000/svg", "svg");
document.body.append(svg);

const wrapper = mount(RectComponent, { attachTo: svg });
// error TS2322: Type 'SVGSVGElement' is not assignable to type 'string | HTMLElement | undefined'.
```

```html
<svg>
  <div data-v-app="">
    <rect x="0" y="0" width="10" height="10" ></rect>
  </div>
</svg>
```

It would be nice if `mount` would be smarter about which element it wraps around the mounted component. If the `attachTo` element is an SVG element, it should perhaps wrap in a `<g>` instead of a `<div>`. Similarly if the `attachTo` element is a MathML element, perhaps it should wrap it in an `<mrow>`.

We could also allow to configure the wrapper element factory, so instead of just blindly creating a `div` we could accept a function which creates the desired element.

Or we could simply change the typing of the `attachTo` from `HTMLElement | string` to `Element | string` and leave everything else as is. Having this weird HTML/SVG structure is unlikely to yield any issues inside an isolated testing environment.

**Additional context**

taku-y-9308:
fix #2194

Changed the type of attachTo in MountingOptions from HTMLElement to Element.
Do I need to update documentation?

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -14,7 +14,7 @@ Note that when mocking dates/timers with Vitest, this must be called after
 
 ```ts
 interface MountingOptions<Props, Data = {}> {
-  attachTo?: HTMLElement | string
+  attachTo?: Element | string
   attrs?: Record<string, unknown>
   data?: () => {} extends Data ? any : Data extends object ? Partial<Data> : any
   props?: (RawProps & Props) | ({} extends Props ? null : never)
@@ -74,7 +74,7 @@ Specify the node to mount the component on. This is not available when using `re
 **Signature:**
 
 ```ts
-attachTo?: HTMLElement | string
+attachTo?: Element | string
 ```
 
 **Details:**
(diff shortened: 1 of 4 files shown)


cexbrayat (changes requested):
Thanks for the PR @taku-y-9308 !

Yes it would be great to update the docs [here](https://test-utils.vuejs.org/api/#mount) (both in English and in French), and to add a test in `attachTo.spec.ts` with an SVGElement element for example.

cexbrayat (approved):
Awesome, thanks @taku-y-9308

PR #2386, opened 2024-03-26
change vm to always provide global property

### LINKED ISSUE #2140: Bug: vm does not provide access to global properties for Composition API component
**Describe the bug**

I sometimes use the `vm` property of a Wrapper to access global properties. This comes up most often in tests that use a real router, then make an assertion about `vm.$route`. However, it seems that `vm` does not provide access to global properties for Composition API components.

**To Reproduce**

https://github.com/matthew-white/global-properties-repro/blob/master/src/components/__tests__/HelloWorld.spec.ts

In this repository, there are 4 variations on a component. A plugin adds a global property named `foo`. For 2 of the components, `vm.foo` is defined, but for the other 2, `vm.foo` is not defined. This is what I see when I run tests:

✓ provides access to a global property from an Options API component
× provides access to a global property from a Composition API component
× provides access to a global property from an Options API component with a setup() function
✓ provides access to a global property from an Options API component with a setup() function that does not return

**Expected behavior**

It would be great to be able to access global properties from `vm` for all components.

**Related information:**

  npmPackages:
    @vue/test-utils: ^2.4.1 => 2.4.1
    vitest: ^0.33.0 => 0.33.0
    vue: ^3.3.4 => 3.3.4

**Additional context**

Just to link issues, I think this issue may be related to #2008.

taku-y-9308:
resolve #2140

Changed createVMProxy function to also return global properties.

### DIFF
--- a/tests/mountingOptions/global.plugins.spec.ts
+++ b/tests/mountingOptions/global.plugins.spec.ts
@@ -2,6 +2,10 @@ import { describe, expect, test, it, vi } from 'vitest'
 import { h, App } from 'vue'
 
 import { mount } from '../../src'
+import ScriptSetup from '../components/ScriptSetup.vue'
+import Option from '../components/OptionComponent.vue'
+import OptionsSetup from '../components/OptionSetupComponent.vue'
+import OptionsSetupWithoutReturn from '../components/OptionSetupWithoutReturnComponent.vue'
 
 describe('mounting options: plugins', () => {
   it('installs a plugin via `plugins`', () => {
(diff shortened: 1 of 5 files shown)


tests/mountingOptions/global.plugins.spec.ts
```diff
@@ -51,6 +55,41 @@ describe('mounting options: plugins', () => {
 
     expect(installed).toHaveBeenCalledWith(options, testString)
   })
+
+  describe('provides access to a global property', () => {
```
taku-y-9308, line 59: If you have a more appropriate name, please let me know 🙏

cexbrayat (approved):
LGTM, thanks @taku-y-9308 👍

PR #2399, opened 2024-04-05
docs(api): fix missing chars

w2xi:
The example of the [shallow](https://test-utils.vuejs.org/api/#shallow) API misses the suffix "stub" when making assertions. 

`Component.spec.js`
```js
import Component from './Component.vue'

test('shallow', () => {
  const wrapper = mount(Component, { shallow: true })

  expect(wrapper.html()).toEqual(
    `<a-component-stub></a-component-stub><another-component></another-component>`
  )
}
```
should change to:
```js
import Component from './Component.vue'

test('shallow', () => {
  const wrapper = mount(Component, { shallow: true })

  expect(wrapper.html()).toEqual(
    `<a-component-stub></a-component-stub><another-component-stub></another-component-stub>`
  )
}
```

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -829,7 +829,7 @@ test('shallow', () => {
   const wrapper = mount(Component, { shallow: true })
 
   expect(wrapper.html()).toEqual(
-    `<a-component-stub></a-component-stub><another-component></another-component>`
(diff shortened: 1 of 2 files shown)


cexbrayat (approved):
Thanks 👍

PR #2403, opened 2024-04-07
docs: use innerHTML in teleport cleanup

### LINKED ISSUE #2402: Docs: afterEach in teleport section causes issues on happy-dom
Currently https://test-utils.vuejs.org/guide/advanced/teleport documents:

```ts
afterEach(() => {
  // clean up
  document.body.outerHTML = ''
})
```

This removes whole body and `document.body` becomes `null` when using `happy-dom` (in `jsdom` it remains fine), resulting in errors like these:

```
TypeError: Cannot set properties of null (setting 'outerHTML')
 ❯ tests/components/STooltip.spec.ts:12:17
     10| 
     11| afterEach(() => {
     12|   document.body.outerHTML = ''
       |                 ^
     13| })
     14| 
```

Will it be better to use `document.body.innerHTML = ''` instead? It works on both happy-dom and jsdom. (Though happy-dom's behavior seems non-standard.)

brc-dd:
closes #2402

Updated `tests/docs-examples/teleport.spec.ts`; `tests/features/teleport.spec.ts` is still using `outerHTML` though.

### DIFF
--- a/docs/fr/guide/advanced/teleport.md
+++ b/docs/fr/guide/advanced/teleport.md
@@ -101,7 +101,7 @@ beforeEach(() => {
 
 afterEach(() => {
   // nous nettoyons un peu
-  document.body.outerHTML = '';
+  document.body.innerHTML = '';
 });
 
 test('teleport', async () => {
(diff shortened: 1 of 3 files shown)


cexbrayat (approved):
Thanks!

PR #2406, opened 2024-04-08
feat: Added dynamic return for element getter

nandi95:
I used to contribute and it seems I had this laying around for years now without committing. Small improvement for getting better type inference.

### DIFF
--- a/test-dts/wrapper.d-test.ts
+++ b/test-dts/wrapper.d-test.ts
@@ -14,6 +14,9 @@ const domWrapper = wrapper.find('#other')
 let inputMaybe = wrapper.find('input')
 expectType<HTMLInputElement | undefined>(inputMaybe.element)
 
+let buttonComp = mount(defineComponent({ template: '<button>Click me</button>' }))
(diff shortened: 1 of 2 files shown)


test-dts/wrapper.d-test.ts
```diff
@@ -14,6 +14,9 @@ const domWrapper = wrapper.find('#other')
 let inputMaybe = wrapper.find('input')
 expectType<HTMLInputElement | undefined>(inputMaybe.element)
 
+let buttonComp = mount(defineComponent({ template: '<button>Click me</button>' }))
+expectType<HTMLButtonElement>(buttonComp.element)
+
```
cexbrayat, line 19: Do we have a test for the case with multiple root elements?

nandi95: Added a type test and here's an actual test for it: https://github.com/vuejs/test-utils/blob/main/tests/element.spec.ts#L46

cexbrayat (approved):

PR #2423, opened 2024-04-26
fix: set global provides before running vue plugins

danielroe:
Happy to have some push-back on this, but encountered a situation in `@nuxt/test-utils` where we set default 'provides' but would like the user to be able to register their own plugins to override these (https://github.com/nuxt/test-utils/issues/828).

Reordering the provides before the plugins seems like a reasonable way to do this, but happy for other ideas, or to try to tackle this downstream instead if you don't like the change or it breaks something.

### DIFF
--- a/src/createInstance.ts
+++ b/src/createInstance.ts
@@ -266,6 +266,14 @@ export function createInstance(
     }
   }
 
+  // provide any values passed via provides mounting option
+  if (global.provide) {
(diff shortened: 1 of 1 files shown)


cexbrayat (approved):
Hi @danielroe 

That looks reasonable to me, thanks for the PR.

PR #2435, opened 2024-05-14
Update index.md to fix typos and clarify behaviors

KatWorkGit:
- Fixed typos (missing period)
-  Clarified behavior of `findComponent` vs `getComponent`.
- Added notes about `get` and when to use it in the `find` section so that if users only check the `find` section of the docs they will see when to use `find` vs `get` (as opposed to before when it was only specified in the `get` section of the docs.


cexbrayat (approved):
Thanks for the PR @KatWorkGit

PR #2441, opened 2024-05-27
add test cases for JSX components

julisch94:
This pull requests aims for fixing this issue: https://github.com/vuejs/test-utils/issues/2438

I added a failing unit test.
Unfortunately I didn't manage to fix the unit test and with that actually fix the issue I'm experiencing. Maybe someone knows how to finish this up?

What I'm missing is:
- My child component is not stubbed, because the transformer doesn't find any component for the given `type`.
- I don't know which one of these lines but at least one should find out the component name: https://github.com/julisch94/test-utils/blob/main/src/vnodeTransformers/stubComponentsTransformer.ts#L185-L186
- when I hardcode the name of my component there, it all works. So all I'm missing is:
- the transformer needs to be able to determine the component name inside my parent component
-  How can that be done?

Please feel free to help out!

### DIFF
--- a/tests/components/ComponentWithoutNameJsx.tsx
+++ b/tests/components/ComponentWithoutNameJsx.tsx
@@ -1,0 +1,7 @@
+import { defineComponent } from 'vue'
+
+export default defineComponent({
+  setup() {
+    return () => <div class="ComponentWithoutNameJsx">No Name JSX</div>
+  }
+})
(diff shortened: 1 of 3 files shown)


cexbrayat:
Hi @julisch94 

Thanks for taking a look!
Yes it looks like the only way to do this is to add a name to the component in JSX.
See https://github.com/vuejs/test-utils/blob/main/src/utils/componentName.ts 

Maybe this is something the vite plugin for jsx could improve by adding a `__name` property to the component based on its file name, as does the unplugin-vue-components? Then VTU could try to rely on that.
You can try to open a discussion there to see if this is doable (as I don't use JSX personally, I'm not even sure this is a good idea). I'll close the PR in the meantime, and we'll revive it if needed!

Issue #2468, opened 2024-07-09
Bug: setValue on nested v-model do not correctly update parent


tharvik:
**Describe the bug**
with nested v-model components, updating the child doesn't propagate changes to the parents, only the direct one.

**To Reproduce**
I've [a test on my fork](https://github.com/tharvik/test-utils/compare/main%5E1...tharvik:test-utils:main) that I can turn into a PR (wanted to write the issue first). the normal vue behavior can be testing in [this playground](https://play.vuejs.org/#eNqVUl1r2zAU/SsXMUgCmZ2x7cUkha30oXvYxlb2Uo1i7JtErSwJfaQB4//eK9lx0zYECn6wdI6Ozrk6LftmTLYLyAq2dJUVxl9wJRqjrYcWLK6hg7XVDUyINBmhy62Q9QBkeVpFFSJwhftEqXFdBkkqXAHUpS8LmM5gdQHTFhQ6j3UBKkgJ3WweKZUmaYXKu4Ju7i/oEtLooIhOp5MWgEN/TTt2V8pprzkAAH4rXNbLwwo4SxZNaUmXs57UzeHrYjGLi44r+pb5mJwWHhsjS4+0Alj2PnYfG12jXHHWS3MGOeHL/IjM5sy7Squ12GT3TiuaaHLFWUwmJNpfxgutHGcUsLfCWSmlfvyR9rwNmPKmM1usHk7s37t93OPst0VHE0DORsyXdoMUM8JXf3/inv5HkAIESewz4B90Wobosad9D6om20e85PY6VUCozY272ntU7hAqGk1TTXzOqBCXZ6I/2/2cfTm8Bk1xbNOLTp5qlbHaUFtuJ+l1/pWSGvg/XY6NiD26nQRD1cPiDeFUp25Egzr415U607ahb0n9uW5VDHBo20j6ED1NOXvjiOZ7JJN6OdT0091iKOo7W1uLXfoBaFuKergJOjpHcN7jL+vbPQEt5FDs
)

**Expected behavior**
in nested v-model, if the child changes, it should update all it's depending parents, as vue does.

**Related information:**
it look like that `setValue` replace the ref of a model by the value directly. it should update the ref instead.

**Additional context**
I was able to workaround it by replacing in my test stack
```ts
comp.setValue("value");
```
by
```ts
comp.vm.model.value = "value";
comp.getCurrentComponent().emit("update:modelValue");
```

PR #2469, opened 2024-07-09
chore: add test for nested v-model

tharvik:
reproduction as test for #2468, read there for context.

### DIFF
--- a/tests/setValue.spec.ts
+++ b/tests/setValue.spec.ts
@@ -295,17 +295,33 @@ describe('setValue', () => {
       template: '<div>{{ foo }} {{ bar }}</div>'
     })
 
+    const NestedInputComponentChild = defineComponent({
+      props: ['modelValue', 'onUpdate:modelValue'],
+      template: '<div>{{ modelValue }}</div>'
+    })
+    const NestedInputComponent = defineComponent({
+      props: ['modelValue', 'onUpdate:modelValue'],
+      template: '<NestedInputComponentChild v-model="modelValue" />',
+      components: { NestedInputComponentChild }
+    })
+
     const Component = defineComponent({
-      template:
-        '<PlainInputComponent v-model="plain" /><MultiInputComponent v-model:foo="foo" v-model:bar="bar" />',
+      template: `<PlainInputComponent v-model="plain" />
+         <MultiInputComponent v-model:foo="foo" v-model:bar="bar" />
+         <NestedInputComponent v-model="nested" />`,
       data() {
         return {
           plain: null,
           foo: null,
-          bar: null
+          bar: null,
+          nested: null
         }
       },
-      components: { PlainInputComponent, MultiInputComponent }
+      components: {
+        PlainInputComponent,
+        MultiInputComponent,
+        NestedInputComponent
+      }
     })
 
     describe('mount', () => {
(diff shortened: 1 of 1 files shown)


nazarepiedady:
@tharvik, I hope you're fine, could you edit your pull request to pass the CI?

tharvik:
> could you edit your pull request to pass the CI?

this PR is meant to fail, it servers to show the issue underlined in #2468.

nazarepiedady:
@cexbrayat, @lmiller1990, is there already any solution for this problem?

cexbrayat:
@nazarepiedady No, I think Lachlan and I don't really have time to look into it, so you're welcome to try!

nazarepiedady:
> @nazarepiedady No, I think Lachlan and I don't really have time to look into it, so you're welcome to try!

@cexbrayat, I will try it as soon as I finish something I am doing now.

lmiller1990:
Good to have a reproduction for a known bug.

I think using `findComponent` is generally not a good idea - your test is now coupled to your component hierarchy, which means you are testing implementation details.

If someone wants to fix this, that would be great - this is the API we support, after all. I suspect you will need some hacks to do it, further complicating the code base. This indicates this kind of test and API is probably not the best idea.

Context for my opinion: I am working on a code base right now that has hundreds of these types of tests (`findComopnent`, `setValue`, etc). It is impossible to refactor or change anything, a lot of time and money was wasted both writing and maintaining this style of test. I got hired specifically for my experience on this kind of project.

tharvik:
any update on that? I came back today to bite me again and the workaround I was using is not working for deeper v-model.

nazarepiedady:
@lmiller1990, @cexbrayat, in your opinion, is a good idea to close this pull request and let a reminder on the documentation about why this is not a good idea?

nazarepiedady:
> any update on that? I came back today to bite me again and the workaround I was using is not working for deeper v-model.

Did you understand the point presented by @lmiller1990?

tharvik:
> > any update on that? I came back today to bite me again and the workaround I was using is not working for deeper v-model.
> 
> Did you understand the point presented by @lmiller1990?

I did.
I didn't wanted to clutter the discussion with opinions but here goes.

> I think using `findComponent` is generally not a good idea - your test is now coupled to your component hierarchy, which means you are testing implementation details.

yeah, I agree, and totally get that it makes it harder to change the implementation. but as I want to test a component wrapping a form made of many custom fields. I can either set the value of every single field, or just set the resulting form model which is way more readable. 

> If someone wants to fix this, that would be great - this is the API we support, after all. I suspect you will need some hacks to do it, further complicating the code base. This indicates this kind of test and API is probably not the best idea.

it all depends on the needs of projects. if `setValue` was to be removed, I'll probably workaround it by setting `modelValue` myself. I think that the API is helpful as to make the tests more readable.

> @lmiller1990, @cexbrayat, in your opinion, is a good idea to close this pull request and let a reminder on the documentation about why this is not a good idea?

well, I find it pretty undelicate to tell devs that's their code is badly though of. they are the most knowledgable on their codebase and what are its specific needs.
also, closing a PR triggering a bug is akin to putting it under the rug: it will come back.

nazarepiedady:
@tharvik, is there any specification or documentation that says that this interface should work the way you're pointing?

nazarepiedady:
@cexbrayat, @lmiller1990 is @tharvik, right?

Is this a bug?

PR #2495, opened 2024-08-15
fixed incorrect checkbox value check

Psingh132:
This PR fixes an issue where the code incorrectly treated wrapper.find('div') as a boolean value.

**Changes Made**
Replaced expect(wrapper.find('div')).toBe(true) with expect(wrapper.find('div').exists()).toBe(true)


cexbrayat (approved):
Thanks @Psingh132 👍

PR #2498, opened 2024-08-22
chore: add migration step for deprecated ref syntax in findAllComponents

lorenzvanherwaarden:
The `ref` syntax is indeed not working, see https://gitlab.com/groups/gitlab-org/-/epics/14954, so I figured this could be part of the migration guide from v1 to v2.

### DIFF
--- a/docs/migration/index.md
+++ b/docs/migration/index.md
@@ -276,6 +276,28 @@ describe('App', () => {
   })
 ```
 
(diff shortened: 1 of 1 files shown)


lorenzvanherwaarden:
@cexbrayat could you review this one?

cexbrayat (approved):
Thanks!

PR #2501, opened 2024-08-26
Fix/issue 2462 - Vue is not defined in projects using Webpack and in-browser testing solutions

### LINKED ISSUE #234: ReferenceError: Vue is not defined
Hi! Today I'm using karma + ( mocha, chai ) unit test in Vue 3, but I've got a reference issue: ` ReferenceError: Vue is not defined`

I think it's not a configuration problem and `Vue` global object should be removed in Vue 3, so maybe is VTU itself:
```shell
  ReferenceError: Vue is not defined
      at Object.<anonymous> (webpack:///node_modules/@vue/test-utils/dist/vue-test-utils.browser.js:2407:5 <- specs/Hello.spec.js:13561:7)
      at Object.<anonymous> (specs/Hello.spec.js:13563:30)
      at __webpack_require__ (webpack:///webpack/bootstrap:19:1 <- specs/Hello.spec.js:20:30)
      at Module.<anonymous> (webpack:///test/unit/specs/Hello.spec.js:1:1 <- specs/Hello.spec.js:95:73)
      at __webpack_require__ (webpack:///webpack/bootstrap:19:1 <- specs/Hello.spec.js:20:30)
      at webpack:///webpack/bootstrap:83:1 <- specs/Hello.spec.js:84:18
      at specs/Hello.spec.js:87:10
```
for more details you can check out and clone my repo (you may ignore the `.yarnrc` by the way): https://github.com/Styx11/vue3-boilerplate/tree/v0.0.1

Thanks : )

### LINKED ISSUE #2462: Bug: ReferenceError: Vue is not defined - Missing exports declaration ?
**Describe the bug**

Hello, I'm kind of reopening #234 here as I'm having some trouble upgrading a software to Vue3 which was running completely fine in Vue 2 with Karma test runner and Vue Test Utils V1.x latest version.

Once everything is updated, unit tests can't run anymore and a `ReferenceError: Vue is not defined` is thrown.
I tried to follow what was written in this post and investigate the different possibilities and found that the easiest way around was to simply add a "module" exports inside the `package.json` of VTU like this:

image

Is there a good reason not to declare the module exports type here ?  I'm not an expert at all in bundling so I don't know if there could be some bad side effects... If not, could it be inserted here ?  I'm O.K. to create a simple PR with this if needed...

To give more information on [the first reply that was given by @lmiller1990 at that time](https://github.com/vuejs/test-utils/issues/234#issuecomment-720868125), `Karma` bundles the unit tests, the project source code and its test runner using `Webpack` then executes the result inside a browser directly and not using node and JSDom as lots of "modern" test runners do. The `target` parameter inside webpack config file is set to default (so `browserslist` or `web`).

Thanks in advance ;-)

sronveaux:
This PR resolves #2462 

Should also resolve #234 even though a bit late now ;-)

Simply inverts `browser` and `require` exports definition inside `package.json` so that `Webpack`-powered projects running tests in a browser context (and not in Node) take the correct build while bundling.  
See #2462 for detailed discussions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,8 +11,8 @@
       "types": "./dist/index.d.ts",
       "node": "./dist/vue-test-utils.cjs.js",
       "import": "./dist/vue-test-utils.esm-bundler.mjs",
-      "browser": "./dist/vue-test-utils.browser.js",
       "require": "./dist/vue-test-utils.cjs.js",
+      "browser": "./dist/vue-test-utils.browser.js",
       "default": "./dist/vue-test-utils.cjs.js"
     },
     "./package.json": "./package.json"


cexbrayat (approved):
Thanks for the PR 👍

PR #2519, opened 2024-09-28
import missing DOMWrapper on Implementation of Data Test ID Plugin

BrendonHenrique:
just followed the implementation of [Data Test ID Plugin section guide ](https://test-utils.vuejs.org/guide/extending-vtu/plugins.html#Data-Test-ID-Plugin)and notice that actually the DOMWrapper was not being imported.

image


cexbrayat (approved):
Thanks

PR #2548, opened 2024-11-20
fix: resolve warnings in docs:dev

DDDDD12138:
### Description

This modification aims to resolve warnings encountered when running `docs:dev` :
image

We are now using Vue 3.5+, so `withDefaults` is [unnecessary](https://vuejs.org/api/sfc-script-setup.html#defineprops-defineemits).

### DIFF
--- a/tests/components/PropWithSymbol.vue
+++ b/tests/components/PropWithSymbol.vue
@@ -1,5 +1,5 @@
 <template>
-  <div>Symbol: {{ $props.sym }}</div>
+  <div>Symbol: {{ sym }}</div>
 </template>
 
 <script lang="ts" setup>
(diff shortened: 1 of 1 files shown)


cexbrayat (approved):

tests/components/PropWithSymbol.vue
```diff
@@ -1,13 +1,7 @@
...
 
 <script lang="ts" setup>
-interface Props {
-  sym?: Symbol
-}
-
-withDefaults(defineProps<Props>(), {
-  sym: () => Symbol()
-})
+defineProps({ sym: { type: Symbol, default: () => Symbol() } })
```
cexbrayat, line 6: Now I think about it, can't we keep the Ts syntax and use the destructuration props with default here?

cexbrayat (approved):
Thanks!

PR #2549, opened 2024-11-20
fix: resolve TypeScript type errors in .vitepress/config

DDDDD12138:
### Before
There are some TypeScript type errors in the `.vitepress/config` directory ：
image

### Description
This PR addresses and resolves TypeScript type errors found in the `.vitepress/config` directory.
And the `collapsable` configuration has been removed in this [vitepress PR](https://github.com/vuejs/vitepress/pull/1865/files#diff-08d1bb1c444a76199b8f44b087883f932a6f8118f5ded25bf5c508fa52aa5b7f), replaced by the new property `collapsed`, so I removed the unnecessary `collapsable` configuration.

### DIFF
--- a/docs/.vitepress/config/en.mts
+++ b/docs/.vitepress/config/en.mts
@@ -1,10 +1,10 @@
-import type { DefaultTheme, LocalSpecificConfig } from 'vitepress'
+import type { DefaultTheme, LocaleSpecificConfig } from 'vitepress'
 
 export const META_URL = ''
 export const META_TITLE = 'Vue Test Utils'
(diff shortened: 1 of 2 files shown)


cexbrayat (commented):
👍 

Would you mind adding a check on Ci so we can catch these warnings in the future?
I suppose running `tsc` in the docs directory would catch this.

DDDDD12138:
Let me give it a try.

DDDDD12138:
> Would you mind adding a check on Ci so we can catch these warnings in the future? I suppose running `tsc` in the docs directory would catch this.

I've added it. Considering that the theme/index.mts file imports `.vue` files, I used `vue-tsc`.

PR #2579, opened 2024-12-31
fix: text() misses content for array functional component

miguelrincon:
Fix for https://github.com/vuejs/test-utils/issues/2568

This change adds a new optional parameter to the DOMWrapper constructor so information about its structure can be later used to produce correct `text()` and `html()` assertions.

### DIFF
--- a/src/baseWrapper.ts
+++ b/src/baseWrapper.ts
@@ -1,3 +1,4 @@
+import { VNode } from 'vue'
 import { textContent } from './utils'
 import type { TriggerOptions } from './createDomEvent'
 import {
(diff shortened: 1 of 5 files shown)


miguelrincon:
@cexbrayat This is an initial idea to fix https://github.com/vuejs/test-utils/issues/2568, all tests are passing but I hacked a bit around the types and method signatures. I'll need help to get those fixed ;)

I am also not sure if changing the `DOMWrapper` constructor to take a `vnode` it's a better fit, as the `vnode` contains more information we may need (rather than just the new subtree param), seems like a bigger change than I'd like to take on.

cc @xanf

cexbrayat (approved):
Thanks @miguelrincon for the PR, sorry it took us so long to review.

This looks like a nice fix to the issue, I think we can go ahead. Can your rebase the PR and maybe remove the comment about `any`? I think we're OK if we leave `any` there.

miguelrincon:
@cexbrayat Thanks for the response. I removed the comment and rebased. Can you have another look?

cexbrayat:
@miguelrincon It looks like the linter in unhappy, can you take a look? thank you!

cexbrayat:
Thank you !

miguelrincon:
@cexbrayat Thanks for merging. I see the last release, 2.4.6, was quite some time ago. Do you plan to create a new release anytime soon?

PR #2586, opened 2025-01-08
fix(stubs): Pass stubbed function to function props

markrian:
The original implementation, added in
22c7698425f2745c573d6b8154e171e158449d8e, passed a string value to a function prop. This instead passes an actual function, which stringifies to the same string.

The problem with the previous approach is that Vue started warning about strings passed to function props in
https://github.com/vuejs/core/commit/7ccd453dd004076cad49ec9f56cd5fe97b7b6ed8. As a result, this produced excessive noise in the logs that the user can do nothing about.

### DIFF
--- a/src/vnodeTransformers/stubComponentsTransformer.ts
+++ b/src/vnodeTransformers/stubComponentsTransformer.ts
@@ -42,6 +42,9 @@ interface StubOptions {
   renderStubDefaultSlot?: boolean
 }
 
+function stubbedFunctionProp() {}
+stubbedFunctionProp.toString = () => '[Function]'
+
 const normalizeStubProps = (props: ComponentPropsOptions) => {
   // props are always normalized to object syntax
   const $props = props as unknown as ComponentObjectPropsOptions
@@ -51,7 +54,7 @@ const normalizeStubProps = (props: ComponentPropsOptions) => {
       return { ...acc, [key]: [value?.toString()] }
     }
     if (typeof value === 'function') {
-      return { ...acc, [key]: ['[Function]'] }
+      return { ...acc, [key]: [stubbedFunctionProp] }
     }
     return { ...acc, [key]: value }
   }, {})


markrian:
@xanf Would you take a look at this please, given you wrote https://github.com/vuejs/test-utils/commit/22c7698425f2745c573d6b8154e171e158449d8e?

For context, see https://gitlab.com/gitlab-org/gitlab-foss/-/jobs/8790444271/viewer, showing the log exceeding the size limit due to [these](https://github.com/vuejs/core/blob/v3.5.13/packages/runtime-dom/src/modules/events.ts#L132-L133) messages.

cexbrayat:
Hi @markrian 

Thank you for your contribution.
IIRC we already fixed this in https://github.com/vuejs/test-utils/pull/2413 and released it in v2.4.6

If you are still experimenting this issue with v2.4.6, can you add a unit test to your PR that reproduces the issue and demonstrates your fix please?

markrian:
@cexbrayat Aha, I didn't realise. As it happens we're on v2.3.1, but we're _also_ monkey patching `createStubs` (long story), which recreates the issue that https://github.com/vuejs/test-utils/pull/2413 fixed. So our own monkey patching is to blame.

In any case:

> We can "trick" the warning introduced in Vue v3.4.22 (See https://github.com/vuejs/core/commit/7ccd453dd004076cad49ec9f56cd5fe97b7b6ed8) by using an array of strings instead of a string (as the new check allows functions and arrays, but not strings).

It seems to me the solution here of passing an actual function with a custom `toString` method is a bit more robust, since if Vue changes how it [checks](https://github.com/vuejs/core/blob/v3.5.13/packages/runtime-dom/src/modules/events.ts#L128) to something like `isFunction(value) || (isArray(value) && value.every(isFunction))` then the current approach would cause the warnings to appear again.

WDYT? I'm happy to close this PR anyway if you'd rather keep it as-is.

cexbrayat:
👍 I think we're fine for now then, lets' close this.
Thank you anyway!

PR #2599, opened 2025-01-15
fix: accept FunctionalComponent<any> as selector

### LINKED ISSUE #2594: Bug: type error with findComponent and FunctionalComponent
**Describe the bug**

Calling `findComponent` with a `FunctionalComponent` requiring more than zero `props` or `emits` causes a TypeScript error.
The parameter type for one override of `findComponent` is `FunctionalComponent` = `FunctionalComponent<{}, {}, any, {}>`, which does not accept `props` or `emits` with required fields.
Even though a type error occurs, running the test has no problem.

**To Reproduce**

- Call `findComponent` with a `FunctionalComponent` requiring more than zero props.
Repro: https://stackblitz.com/edit/github-13tgjghs?file=src%2Fcomponents%2F__tests__%2FHelloWorld.spec.ts

**Expected behavior**

`findComponent` should accept `FunctionalComponent<any, EmitsOptions | Record<string, any[]>, any, EmitsOptions>` as its parameter.

**Related information:**

```
  System:
    OS: Linux 5.15 Debian GNU/Linux 11 (bullseye) 11 (bullseye)
    CPU: (28) x64 Intel(R) Core(TM) i7-14700F
    Memory: 10.67 GB / 15.50 GB
    Container: Yes
    Shell: 5.1.4 - /bin/bash
```

**Additional context**

takejohn:
This PR changes the signature of `findComponent`, `findComponents`, `getComponent` to accept `FunctionalComponent<any>` as the selector.

Resolves #2594.

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -1106,8 +1106,8 @@ Finds a Vue Component instance and returns a `VueWrapper` if found. Returns `Err
 
 ```ts
 findComponent<T extends never>(selector: string): WrapperLike
-findComponent<T extends DefinedComponent>(selector: T | Exclude<FindComponentSelector, FunctionalComponent>): VueWrapper<InstanceType<T>>
-findComponent<T extends FunctionalComponent>(selector: T | string): DOMWrapper<Element>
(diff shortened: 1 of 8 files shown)


cexbrayat (approved):
Thanks!

PR #2612, opened 2025-02-10
Update v-model.md

ilyasherstoboev:
Test throw error `AssertionError: expected '' to be 'test' // Object.is equality` because modelValueInput `<input :value="modelValue" type="number" @input="$emit('update:modelValue', $event.target.value)"/>` have `type="number"`


cexbrayat (approved):
Good catch, thanks!

PR #2632, opened 2025-03-13
fix: use Teleport with disabled=true instead of stub

### LINKED ISSUE #2628: Bug: Reactivity breaks when Teleport is stubbed in unit tests, causing unexpected unmounts on teleported components
**Describe the bug**
Stubbing out Teleport in unit tests causes reactivity to behave unexpectedly. Specifically, the teleport component unmounts when its props are updated, but only when the Teleport is stubbed. This behavior differs from what is observed in the browser console, where the Child component stays mounted and reactivity works correctly.

**To Reproduce**
A minimal reproduction is available here: https://codesandbox.io/p/devbox/adoring-water-3pxr5p

**Repro Steps**
1. Fork the project using the link above
2. Open a new terminal
3. Run pnpm test

**Expected behavior**
Both unit tests should pass, with reactivity working regardless of whether Teleport is stubbed or pointing to a real target.

**Observed behaviour**
Only the non-stubbed teleport test is passing.

**Related information:**
System:
    OS: Linux 6.1 Debian GNU/Linux 12 (bookworm) 12 (bookworm)
    CPU: (2) x64 AMD EPYC
    Memory: 1.57 GB / 4.14 GB
    Container: Yes
    Shell: Unknown
  npmPackages:
    @vue/test-utils: 2.4.6 => 2.4.6 
    vitest: 3.0.8 => 3.0.8 
    vue: ^3.5.13 => 3.5.13 

**Additional context**
Credits to @victor141516 for helping triaging this issue.

victor141516:
fixes #2628 

cc @Alevagre7

As part of #1889 the child of Teleport is wrapped with an arrow function. This aligns with how Vue expects it (that's why we don't have the warning anymore). However the stub doesn't use the same implementation for the Teleport as Vue does: Vue doesn't render the children, but waits for the Teleport component to be mounted, then render the children in the teleport target element.

Instead of using a stub, we have a proposal to utilize the `disabled` prop from the Vue API for the Teleport component (https://vuejs.org/guide/built-ins/teleport#disabling-teleport) instead of creating a custom stub for it.

By using the official API, we can rely on native functionality rather than introducing a stub, which can lead to more friction and increase the likelihood of errors. The `disabled` prop allows us to effectively "disable" the teleport behavior, which aligns with the intent behind stubbing—removing specific behavior from the unit test.

In essence, whenever someone creates a stub for a component, the goal is typically to abstract away its behavior during testing. Using the `disabled` prop achieves the same outcome in a cleaner and more reliable manner.

The implementation may not be the best, so @cexbrayat if you have suggestions for it we can change it.

Thanks!

### DIFF
--- a/src/vnodeTransformers/util.ts
+++ b/src/vnodeTransformers/util.ts
@@ -85,7 +85,16 @@ export const createVNodeTransformer = ({
       registerStub({ source: originalType, stub: transformedType })
       // https://github.com/vuejs/test-utils/issues/1829 & https://github.com/vuejs/test-utils/issues/1888
       // Teleport/KeepAlive should return child nodes as a function
-      if (isTeleport(originalType) || isKeepAlive(originalType)) {
+      if (isTeleport(originalType)) {
+        return [
+          originalType,
+          { ...props, disabled: true },
+          children,
+          ...restVNodeArgs
+        ]
+      }
+
+      if (isKeepAlive(originalType)) {
         return [transformedType, props, () => children, ...restVNodeArgs]
       }
     }
--- a/tests/components/TeleportRemountChild.vue
+++ b/tests/components/TeleportRemountChild.vue
@@ -1,0 +1,17 @@
+<script setup lang="ts">
+import { onMounted, onUnmounted } from 'vue'
+
+const emit = defineEmits<{ myEvent: ['mount' | 'unmount'] }>()
+
+onMounted(() => {
+  emit('myEvent', 'mount')
+})
+
+onUnmounted(() => {
+  emit('myEvent', 'unmount')
+})
+</script>
+
+<template>
+  <span>child</span>
+</template>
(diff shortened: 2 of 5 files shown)


cexbrayat (commented):
Thanks for looking into this @victor141516 

The idea is interesting but I think we shouldn't lose the `teleport-stub` element, to be more aligned with all the other stubs. I guess it would be fine if the result was:

```vue
<teleport-stub to="body">
    <!--teleport start-->
    <div id="count">1</div>
    <!--teleport end-->
</teleport-stub>
```

Even if that might already be breaking for some people, I suppose we could live with that.

Or maybe we have a better alternative: we could provide a `TeleportStub` that does exactly what you suggest, which would be similar to what we do with `RouterLinkStub` (see https://github.com/vuejs/test-utils/blob/main/src/components/RouterLinkStub.ts). People would then use it like:

```ts
  global: {
    stubs: {
      Teleport: TeleportStub,
    },
  },
})
```

I think this could be a good option. Would that be enough for your use-case?

cexbrayat:
Yes, exactly.

VTU will just have to offer the `TeleportStub,` and developers can opt-in when needed.

PR #2713, opened 2025-09-19
fix: make mount options slots compatible with noUncheckedIndexedAccess true

Thomaash:
It's not necessary to define all slots, pratially defined slots are used all over the place throughout the tests here and routinely by users. However, the type for slots in mount options was not partiall and that caused type errors for people using `noUncheckedIndexedAccess: true` even though it work fine in runtime. Adding `?` doesn't change anything for users with `noUncheckedIndexedAccess: false`, it has passed type checking before and will continue to pass. Type checking will now also pass for users with partially defined slots using `noUncheckedIndexedAccess: true`.


Thomaash:
I added the missing test

Issue #2591, opened 2025-01-09
Bug: `findComponent` doesn't not return exposed properties in vm if component is bundled


renatodeleao:
**Describe the bug**
This is a bit hard to describe but hopefully the reproduction demo will clear things up.
I have a big project with 7000+ thousand tests. Recent ones are good, old ones are bad and rely heavily on `.vm`. 

I'm migrating the code base to a monorepo and the design system is now bundled as separate package by `vite` and the apps import the components’ bundled version instead of the `.vue`, tests inherently do too. I noticed that most of the tests that rely on `vm` stopped working.
```js
wrapper.findComponent(VComp).vm.someMethod // someMethod is always undefined, before bundle it wasn't.
```

> [!NOTE]
> I don't know if this is `VTU` fault or if it's by design, but having in mind that if I import the raw `.vue` component instead of the bundled `.js` version everything works I am assuming it is — Please close this issue if my assumption is wrong.

**To Reproduce**
https://stackblitz.com/edit/vitest-dev-vitest-h8drorow?file=apps%2Fdemo%2Ftest%2Fbasic.test.ts
Should run automatically, if not `npm run test`

**Expected behavior**
exposed properties via `defineExpose` are available though of `findComponent` returned `VueWrapper`

Assuming a basic `VExample.vue` component
```html
<!-- packages/ui/src/VExample.vue -->
<template>
  <div>Example</div>
</template>

<script setup>
const props = defineProps({
  a: String,
});

const exposedFn = () => {
  console.log('hey');
};

defineExpose({
  exposedFn,
});
</script>
```

Bundles to 
```js
import { openBlock as o, createElementBlock as p } from "vue";
const l = {
  __name: "VExample",
  props: {
    a: String
  },
  setup(n, { expose: e }) {
    return e({
      exposedFn: () => {
        console.log("hey");
      }
    }), (s, t) => (o(), p("div", null, "Example"));
  }
};
export {
  l as VExample
};
```

Faulty test
```js
import { VExample as VExampleRaw } from '../../packages/ui/src/VExample.vue'
import { VExample as VExampleBundled } from 'ui'

// swapping VExampleBundled per VExampleRaw in the components option makes it pass.
it('finds exposed "vm" properties in *bundled* component wrapper via findComponent', () => {
  const wrapped = mount({
    components: {
      VExampleBundled,
    },
    template: `<div><v-example /></div>`,
  });

  const test = wrapped.findComponent({ name: 'VExample' });

[...]

PR #2715, opened 2025-09-30
fix(#2591): `findComponent` doesn't not return exposed properties in vm

adamberecz:
After investigating the issue down to `@vue/runtime-core` it appears that `exposeProxy` is only set in a script-setup component if it uses at least one directive:
https://github.com/vuejs/core/blob/main/packages/runtime-core/src/component.ts#L1187
https://github.com/vuejs/core/blob/main/packages/runtime-core/src/directives.ts#L143

Meaning a component can have an exposed API while `exposeProxy` is `null`. In this case we can create manually from `exposed` which essentially has the same effect, because the one created by runtime-core is only different by also proxying `publicPropertiesMap` which aren't available anyway.


cexbrayat (changes requested):
Thank you for looking into it. Would you mind adding a unit test that showcases that the issue no longer exists?

adamberecz:
I've stumbled upon another issue related to using `devtoolsRawSetupState` which might not be available in some cases. I'm still investigating it as it might also affect this behavior. I'll get to you and possibly update the PR (+ add some tests).

cexbrayat:
Hi @adamberecz 

When you have some time, can you finalize the PR so we can merge it?

adamberecz:
Aplogies @cexbrayat – I am pretty busy at the moment, but I will keep this visible in my to-do list

adamberecz:
Superseded by https://github.com/vuejs/test-utils/pull/2927 — rebased onto latest `main` with tests and the bundled `findComponent` case from #2591.

adamberecz:
Superseded by #2927, which rebases the fix onto latest main and covers the bundled findComponent case from #2591. Closing this one in favor of that PR.

PR #2727, opened 2025-10-23
Updated example TodoApp to script setup

hdJerry:
I noticed a few things which I am not sure if it was delibrate

- I updated the todo app in the examples to use script setup
- updated the test to use `describe` and `it` just like all the tests on the project
- updated the comment on the test in examples
- added some types

I noticed I couldn't run the test until I included the path to where the todo app test is, like so:
```js
include: ['tests/**/*.spec.ts', 'examples/**/*.spec.js'],
```

### DIFF
--- a/examples/TodoApp.vue
+++ b/examples/TodoApp.vue
@@ -1,57 +1,48 @@
 <template>
-<div>
-  <div
-    v-for="todo in todos"
-    :key="todo.id"
-    data-test="todo"
-    :class="getClass(todo.completed)"
+  <div>
+    <div
+      v-for="todo in todos"
+      :key="todo.id"
+      data-test="todo"
+      :class="{ completed: todo.completed }"
     >
-    {{ todo.text }}
-    <input
-      type="checkbox"
-      v-model="todo.completed"
-      data-test="todo-checkbox"
-    />
-  </div>
+      {{ todo.text }}
+      <input
+        type="checkbox"
+        v-model="todo.completed"
+        data-test="todo-checkbox"
+      />
+    </div>
 
-  <form data-test="form" @submit.prevent="createTodo">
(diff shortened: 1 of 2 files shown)


examples/TodoApp.vue
```diff
@@ -1,52 +1,57 @@
...
+<script setup lang="ts">
+import { ref } from 'vue'
 
-export default defineComponent({
-  name: 'TodoApp',
+type TODO = {
+  id: number;
+  text: string;
+  completed: boolean;
+};
```
cexbrayat: I would use an interface named `Todo` here:
```ts
interface Todo {
  id: number;
```

examples/TodoApp.vue
```diff
@@ -1,52 +1,57 @@
...
-      this.todos.push({
-        id: 2,
-        text: this.newTodo,
-        completed: false
-      })
-    }
+const createTodo = () =>{
+  const listLength = todos.value.length;
+  const lastItem = todos.value[listLength - 1];// get last todo item
+  const todoExist = todos.value.find(res => res.text.toLowerCase() === newTodo.value.toLowerCase());//check is todo already exists
```
cexbrayat: I think we should keep the example as simple as it was and remove this

examples/TodoApp.spec.js
```diff
@@ -1,32 +1,49 @@
 /**
  * This is the example app used in the documentation.
  * If you want to run it, you will need to build the final bundle with
- * yarn build. Then you can run this with `yarn test examples`
+ * pnpm build. Then you can run this with `pnpm test examples`
```
cexbrayat: I think the test file should be in TypeScript (and all references in docs should be updated accordingly)

PR #2734, opened 2025-11-09
feat: enforce consistent usage of type imports

marcalexiei:
I noticed that the package was exporting few types as part of a "runtime" export:

https://github.com/vuejs/test-utils/blob/94b2eee1981e6c99fc39df652f78e698ac5f6126/src/index.ts#L13-L28

https://github.com/vuejs/test-utils/blob/94b2eee1981e6c99fc39df652f78e698ac5f6126/src/index.ts#L25-L26

---

https://github.com/vuejs/test-utils/blob/94b2eee1981e6c99fc39df652f78e698ac5f6126/src/types.ts#L89

https://github.com/vuejs/test-utils/blob/94b2eee1981e6c99fc39df652f78e698ac5f6126/src/types.ts#L111

---

This PR enables and fixes all the issues reported by [`consistent-type-imports`](https://typescript-eslint.io/rules/consistent-type-imports/)  `typescript-eslint` rule to ensure that `import` / `export` types are marked correctly.

I have also updated the `eslint.config.mjs` to use helpers from `eslint/config` instead of the deprecated `tseslint.config`.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,12 +1,11 @@
+import { defineConfig, globalIgnores } from 'eslint/config'
 import eslint from '@eslint/js'
 import tseslint from 'typescript-eslint'
 import prettierPluginRecommended from 'eslint-plugin-prettier/recommended'
 
-export default [
+export default defineConfig(
+  globalIgnores(['docs/.vitepress', 'coverage', 'dist']),
   {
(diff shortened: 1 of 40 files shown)


cexbrayat (approved):
Thanks for the PR!

PR #2806, opened 2026-04-05
fix: correct <summary> element behavior

nn-morishita:
This PR addresses https://github.com/vuejs/test-utils/issues/2626.

The issue is that <summary> element behavior is inconsistent with the real browser.

I verified this in a real browser, and the issue still occurs.

Reproduction:
https://stackblitz.com/edit/vitejs-vite-qxxwzrgz?file=src%2Fcomponents%2Ftests%2FDetailsComponent.test.ts

### DIFF
--- a/src/utils/isElementVisible.ts
+++ b/src/utils/isElementVisible.ts
@@ -19,30 +19,35 @@ function isStyleVisible<T extends Element>(element: T) {
   )
 }
 
-function isAttributeVisible<T extends Element>(element: T) {
-  if (element instanceof HTMLInputElement && element.type === 'hidden') {
+/**
+ * checks if an element is visible by checking its attributes and its parents' attributes.
+ * @param element 
+ * @returns boolean
+ */
+function isElementVisibleByAttribute<T extends Element>(element: T) {
+  if (element instanceof HTMLInputElement && (element.type === 'hidden' || element.hidden)) {
     return false
   }
-
-  if (element instanceof HTMLElement && element.hidden) {
-    return false
-  }
-
   return true
 }
 
+/**
+ * checks if an element is hidden by a closed details element.
+ * @param element 
+ * @returns boolean
+ */
 function isHiddenByClosedDetails<T extends Element>(element: T) {
+  const summary = element.closest('summary')
   let parent = element.parentElement
 
   while (parent) {
     if (parent.nodeName === 'DETAILS' && !(parent as HTMLDetailsElement).open) {
-      if (
-        !(element.nodeName === 'SUMMARY' && element.parentElement === parent)
-      ) {
+      // If no <summary> ancestor exists, or the <summary> is not a direct child of this <details>, 
+      // then the content is hidden
+      if (!summary || summary.parentElement !== parent) {
         return true
       }
     }
-
     parent = parent.parentElement
   }
 
@@ -53,7 +58,7 @@ export function isElementVisible<T extends Element>(element: T): boolean {
   return (
     element.nodeName !== '#comment' &&
     isStyleVisible(element) &&
-    isAttributeVisible(element) &&
+    isElementVisibleByAttribute(element) &&
     !isHiddenByClosedDetails(element) &&
     (!element.parentElement || isElementVisible(element.parentElement))
   )
--- a/src/utils/isElementVisible.ts
+++ b/src/utils/isElementVisible.ts
@@ -20,16 +20,31 @@ function isStyleVisible<T extends Element>(element: T) {
 }
 
 function isAttributeVisible<T extends Element>(element: T) {
-  if (
-    element.nodeName === 'SUMMARY' &&
-    element.closest('details')
-  ) {
-    return true
+  if (element instanceof HTMLInputElement && element.type === 'hidden') {
+    return false
   }
-  return (
-    !element.hasAttribute('hidden') &&
-    (element.nodeName === 'DETAILS' ? element.hasAttribute('open') : true)
-  )
+
+  if (element instanceof HTMLElement && element.hidden) {
+    return false
+  }
+
+  return true
+}
+
+function isHiddenByClosedDetails<T extends Element>(element: T) {
+  let parent = element.parentElement
+
+  while (parent) {
+    if (parent.nodeName === 'DETAILS' && !(parent as HTMLDetailsElement).open) {
+      if (!(element.nodeName === 'SUMMARY' && element.parentElement === parent)) {
+        return true
+      }
+    }
+
+    parent = parent.parentElement
+  }
+
+  return false
 }
 
 export function isElementVisible<T extends Element>(element: T): boolean {
@@ -37,6 +52,7 @@ export function isElementVisible<T extends Element>(element: T): boolean {
     element.nodeName !== '#comment' &&
     isStyleVisible(element) &&
     isAttributeVisible(element) &&
-    (element.nodeName === 'SUMMARY' || !element.parentElement || isElementVisible(element.parentElement))
+    !isHiddenByClosedDetails(element) &&
+    (!element.parentElement || isElementVisible(element.parentElement))
   )
 }
--- a/src/utils/isElementVisible.ts
+++ b/src/utils/isElementVisible.ts
@@ -20,6 +20,12 @@ function isStyleVisible<T extends Element>(element: T) {
 }
 
 function isAttributeVisible<T extends Element>(element: T) {
+  if (
+    element.nodeName === 'SUMMARY' &&
+    element.closest('details')
+  ) {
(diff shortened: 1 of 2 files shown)


cexbrayat (changes requested):
Thanks for the PR!

I don't think this is enough though. The PR makes `<summary>` skip ancestor visibility checks entirely in isElementVisible.
That means a `<summary>` can be reported visible even when one of its ancestors is hidden (for example display: none), which is a regression because hidden ancestors should make descendants not visible.

pseudo-test to reproduce (did not run it, so maybe it needs tweaking):
```ts
it('should consider a summary as hidden when an ancestor is hidden', () => {
  const HiddenAncestorSummary = defineComponent({
    template: `
      <div style="display: none;">
        <details>
          <summary>Summary</summary>
        </details>
      </div>
    `
  })

  const wrapper = mount(HiddenAncestorSummary)
  expect(wrapper.find('summary').isVisible()).toBe(false)
})
```

I also think the PR treats any `<summary>` inside a `<details>` as visible, regardless of structure.
That is too broad: only the primary summary should be visible in a closed `<details>`. A nested `<summary>` inside closed details content should remain hidden.

```ts
it('should consider a summary as hidden when nested inside closed details content', () => {
  const NestedSummaryInClosedDetails = defineComponent({
    template: `
      <details>
        <summary>Main summary</summary>
        <div>
          <summary>Nested summary</summary>
        </div>
      </details>
    `
  })

  const wrapper = mount(NestedSummaryInClosedDetails)
  const summaries = wrapper.findAll('summary')

  expect(summaries[0].isVisible()).toBe(true)
  expect(summaries[1].isVisible()).toBe(false)
});
```

cexbrayat:
@nn-morishita are you still interested in fixing the issue or should we close the PR?

nn-morishita:
@cexbrayat Sorry for the delayed response. I’ve just pushed a [new commit](https://github.com/vuejs/test-utils/pull/2806/commits/d23d0a8f132e2f7467b685eab1d6710a520704ae) addressing the feedback. I’d appreciate it if you could take a look when you have a chance.

src/utils/isElementVisible.ts
```diff
@@ -20,17 +20,41 @@ function isStyleVisible<T extends Element>(element: T) {
 }
 
 function isAttributeVisible<T extends Element>(element: T) {
-  return (
-    !element.hasAttribute('hidden') &&
-    (element.nodeName === 'DETAILS' ? element.hasAttribute('open') : true)
-  )
+  if (element instanceof HTMLInputElement && element.type === 'hidden') {
+    return false
+  }
```
cexbrayat: This change is unrelated to the issue, but an interesting addition. Could you either:
- add a unit test in this PR
- remove this change from this PR and open another one with a unit test

cexbrayat: As the funciton no longer use attribute checking, maybe rename it to `isElementVisible`?

tests/isVisible.spec.ts
```diff
@@ -228,6 +228,50 @@ describe('isVisible', () => {
         })
         expect(wrapper.isVisible()).toBe(false)
       })
+      it('DetailContent should be visible when summary is visible', () => {
```
cexbrayat, line 245: the tests are under an unrelated test suite, maybe move them a new test suite:

```ts
describe('details and summary elements', () => {
  // test cases here
})
```

tests/isVisible.spec.ts
```diff
@@ -228,6 +228,50 @@ describe('isVisible', () => {
...
+        const DetailContent = defineComponent({
+          template: `<details><summary>Summary</summary><div>Content</div></details>`
+        })
+
+        const wrapper = mount(DetailContent)
+        expect(wrapper.find('details').isVisible()).toBe(true)
+        expect(wrapper.find('summary').isVisible()).toBe(true)
+        expect(wrapper.find('div').isVisible()).toBe(false)
+      })
```
cexbrayat, line 254: Maybe add a test case, like:

```ts
template: `<details><summary><span>Summary</span></summary><div>Content</div></details>`

expect(wrapper.find('summary span').isVisible()).toBe(true)
```

src/utils/isElementVisible.ts
```diff
@@ -20,17 +20,41 @@ function isStyleVisible<T extends Element>(element: T) {
...
+
+  if (element instanceof HTMLElement && element.hidden) {
+    return false
+  }
+
+  return true
+}
+
+function isHiddenByClosedDetails<T extends Element>(element: T) {
```
cexbrayat, line 40: Maybe add a few lines of jsdoc to explain what this function does

cexbrayat (changes requested):

src/utils/isElementVisible.ts
nn-morishita: Since a function named `isElementVisible` already exists, I renamed this function to `isElementVisibleByAttribute`.

tests/isVisible.spec.ts
nn-morishita: Thank you for pointing that out. I've made the changes as suggested!

tests/isVisible.spec.ts
nn-morishita: I added the nested test cases you pointed out.

nn-morishita:
@cexbrayat 
Thank you for your feedback! I've made the changes—please let me know if this aligns with your intention.

tests/isVisible.spec.ts
```diff
@@ -228,6 +228,31 @@ describe('isVisible', () => {
         })
         expect(wrapper.isVisible()).toBe(false)
       })
+    })
+
+    describe('details and summary elements', () => {
+      it('DetailContent should be invisible when display:none is applied by class', () => {
+        const style = document.createElement('style')
+        document.head.appendChild(style)
+        style.sheet!.insertRule('.hidden { display: none; }')
```
cexbrayat: I'm afraid that this will leak into other tests. I would remove this test, the following one should be enough.

tests/isVisible.spec.ts
```diff
@@ -228,6 +228,31 @@ describe('isVisible', () => {
...
+          template: '<div id="my-div" class="hidden"><details><summary>Summary</summary></details></div>'
+        })
+
+        expect(wrapper.get('#my-div').isVisible()).toBe(false)
+        expect(wrapper.find('summary').isVisible()).toBe(false)
+        expect(wrapper.find('details').isVisible()).toBe(false)
+      })
+      it('DetailContent should be invisible when display:none is applied by style attribute', () => {
+        const wrapper = mount({
+          template: '<div id="my-div" style="display: none;"><details><summary>Summary</summary></details></div>'
```
cexbrayat: use `defineComponent` to declare the component.

src/utils/isElementVisible.ts
```diff
@@ -19,30 +19,35 @@ function isStyleVisible<T extends Element>(element: T) {
...
-  }
-
-  if (element instanceof HTMLElement && element.hidden) {
+/**
+ * checks if an element is visible by checking its attributes and its parents' attributes.
+ * @param element 
+ * @returns boolean
+ */
+function isElementVisibleByAttribute<T extends Element>(element: T) {
+  if (element instanceof HTMLInputElement && (element.type === 'hidden' || element.hidden)) {
```
cexbrayat: This now only checks `hidden` for HTMLInputElement whereas it was checking it for all HTMLElements previously.

And as I was mentioning before, the hidden check for inputs is a nice addition but should have a unit test (or even better, a dedicated PR).

cexbrayat (changes requested):

tests/isVisible.spec.ts
nn-morishita: I fixed this point. I remove this test case.

src/utils/isElementVisible.ts
nn-morishita: Thanks for pointing that out! You're right — my change unintentionally removed the hidden check for other HTMLElements. I restored the previous behavior and removed the unrelated input hidden change from this PR.

nn-morishita:
@cexbrayat 
Sorry for the late reply. I've addressed your comments.

cexbrayat:
@nn-morishita can you make sure format and tests are working (I suspect the removal of `mount` in a test in your last commit is unintended)?

nn-morishita:
@cexbrayat 
Thanks for pointing that out. The removal of `mount` was indeed unintended. I've restored it and verified that the formatting and tests pass locally.

cexbrayat:
@nn-morishita Thanks for the commit, but it looks like the formatting is still failing?

cexbrayat (approved):

PR #2835, opened 2026-04-27
fix: restore text()/html() for single-root components

lorenzvanherwaarden:
Fixes a regression introduced in #2579 where `text()` / `html()` returns an empty string for single-root components whose DOM content isn't reflected in the vnode `children` array.                                    
                                                                                                                                                                                                                                                    
#2579 made `DOMWrapper.getRootNodes()` walk `subTree.children` whenever it was an array, in order to support array-returning functional components like `() => ['foo', 'bar']`. But `subTree.children` only represents "rendered roots" when `subTree` is a `Fragment`. For single-root components, `subTree` is the rendered element vnode itself and `children` represents that element's own children.                                                                                                                                                                                  
                                                                                                                                                                                                                                                    
Concrete failure: bootstrap-vue's `BButton` is a Vue 2 functional component that injects its label via `domProps: { textContent }`. Under `@vue/compat`, the rendered element has the correct `textContent`, but the vnode's `children` stays `[]`.  See https://gitlab.com/gitlab-org/gitlab-services/design.gitlab.com/-/merge_requests/5890 for the failures.

This fixes the regression by adding `subTree.type === Fragment` guard to `DOMWrapper.getRootNodes()`, so the children-walk only runs for true multi-root subtrees and single-root components fall back to the wrapper element

### DIFF
--- a/src/domWrapper.ts
+++ b/src/domWrapper.ts
@@ -26,7 +26,10 @@ export class DOMWrapper<NodeType extends Node> extends BaseWrapper<NodeType> {
   }
 
   getRootNodes() {
-    if (this.subTree?.type === Fragment && Array.isArray(this.subTree?.children)) {
+    if (
+      this.subTree?.type === Fragment &&
+      Array.isArray(this.subTree?.children)
+    ) {
       return this.subTree.children.map(node => (node as any)?.el)
     }
     return [this.wrapperElement]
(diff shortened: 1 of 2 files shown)


lorenzvanherwaarden:
@cexbrayat could you take a look at a fix for a regression we're seeing?

cexbrayat (approved):
Thanks for this PR @lorenzvanherwaarden 👍

lorenzvanherwaarden:
Thanks @cexbrayat for the quick review! 😊 

Do you have an idea when a new release might be made?

cexbrayat:
2.4.10 is on its way (waiting for someone to approve the release)

lorenzvanherwaarden:
Fantastic @cexbrayat! Thank you.

PR #2846, opened 2026-05-04
fix(#2845): setData handling mixed setup()/data() components correctly 

### LINKED ISSUE #2845: Bug: setData silently no-ops on data() properties when component also has setup()
### Describe the bug

Regression from https://github.com/vuejs/test-utils/pull/2655

The PR introduced a heuristic in `.setData` to distinguish Options API from Composition API components, but it currently fails for "mixed-api" components: these components have both options `data()` and and `setup()` function returning data.

https://github.com/vuejs/test-utils/blob/57c138c7ca1013f9db0ad3b45bc3a11ea7ab80c5/src/vueWrapper.ts#L250-L275

The assumption behind branch 2 is that an unfrozen setupState means the component is purely setup()-based. That assumption breaks for "mixed-API" components: a `setup()` returning any value leaves `setupState` unfrozen, yet the component's reactive state can also still live in `data()`. 

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-3waeemm1?file=package.json

> [!NOTE] 
> Edit package.json to install 2.4.6 and then run `npm i && npm run test` in the terminal to see the test passing.

### Expected 
`.setData` updates reactive bindings in `data()` and the template re-renders.

### Actual 

`.setData` silently writes to `setupState` instead, where the key doesn't exist. `$data` is never touched and the component stays in its initial state.

### Environment

```
  System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  npmPackages:
    @vue/test-utils: 2.4.7 => 2.4.7 
    vitest: latest => 4.1.5 
    vue: latest => 3.5.33 
```

### Additional context

I can workaround by re-writing my tests from `.setData` to 

```js
wrapper.vm.foo = 'bar'
await wrapper.vm.$nextTick()
```

but i would rather not.

renatodeleao:
Closes https://github.com/vuejs/test-utils/issues/2845

By mixed-api components we mean components that have both `setup()` and `data()` options returning values.

### DIFF
--- a/src/vueWrapper.ts
+++ b/src/vueWrapper.ts
@@ -247,16 +247,18 @@ export class VueWrapper<
     return domWrapper.isVisible()
   }
 
+  /*
+   * Depending on how the component was defined, data can live in different places.
+   * Vue sets some default placeholder in all the locations however, so we cannot
+   * just check if the data object exists or not.
+   *
+   * - <script setup>: data lives in setupState, marked with __isScriptSetup
+   * - setup() function: data lives in setupState, not marked with __isScriptSetup
+   * - Options API data(): data lives in $data, proxied through $data; setupState exists but is frozen
+   * - Mixed API (setup() + data()): setupState is unfrozen (has setup() returns) AND $data exists.
+   *   Each key must be routed individually — keys present in $data go there, the rest go to setupState.
+   */
   setData(data: Record<string, unknown>): Promise<void> {
-    /*
-    Depending on how the component was defined, data can live in different places 
-    Vue sets some default placeholder in all the locations however, so we cannot just check
-    if the data object exists or not.
-    When using <script setup>, data lives in the setupState object, which is then marked with __isScriptSetup
-    When using the setup() function, data lives in the setupState object, but is not marked with __isScriptSetup
-    When using the object api, data lives in the data object, proxied through $data, HOWEVER
-    the setupState object will also exist, and be frozen.
-    */
     // @ts-expect-error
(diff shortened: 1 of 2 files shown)


src/vueWrapper.ts
```diff
@@ -247,26 +247,35 @@ export class VueWrapper<
...
       // @ts-expect-error
     } else if (!Object.isFrozen(this.componentVM.$.setupState)) {
-      // data from setup() function when using the object api
-      // @ts-expect-error
-      mergeDeep(proxyRefs(this.componentVM.$.setupState), data)
+      // data from setup() function — may also have data() properties in mixed-API components,
+      // so route each key to wherever it actually lives
+      for (const [key, value] of Object.entries(data)) {
+        if (key in (this.componentVM.$data as Record<string, unknown>)) {
+          ;(this.componentVM.$data as Record<string, unknown>)[key] = value
```
renatodeleao: review-to-self: add mergeDeep here too for consistency.

cexbrayat (approved):

cexbrayat:
Thanks @renatodeleao 👍

MartinKassarBof:
In what version of test-utils is this fix in? or coming out in? :)

vlad-frontdev:
Can we have it released?

andreclemente:
Hey, any idea when will this be tagged?

Issue #2626, opened 2025-03-07
Bug: expect summary element inside a details element to be visible


jashaj:
**Describe the bug**

`isVisible()` always returns `false` for the `details` element and its descendants when there is no `open` attribute.
The `summary` is visible in the browser even when `details` is not open. `details` is visible, but any other children than `summary` are hidden when the attribute `open` is not present. 

**To Reproduce**

Given component `DetailsComponent.vue`:

```vue
<template>
  <details :open="open">
    <summary>Details</summary>
    <p>Something small enough to escape casual notice.</p>
  </details>
</template>

<script setup lang="ts">
defineProps<{
  open?: boolean;
}>();
</script>
```

**Expected behavior**

```ts
const wrapper = mount(DetailsComponent)
const details = wrapper.find('details');
expect(details.exists()).toEqual(true);
expect(details.isVisible()).toEqual(true); // debatable, because only `summary` inside `details` is visible

const summary = details.find('summary');
expect(summary.isVisible()).toEqual(true); // currently fails

expect(details.find('p').isVisible()).toEqual(false)
```

**Additional context**

A text node can also be a direct child of `details`. That direct text node is hidden when  the attribute `open` is missing;

Used test-utils version 2.4.6

cexbrayat:
Hi @jashaj 

Can you provide us a small repro online using https://stackblitz.com/github/vuejs/create-vue-templates/tree/main/typescript-vitest?file=src%2Fcomponents%2F__tests__%2FHelloWorld.spec.ts ? 

It only takes a few minutes, and we'll be able to take a look

jashaj:
Hi @cexbrayat, here's the test file: https://stackblitz.com/edit/github-utvcwwef?file=src%2Fcomponents%2F__tests__%2FDetailsComponent.spec.ts

cexbrayat:
I wouldn't be surprised if this was a jsdom issue. have you tried to run the test in a real browser? (see https://vitest.dev/guide/browser/)

jashaj:
It's more likely [this line](https://github.com/vuejs/test-utils/blob/297965cb9fc0c3deb8131032ac9252278cd31d6e/src/utils/isElementVisible.ts#L25) but I wasn't sure if `summary` was overlooked while implementing this.

cexbrayat:
That could be 👍 Would you like to open a PR to fix this? We would gladly review and merge it!

PR #2847, opened 2026-05-04
fix(isVisible): treat <summary> in closed <details> as visible

AhmadHannan037:
Fixes [#2626](https://github.com/AhmadHannan037/test-utils/issues/2626).

isVisible() previously returned false for `<summary>` (and `<details>` itself) when the parent `<details>` had no open attribute, even though the browser renders both. The check now mirrors HTML semantics: only non-<summary> descendants of a closed `<details>` are hidden.

`<details>` (open or closed) → visible
`<summary>` (direct child) → visible
Other descendants → visible only when open
Tests added in tests/isVisible.spec.ts covering the open, closed, and nested-non-summary cases.

### DIFF
--- a/src/utils/isElementVisible.ts
+++ b/src/utils/isElementVisible.ts
@@ -19,18 +19,34 @@ function isStyleVisible<T extends Element>(element: T) {
   )
 }
 
-function isAttributeVisible<T extends Element>(element: T) {
-  return (
-    !element.hasAttribute('hidden') &&
-    (element.nodeName === 'DETAILS' ? element.hasAttribute('open') : true)
-  )
+function isAttributeVisible<T extends Element>(
+  element: T,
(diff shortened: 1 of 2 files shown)


cexbrayat:
Hi @AhmadHannan037 

Thanks for the PR, but we already have one opened to solve the same issue in #2806 that we are going to prioritize.
Feel free to take a look at the other PR and suggest improvements

AhmadHannan037:
> Hi @AhmadHannan037
> 
> Thanks for the PR, but we already have one opened to solve the same issue in #2806 that we are going to prioritize. Feel free to take a look at the other PR and suggest improvements

Going to close this since [#2806](https://github.com/vuejs/test-utils/pull/2806) is already addressing the issue and is further along in review. Happy to help review there. Thanks!

Issue #2544, opened 2024-11-14
Bug: 


spthiel:
**Describe the bug**
When using element.trigger('keydown.xyz'), some keys incorrectly pass the keyCode as code to handlers of KeyboardEvents.

**To Reproduce**
https://stackblitz.com/edit/github-bbenms?file=src%2FButton.vue&view=editor

->
Event.code logs a string representation of Event.keyCode for various keys.
Tested keys that do not work correctly: Home, End, Enter, Space, Escape

**Expected behavior**
Event.code should correctly log the name of the key

**Related information:**
  System:
    OS: Linux 5.15 Linux Mint 21.3 (Virginia)
    CPU: (20) x64 12th Gen Intel(R) Core(TM) i7-12700H
    Memory: 14.30 GB / 31.02 GB
    Container: Yes
    Shell: 0.96.0 - /home/n/.cargo/bin/nu
  npmPackages:
    @vue/test-utils: ^2.4.5 => 2.4.6 
    vitest: ^1.5.0 => 1.6.0 
    vue: ^3.5.12 => 3.5.12 

**Additional context**

cexbrayat:
Yes, I think this behavior is there for backward compatibility (see https://github.com/vuejs/test-utils/blob/3988a3161074d3ea6bbcbc57995f4f2829a83b39/src/createDomEvent.ts#L102-L117)

Feel free to open a PR if you think that could be omproved without breaking existing applciations 👍

spthiel:
> for backward compatibility 

What kind of backwards compatability are we talking here? To my knowledge Event.code has never used to house the keyCode.
This seems to me to just break web standard

cexbrayat:
keyCode was a thing but is deprecated https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

But as I said, we're welcoming any contribution to improve this use case

spthiel:
> keyCode was a thing but is deprecated
Yes I understand keyCode was deprecated but "code" wasn't and was never used to save a number.
I can contribute here but my idea would be to just remove the "backwards compatibility"

spthiel:
> (see)

Actually looking at your code snippet it seems this issue would already be partially solved if a new version was built due to https://github.com/vuejs/test-utils/pull/2434 which would at least allow the 2nd test case in the repro to succeed

PR #2850, opened 2026-05-06
fix(trigger): set spec-compliant `event.code` on keydown/keyup

AhmadHannan037:
Fixes [#2544](https://github.com/AhmadHannan037/test-utils/issues/2544).

trigger('keydown.enter') (and other keydown/keyup modifiers) was setting event.code to the numeric keyCode (e.g. 13) instead of the [W3C UI Events](https://www.w3.org/TR/uievents/#code-virtual-keyboards) string identifier ("Enter"). Tests checking event.code for accessibility/keyboard-navigation handlers couldn't pass because of this.

Change
Added codesByKeyName next to keyCodesByKeyName in src/constants/dom-events.ts, mapping every supported key modifier to its W3C code string.
createDomEvent.ts now picks event.code from codesByKeyName when a key modifier is present. Explicit options.code still takes priority. The numeric keyCode fallback is kept for legacy callers that pass { keyCode } without a modifier, so existing behavior is unchanged.
Coverage
Added a sibling test in tests/trigger.spec.ts (next to the existing keyCode-mapping test) that iterates codesByKeyName and asserts the handler receives the correct event.code string for every modifier.

Affected modifiers: backspace, tab, enter, esc, space, pageup, pagedown, end, home, left, up, right, down, insert, delete.

### DIFF
--- a/src/constants/dom-events.ts
+++ b/src/constants/dom-events.ts
@@ -91,6 +91,26 @@ export const keyCodesByKeyName = {
   delete: 46
 } as const
 
+// Maps each supported key modifier to the W3C UI Events `KeyboardEvent.code`
+// string. Keep keys aligned with `keyCodesByKeyName`.
+export const codesByKeyName = {
+  backspace: 'Backspace',
+  tab: 'Tab',
+  enter: 'Enter',
+  esc: 'Escape',
+  space: 'Space',
+  pageup: 'PageUp',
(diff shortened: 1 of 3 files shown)


cexbrayat (approved):
Thanks @AhmadHannan037

AhmadHannan037:
> Thanks @AhmadHannan037

Thanks for the quick review!

PR #2860, opened 2026-05-12
fix(security): pin patched transitive dependencies

arpitjain099:
## Summary
- add `pnpm.overrides` for vulnerable transitive dependencies reported by Dependabot
- refresh `pnpm-lock.yaml` to remove vulnerable resolutions
- keep app/library source unchanged; this is dependency-resolution hardening only

## Patched packages pinned
- `simple-git` -> `3.36.0`
- `minimatch` -> `5.1.8`, `9.0.7` (range-specific)
- `picomatch` -> `2.3.2`, `4.0.4` (range-specific)
- `brace-expansion` -> `2.0.3`
- `esbuild` -> `0.25.0`
- `mdast-util-to-hast` -> `13.2.1`
- `vite` -> `8.0.12`

## Validation
- `pnpm test`

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,6 +119,17 @@
   },
   "packageManager": "pnpm@10.33.4",
   "pnpm": {
+    "overrides": {
+      "brace-expansion@>=2.0.0 <2.0.3": "2.0.3",
+      "esbuild": "0.25.0",
+      "mdast-util-to-hast": "13.2.1",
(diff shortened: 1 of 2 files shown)


cexbrayat:
I don't think we need this, all of these dependencies are just for people developing VTU, not for end users of VTU, so there is no risk.

PR #2890, opened 2026-06-22
fix(#2675): keep re-rendering root component after another wrapper is mounted

### LINKED ISSUE #2675: Bug: setProps not behaving as expected
**Describe the bug**
`setProps` seems to work only in isolated tests, if more tests are present it doesn't seem to be updating the component props as expected

**This started failing from 2.3.1**

**To Reproduce**
https://codesandbox.io/p/devbox/eager-waterfall-57d5kh?file=%2Fsrc%2FApp.spec.js%3A19%2C6

When running `npm run test` you can see that the `can change prop` test fails, but if you only run that test with `it.only("can change prop"` it passes

this was not the case before 2.3.1, wonder if this is expected behaviour or not

also worth noting that if you remove the `just an extra test` the suite passes.

**Related information:**
This happens both on vitest and jest, the Vue version doesn't seem to matter either, I've seen these errors pop up in my codebase after upgrading vue test utils from `2.3.0` to `2.4.6`

**Additional context**

lazerg:
Closes #2675

`shallowMount`/`mount` install a vnode transformer via Vue's global `transformVNodeArgs`, which is last-call-wins: mounting a second wrapper replaces the first wrapper's transformer. That transformer only treats *its own* root component as "don't stub", so when an earlier wrapper later re-renders (e.g. via `setProps`), its root no longer matches and gets stubbed. The DOM keeps the old output and `setProps` looks like it does nothing.

That's why the issue only appears when a wrapper declared outside the test is reused across several tests, and mostly with `shallowMount`, where stubbing is active.

The fix tracks every mounted root component in a module-level `WeakSet` and treats any of them as a root in `isRootComponent`. It stays gated by `!instance.parent`, so genuine child components are still stubbed as before. Only a re-rendered root is spared.

Added a `shallowMount` test that mounts a second wrapper and then calls `setProps` on the first one. It fails on `main` (renders `initial`) and passes with this change (renders `updated`).

### DIFF
--- a/src/vnodeTransformers/util.ts
+++ b/src/vnodeTransformers/util.ts
@@ -29,6 +29,21 @@ export interface RootComponents {
   // If component is functional then contains the original component otherwise empty
   functional?: Component
 }
+// `transformVNodeArgs` is a global, last-call-wins Vue API, so when several
+// wrappers are mounted the active transformer only knows about the most
+// recently mounted root component. Track every mounted root component globally
+// so that re-rendering an earlier wrapper (e.g. via `setProps`) doesn't stub
+// its root just because another wrapper was mounted afterwards.
+// See https://github.com/vuejs/test-utils/issues/2675
+const mountedRootComponents = new WeakSet<object>()
+const registerRootComponent = (type?: Component): void => {
+  // WeakSet keys must be objects/functions; component definitions always are,
+  // but `Component` also allows a string name which we simply ignore.
+  if (type && typeof type !== 'string') {
+    mountedRootComponents.add(type)
+  }
+}
+
 export const isRootComponent = (
   rootComponents: RootComponents,
   type: VNodeTransformerInputComponentType,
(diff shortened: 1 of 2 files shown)


cexbrayat:
Hi @lazerg , thanks for the PR.

I don't think this really fixes the linked issue, as there is no usage of enableAutoUnmount in #2675 
And the issue occurs with shallowMount but not with mount

See https://stackblitz.com/edit/github-uatfqjiq?file=src%2Fcomponents%2F__tests__%2FHelloWorld.spec.ts for a repro

lazerg:
You're right, thanks for the repro. My first attempt misread the issue (no `enableAutoUnmount` involved).

Real cause: `shallowMount`'s vnode transformer is installed via Vue's global `transformVNodeArgs`, which is last-call-wins. Mounting a second wrapper replaces the first's transformer, and the new one only spares its own root from stubbing. So re-rendering the first wrapper via `setProps` stubs its root and the DOM keeps the old value (hence shallowMount-only, and only after a second mount).

Pushed a fix that tracks every mounted root in a `WeakSet` and treats any of them as root in `isRootComponent` (still gated by `!instance.parent`, so real children stay stubbed). Added a test that fails on `main` and passes with the change.

cexbrayat (approved):
Thanks @lazerg

PR #2895, opened 2026-06-29
fix: stub globally registered components by alias

### LINKED ISSUE #2541: Bug: `stubs` does not respect globally registered component names
**Describe the bug**
When stubbing child components using the `stubs` option, vue-test-utils will not match components against the name they have been registered under, but by the internal name given to them by the Vue compiler.

e.g. If the `Example.vue` component is added to a Vue app using `app.component('VtuExample', Example)`, `stubs : { VtuExample: true }` will not mock the component, but `stubs : { example: true }` will.

**To Reproduce**
1. Open the [minimal reproduction](https://stackblitz.com/edit/github-igs26c?file=app%2Fsrc%2F__tests__%2FApp.spec.ts&view=editor)
2. In the terminal, navigate to the `/lib` folder and execute the command `npm i && npm run build`
2. In the terminal, navigate to the `/app` folder and execute the command `npm i && npm test`

**Expected behavior**
Both tests in `App.spec.ts` should pass, as the component imported from /lib has been registered and used under the name "HelloWorld", not "hello".

**Related information:**
```
  System:
    OS: macOS 14.7.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 139.58 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    @vue/test-utils: 2.4.6 => 2.4.6 
    vitest: 1.6.0 => 1.6.0 
    vue: 3.5.12 => 3.5.12 
```

**Additional context**
With the increasing prevalence of the Composition API, it is no longer common to manually define a name for each written component (as can be done when using Options API). When you globally register a component using `app.component('name', component)`, for all intents and purposes within your code the component is called `name`. It's not possible anywhere in your code to use it's internal name to include it in templates. So from a test authors perspective, it is not obvious that the component can't be stubbed using the registered name, but only the Vue compiler generated name.

ifer47:
﻿## Summary

- resolve component registration names from the app-level component registry when matching `global.stubs`
- keep local component registration names as the first priority, then fall back to global registrations and existing script setup/name inference
- add a regression test for components registered by a plugin under a public alias that differs from the component's internal name

## Root Cause

`getComponentRegisteredName()` only inspected the parent component's local `components` option. Components installed through plugins or libraries with `app.component('GlobalAlias', Component)` were resolved by Vue at runtime, but Vue Test Utils could only see the component's internal `name`. As a result, `global.stubs.GlobalAlias` did not match unless the internal component name happened to be the same.

Fixes #2541.

## Validation

- `pnpm test tests/mountingOptions/global.stubs.spec.ts -t "globally registered component by its registration name" --run`
- `pnpm test tests/mountingOptions/global.stubs.spec.ts --run`
- `pnpm test tests/mountingOptions/global.components.spec.ts --run`
- `pnpm test --run`
- `pnpm run lint`
- `pnpm run vue-tsc`
- `pnpm exec oxfmt --check src/utils/componentName.ts tests/mountingOptions/global.stubs.spec.ts`

### DIFF
--- a/src/utils/componentName.ts
+++ b/src/utils/componentName.ts
@@ -28,6 +28,14 @@ export const getComponentRegisteredName = (
     }
   }
 
+  // try to infer the name based on global resolution
+  const globalRegistry = instance.appContext.components
+  for (const key in globalRegistry) {
+    if (globalRegistry[key] === type) {
+      return key
+    }
+  }
+
   // try to retrieve name imported in script setup
   return getComponentNameInSetup(instance.parent, type) || null
 }
--- a/tests/mountingOptions/global.stubs.spec.ts
+++ b/tests/mountingOptions/global.stubs.spec.ts
@@ -367,6 +367,36 @@ describe('mounting options: stubs', () => {
     expect(wrapper.html()).toBe('<foo-bar-stub></foo-bar-stub>')
   })
 
+  it('stubs a globally registered component by its registration name', () => {
+    const LibraryComponent = defineComponent({
+      name: 'LibraryInternalName',
+      template: '<span>real library component</span>'
+    })
+
+    const Plugin = {
+      install(app: any) {
+        app.component('GlobalAlias', LibraryComponent)
+      }
+    }
+
+    const Component = defineComponent({
+      template: '<div><GlobalAlias /></div>'
+    })
+
+    const wrapper = mount(Component, {
+      global: {
+        plugins: [Plugin],
+        stubs: {
+          GlobalAlias: true
+        }
+      }
+    })
+
+    expect(wrapper.html()).toBe(
+      '<div>\n' + '  <global-alias-stub></global-alias-stub>\n' + '</div>'
+    )
+  })
+
   it('stubs components within script setup', () => {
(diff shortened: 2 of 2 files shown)


src/utils/componentName.ts
```diff
@@ -28,6 +28,14 @@ export const getComponentRegisteredName = (
...
   }
 
+  // try to infer the name based on global resolution
+  const globalRegistry = instance.appContext.components
+  for (const key in globalRegistry) {
+    if (globalRegistry[key] === type) {
+      return key
+    }
+  }
+
```
cexbrayat, line 38: maybe add this as the last case (after getComponentNameInSetup): this will avoid breaking existing tests and make sure local registration wins over global registration

cexbrayat (changes requested):
Hi @ifer47 
Thank you for the PR. This looks overall good to me, just move the global check as the last way to retrieve the name and I think we can merge.

cexbrayat:
Hi @ifer47 

When you have some time, can you address the feedback so we can merge the PR?

PR #2898, opened 2026-07-05
fix: clear emitted event history for child components on unmount

### LINKED ISSUE #2897: Bug: emitted event history for child components is not cleared on unmount
### Describe the bug

`@vue/test-utils` records component `emit` calls into a module-level `events` object in `src/emit.ts`, keyed by each component's `uid`:

```ts
// src/emit.ts
const events: Events = {}

export const recordEvent = (vm, event, args) => {
  // ...
  events[cid][event].push(args) // cid = emitting component's uid
}

export const removeEventHistory = (vm) => {
  const cid = vm.$.uid
  delete events[cid]        // only the root wrapper's uid
}
```

`recordEvent` runs for any component that emits. The devtools `component:emit` hook is global (`attachEmitListener` overrides `__VUE_DEVTOOLS_GLOBAL_HOOK__.emit`), and native DOM events attached by each `VueWrapper` also call `recordEvent(vm.$, ...)`. On `wrapper.unmount()`, however, `removeEventHistory(this.vm)` (in `src/vueWrapper.ts`) deletes only the root component's uid.

So entries recorded under a child component's uid stay in `events` after unmount. Vue `uid`s are monotonically increasing, so these entries are not reused or overwritten by later mounts. When a recorded payload is a native DOM `Event`, it holds a reference to `event.target`, which keeps the associated DOM nodes (and the handlers and reactive values they reference) reachable. With Vitest `isolate: false` (`--no-isolate`), `events` is shared across test files, so uncleared entries carry over between files.

The root cause is the asymmetry: `removeEventHistory` deletes one uid (the root), while `recordEvent` writes under the uid of every emitting component, including descendants.

### Reproduction

https://stackblitz.com/edit/github-goth86bt-gh1t8tyt?file=src%2Fcomponents%2F__tests__%2Femit-leak.spec.ts

### Expected behavior

After `wrapper.unmount()`, the recorded emit history for the whole component tree (root and descendants, including native DOM `Event` payloads) should be removed from `events`. In the reproduction, the child case should pass like the root control.

### Environment

```
  System:
    OS: macOS 26.5.1
    CPU: (8) arm64 Apple M1
    Memory: 87.31 MB / 8.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    vitest: 4.1.9 => 4.1.9 
    vue: 3.5.38 => 3.5.38
```

### Additional context

- Native DOM events go through the same path: `attachNativeEventListener` (in `src/vueWrapper.ts`) runs in the constructor of each `VueWrapper`, including child wrappers from `findComponent`, and records the raw DOM `Event` via `recordEvent(vm.$, eventName, args)` under the child's uid, which is not cleared.
- Related inconsistency: for functional components `recordEvent` records under `parent.uid` (it walks up while `type` is a function), while `removeEventHistory` uses `vm.$.uid`, so the keys can differ there as well.
- Possible fixes: on unmount, walk the component tree and delete each descendant uid (or group entries per app and drop them together); or key `events` by the component instance in a `WeakMap` so entries are released when the instance is collected.
- Context: `removeEventHistory` was introduced in #1449 to clear emit history on unmount, but only for the root uid. A separate unmount leak in `src/vueWrapper.ts` was addressed in #1793 (for #1787), which noted "there might be some more to optimize"; the `events` map in `src/emit.ts` was not part of that change.

ebiryu:
## Summary

`removeEventHistory` cleared only the root component's entry from the module-level `events` map in `src/emit.ts`, so emit histories recorded under descendant components' uids were never removed on `wrapper.unmount()`.

`recordEvent` writes under the uid of every component that emits: the devtools `component:emit` hook is global, and `attachNativeEventListener` records native DOM events against each wrapper (including child wrappers from `findComponent`). Native DOM `Event` payloads keep `event.target` and its DOM subtree reachable. With Vitest `isolate: false` the `events` map is shared across files, so these entries accumulate.

This walks the component tree on unmount and removes every component instance's entry.

Fixes #2897

## Notes
- `aggregateChildren` / `nodesAsObject` are duplicated from `src/utils/find.ts` to avoid coupling this path to the selector code; happy to extract a shared util if preferred.
- Components unmounted earlier in a test (e.g. via `v-if`) are not reachable from the tree at root-unmount time. A `WeakMap`-keyed store would also cover that and could be a follow-up.

### DIFF
--- a/src/emit.ts
+++ b/src/emit.ts
@@ -13,6 +13,7 @@ const enum DevtoolsHooks {
 }
 
 const events: Events = {}
+const recordedUidsByRoot = new Map<number, Set<number>>()
 
 export function emitted<T = unknown>(
   vm: ComponentPublicInstance,
@@ -75,6 +76,13 @@ export const recordEvent = (
   while (typeof wrapperVm?.type === 'function') wrapperVm = wrapperVm.parent!
 
   const cid = wrapperVm.uid
+  const rootCid = wrapperVm.root.uid
+
+  if (!recordedUidsByRoot.has(rootCid)) {
+    recordedUidsByRoot.set(rootCid, new Set())
+  }
+  recordedUidsByRoot.get(rootCid)!.add(cid)
+
   if (!(cid in events)) {
     events[cid] = {}
   }
@@ -87,6 +95,12 @@ export const recordEvent = (
 }
 
 export const removeEventHistory = (vm: ComponentPublicInstance): void => {
-  const cid = vm.$.uid
-  delete events[cid]
+  const rootCid = vm.$.root.uid
+  const uids = recordedUidsByRoot.get(rootCid)
+  if (!uids) return
+
+  for (const uid of uids) {
+    delete events[uid]
+  }
+  recordedUidsByRoot.delete(rootCid)
 }
--- a/tests/emit.spec.ts
+++ b/tests/emit.spec.ts
@@ -381,4 +381,59 @@ describe('emitted', () => {
     wrapper1.unmount()
     wrapper2.unmount()
   })
+
+  it('clears emitted event history of child components on unmount', async () => {
+    const Child = defineComponent({
+      name: 'Child',
+      emits: ['foo'],
+      setup(_, { emit }) {
+        emit('foo', 'bar')
+        return () => h('div', 'child')
+      }
+    })
+    const Parent = defineComponent({
+      name: 'Parent',
+      render: () => h(Child)
+    })
+
+    const wrapper = mount(Parent)
+    const child = wrapper.findComponent(Child)
+    expect(child.emitted('foo')).toHaveLength(1)
+
+    wrapper.unmount()
+
+    // the child's history must be removed too, not only the root's
+    expect(child.emitted('foo')).toBeUndefined()
+  })
+
+  it('clears history of a child removed before the root is unmounted', async () => {
+    const Child = defineComponent({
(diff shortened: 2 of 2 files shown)


cexbrayat (changes requested):
Thanks for the fix!

I think the tree-walk approach still misses one case: a child component that emits, then gets removed before the root wrapper is unmounted, for example via v-if.

At `wrapper.unmount()` time, that child is no longer reachable from `root.subTree`, so its uid is not collected and its entry remains in the module-level `events` map.

A possible fix is maybe to track recorded uids in a `Set` while events are recorded, scoped by root component:

```ts
const events: Events = {}
const recordedUidsByRoot = new Map<number, Set<number>>()
```

Then in `recordEvent`, after resolving `wrapperVm`:

```ts
const cid = wrapperVm.uid
const rootCid = wrapperVm.root.uid

if (!recordedUidsByRoot.has(rootCid)) {
  recordedUidsByRoot.set(rootCid, new Set())
}

recordedUidsByRoot.get(rootCid)!.add(cid)
```

and then use this set in the cleanup, without having to collect uids?

This is pseudo cod, maybe this does not work, but can you give it a try?

ebiryu:
Thanks @cexbrayat.

I switched to the approach you suggested and it works well. Also added a test for the case you mentioned (a child removed before the root unmounts).

cexbrayat (approved):
Thanks @ebiryu

PR #2904, opened 2026-07-12
chore!: drop class component support

cexbrayat:
vue-class-component is deprecated and no longer actively maintained, and class-based components are no longer recommended for Vue 3.

Vue language-tools is also dropping class component support as the ecosystem aligns around Vue 3 and the Composition API, so Test Utils should no longer carry dedicated class-component compatibility coverage.

We are starting to see tooling not supporting decorators out the box as well.

BREAKING CHANGE: class components are no longer supported by Vue Test Utils.


danielroe (approved):
looks good to me!

I might personally choose a `fix` or `refactor` rather than chore, but up to you!

PR #2913, opened 2026-07-29
chore: add actionlint with ShellCheck for workflow linting

chernojagne:
Adds [actionlint](https://github.com/rhysd/actionlint) to catch GitHub Actions workflow issues — schema errors, invalid expression contexts, and shell bugs in `run:` blocks via ShellCheck.

**What's added:**

- `.github/actionlint.yaml` — actionlint config that suppresses four ShellCheck rules that are known false positives in GitHub Actions `run:` blocks (`SC2086`, `SC2129`, `SC2155`, `SC1091`)
- `.github/workflows/actionlint.yml` — a dedicated workflow that runs actionlint on any push or PR that touches `.github/workflows/**`

The repo already has three workflow files (`ci.yml`, `release.yml`, `release-nightly.yml`) but no linting for them. This closes that gap with zero blast radius — two new files, no changes to existing ones.

---

This change was generated by [Autar](https://autar.ai) and reviewed by a human before submission.

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -6,7 +6,7 @@ self-hosted-runner:
 # ShellCheck rules disabled inside run: blocks.
 # Each `run:` block is checked through ShellCheck; suppress noisy rules here.
 paths:
-  ".github/workflows/**/*.{yml,yaml}":
+  '.github/workflows/**/*.{yml,yaml}':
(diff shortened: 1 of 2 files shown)


cexbrayat:
Thanks for the PR, but i don't think we need that

PR #2916, opened 2026-07-30
Add global configuration for plugins

chiaraani:
I've been trying to test my component which uses Quasar Plugin with the method explained here, but it didn't work. So I'd like to add to this guide the way it worked to me.


cexbrayat (approved):

cexbrayat:
Thanks for the PR

PR #2926, opened 2026-08-16
fix: trigger button click from keyboard activation

### LINKED ISSUE #2545: Bug: Keydown.enter and Keydown.space do not emit the click event on buttons
**Describe the bug**
Triggering the keydown event on a button for space or enter will not trigger the click event (which it does in an actual browser)

**To Reproduce**
https://stackblitz.com/edit/github-bbenms-4fnuwh?file=src%2FButton.vue&view=editor

**Expected behavior**
The click event should be triggered to accurately test keyboard navigation for accessibility

**Related information:**
```
  System:
    OS: Linux 5.15 Linux Mint 21.3 (Virginia)
    CPU: (20) x64 12th Gen Intel(R) Core(TM) i7-12700H
    Memory: 14.30 GB / 31.02 GB
    Container: Yes
    Shell: 0.96.0 - /home/n/.cargo/bin/nu
  npmPackages:
    @vue/test-utils: ^2.4.5 => 2.4.6 
    vitest: ^1.5.0 => 1.6.0 
    vue: ^3.5.12 => 3.5.12 
```

**Additional context**

rajanpanth:
### Description

Fixes #2545.

`wrapper.trigger('keydown.enter')` and `wrapper.trigger('keydown.space')` now synthesize the browser-style click activation for enabled `<button>` elements. The synthesized click is only dispatched if the keydown event is not canceled, so handlers using `.prevent` can suppress activation like they do in the browser.

### Validation

- `corepack pnpm vitest run tests/trigger.spec.ts`
- `corepack pnpm exec oxfmt --check src/baseWrapper.ts tests/trigger.spec.ts`
- `corepack pnpm exec oxlint src/baseWrapper.ts tests/trigger.spec.ts` (exits 0; existing warning-only findings remain in the file)
- `corepack pnpm run build`
- `git diff --check`

### DIFF
--- a/src/baseWrapper.ts
+++ b/src/baseWrapper.ts
@@ -33,6 +33,31 @@ import { stringifyNode } from './utils/stringifyNode'
 import type { HTMLBeautifyOptions } from 'js-beautify'
 import beautify from 'js-beautify'
 
+function isKeyboardClickActivation(
+  element: Node,
+  event: Event & TriggerOptions
+) {
+  if (
+    !isElement(element) ||
+    !(element instanceof HTMLButtonElement) ||
+    event.type !== 'keydown'
+  ) {
+    return false
+  }
+
+  const key = typeof event.key === 'string' ? event.key.toLowerCase() : ''
+
+  return (
+    key === 'enter' ||
+    key === ' ' ||
+    key === 'space' ||
+    event.code === 'Enter' ||
+    event.code === 'Space' ||
+    event.keyCode === 13 ||
+    event.keyCode === 32
+  )
+}
+
 export default abstract class BaseWrapper<
   ElementType extends Node
 > implements WrapperLike {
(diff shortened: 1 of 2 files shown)


cexbrayat:
Thanks for the PR. I took a deeper look into the issue, and I'm not convinced we should add this.

After doing some research, I don't think other testing libraries do this:
 - React Testing Library fireEvent(), Cypress trigger(), and Playwright dispatchEvent() dispatch only the requested event.
 - Browser-like APIs testing library userEvent.keyboard(), Cypress press(), and Playwright press() perform complete keyboard interactions and may produce clicks. These higher-level APIs correctly activate Enter before release and Space on keyup, while honoring preventDefault().

So I don't think we should alter the behavior of `trigger` in VTU.
`@testing-library/user-event` is framework independent and could be used for this use-case:

```ts
import userEvent from '@testing-library/user-event'
import { nextTick } from 'vue'

const user = userEvent.setup()
  attachTo: document.body
})

const button = wrapper.get('button').element as HTMLButtonElement
button.focus()

await user.keyboard('{Enter}')
// or:
await user.keyboard(' ')

await nextTick()
```

PR #2934, opened 2026-08-27
Fix generic SFC findComponent types

### LINKED ISSUE #2436: Bug: TypeScript errors with Vue generic components
### Describe the bug

I get TypeScript errors when trying to find a component that has a [generic](https://vuejs.org/api/sfc-script-setup.html#generics) with `findComponent`.

### Steps to reproduce

I have made a minimal reproduction. Here a small component is created with a generic, called `Select.vue`. I use the Select component in `App.vue`.

1. [Open minimal reproduction](https://stackblitz.com/edit/vitejs-vite-jnknze?file=src%2F__tests__%2Fapp.spec.ts)
2. You should now see `app.spec.ts`. This is the test file where the TypeScript errors occur.
3. Run `npm run type-check` in the terminal below the file editor.
4. You should now see two TypeScript errors.

I think the two errors are related. They are only thrown if the component has a generic.

### Expected behaviour

It should not throw these errors. The `findComponent` for the component with the generic, should return a `VueWrapper`.

### Actual behaviour

It somehow returns a `WrapperLike`, instead of the expected `VueWrapper`.

simonyang08:
Fixes #2436

## Summary

- recognize generic SFC call signatures in `findComponent`
- return `VueWrapper` instead of falling through to `DOMWrapper<Node>`
- add declaration-level regression coverage for `<script setup generic>` components

## Root cause

Generic SFCs emitted by Vue's `<script setup generic>` syntax expose a generic call signature rather than the construct signature used by `DefinedComponent`. The existing overloads therefore selected the DOM-wrapper fallback.

## Compatibility

The new overload is limited to generic call signatures returning `VNode`. Existing functional-component and defined-component overloads remain unchanged. The result intentionally uses an unparameterized `VueWrapper` because `findComponent` cannot instantiate the SFC's generic type parameter soundly.

## Validation

- actual generic-SFC `vue-tsc` reproduction - failed before, passed after
- `pnpm tsd` - passed
- `pnpm test --run tests/findComponent.spec.ts` - 36 passed
- `pnpm build` - passed
- `pnpm lint` - passed with unrelated existing warnings
- changed-file formatting and `git diff --check` - passed

## AI disclosure

This change was prepared with OpenAI Codex assistance and reviewed and validated locally by the contributor.

### DIFF
--- a/src/interfaces/wrapperLike.ts
+++ b/src/interfaces/wrapperLike.ts
@@ -1,5 +1,6 @@
 import type { DomEventNameWithModifier } from '../constants/dom-events'
 import type { TriggerOptions } from '../createDomEvent'
+import type { VNode } from 'vue'
 import type {
   DefinedComponent,
   FindAllComponentsSelector,
@@ -36,6 +37,9 @@ export default interface WrapperLike {
   findComponent<T extends DefinedComponent>(
     selector: T | Exclude<FindComponentSelector, FunctionalComponent<any>>
   ): VueWrapper<InstanceType<T>>
+  // Generic SFCs emitted by vue-tsc have a generic call signature instead of
+  // the construct signature used by DefinedComponent.
+  findComponent<T extends <U>(...args: any[]) => VNode>(selector: T): VueWrapper
   findComponent<T extends FunctionalComponent<any>>(
     selector: T | string
   ): DOMWrapper<Element>
--- a/src/interfaces/wrapperLike.ts
+++ b/src/interfaces/wrapperLike.ts
@@ -37,9 +37,17 @@ export default interface WrapperLike {
   findComponent<T extends DefinedComponent>(
     selector: T | Exclude<FindComponentSelector, FunctionalComponent<any>>
   ): VueWrapper<InstanceType<T>>
-  // Generic SFCs emitted by vue-tsc have a generic call signature instead of
-  // the construct signature used by DefinedComponent.
-  findComponent<T extends <U>(...args: any[]) => VNode>(selector: T): VueWrapper
+  // vue-tsc emits generic SFCs as call signatures returning `VNode & { __ctx?: ... }`,
+  // not as `DefineComponent` constructs. The `__ctx` property is the discriminator:
+  // a plain generic functional component returns bare `VNode` and must keep returning
+  // `DOMWrapper` so `.vm` does not type-check.
+  findComponent<T extends (...args: any[]) => VNode>(
+    selector: T
+  ): unknown extends ReturnType<T>
+    ? DOMWrapper<Element>
+    : '__ctx' extends keyof ReturnType<T>
+      ? VueWrapper
+      : DOMWrapper<Element>
   findComponent<T extends FunctionalComponent<any>>(
     selector: T | string
   ): DOMWrapper<Element>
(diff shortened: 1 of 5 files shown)


src/interfaces/wrapperLike.ts
```diff
@@ -36,6 +37,9 @@ export default interface WrapperLike {
   findComponent<T extends DefinedComponent>(
     selector: T | Exclude<FindComponentSelector, FunctionalComponent<any>>
   ): VueWrapper<InstanceType<T>>
+  // Generic SFCs emitted by vue-tsc have a generic call signature instead of
+  // the construct signature used by DefinedComponent.
+  findComponent<T extends <U>(...args: any[]) => VNode>(selector: T): VueWrapper
```
cexbrayat: shouldn't the same be done for getComponent and findAllComponent?

cexbrayat: I'm also afraid that this matches any generic function returning `VNode`, including ordinary generic functional components. Those produce a `DOMWrapper`at runtime, so properties such as  `vm` incorrectly type-check. The overload needs to identify vue-tsc’s generic-SFC shape

cexbrayat (changes requested):

simonyang08:
Thanks for the careful review — both points are addressed in the latest push (fea3b56).

**1. Same treatment for `getComponent` and `findAllComponents`.**

Done — both now carry the same discriminated overload, with return types mirroring their existing overloads (`Omit<VueWrapper, 'exists'>` / `VueWrapper[]` on the vue-tsc path, and the DOM wrapper types they already return otherwise).

**2. Discriminating vue-tsc generic SFCs from ordinary generic functional components.**

Valid concern — the original constraint matched any `VNode`-returning function. I verified empirically (running `vue-tsc --declaration` on a minimal `<script setup lang="ts" generic="...">` SFC) that vue-tsc emits `VNode & { __ctx?: ... }`, while ordinary generic functionals return bare `VNode`. The discrimination now happens in the return type: `'__ctx' extends keyof ReturnType<T> ? VueWrapper : DOMWrapper<...>`.

While implementing, one more edge turned up that needed an explicit guard: Vue's `FunctionalComponent` call signature returns `any`, which satisfies the `VNode`-returning constraint and makes `'__ctx' extends keyof ReturnType<T>` evaluate to true (`keyof any` includes every string). The overloads therefore start with an `unknown extends ReturnType<T>` check that routes `FunctionalComponent` selectors to the DOM wrapper before the `__ctx` check. With that in place, `FunctionalComponent`-typed components resolve to `DOMWrapper` on all three APIs and `.vm` no longer type-checks on them — asserted via `@ts-expect-error` in `test-dts/genericComponent.d-test.ts`, along with `WrapperLike`-level assertions mirroring the `BaseWrapper` overloads.

`pnpm run tsd` is green, the full vitest suite passes (503 tests), and oxlint/oxfmt are clean on the changed files. Ready for another look.

test-dts/genericComponent.d-test.ts
```diff
@@ -0,0 +1,106 @@
...
+// `vue-tsc` emits a generic SFC as a function with a generic call signature whose
+// return type is `VNode & { __ctx?: ... }`. The `__ctx` property is the
+// discriminator between vue-tsc generic SFCs and ordinary generic functional
+// components (which return bare `VNode`).
+declare const VueTscGenericSfc: <T extends string | number>(
+  __VLS_props: any,
+  __VLS_ctx?: any,
+  __VLS_expose?: any,
+  __VLS_setup?: Promise<any>
+) => VNode & { __ctx?: any }
```
cexbrayat: I think it would be better to use a real `<script setup generic="Item">` component in a .vue file imported by the test, instead of relying on internals of vue-tsc

cexbrayat (changes requested):
Even if it is a bit too much of LLM gibberish, it sounds like a correct approach.

Just update the test and I think we'll be good to merge.

simonyang08:
Thanks — fair point. Dropped the hand-written declaration and added `test-dts/GenericSfc.vue`, a real `<script setup generic="Item extends string | number">` SFC that the d-test imports, so the assertions now exercise the actual type vue-tsc emits. Since plain `tsc` can't resolve `.vue`, the `tsd` script now runs `vue-tsc --noEmit` instead (vue-tsc was already a devDependency, and it's the same tool that produces the `__ctx` shape the overload discriminates on).

The negative cases are unchanged: plain generic functionals still resolve to `DOMWrapper` with `@ts-expect-error` on `.vm`, and the `FunctionalComponent` shape is covered the same way.

New commit: 7bd2bef. vue-tsc, vitest (56 focused / 503 full) and lint are all green.

cexbrayat (approved):

PR #2941, opened 2026-09-06
fix: keep the namespace prefix in attributes()

MaxFreedomPollard:
## Summary

`attributes()` indexed its result by the attribute's local name, so a namespaced attribute lost its prefix. `xlink:*` attributes on SVG elements were reported under the wrong key, and an element carrying both `xlink:href` and `href` silently lost one of the two.

## Root cause

`src/baseWrapper.ts` line 286 built the map with `attributeMap[attribute.localName] = attribute.value`. `Attr.localName` is the name with the namespace prefix removed, so `xlink:href` is stored as `href`. Vue sets `xlink:*` attributes on SVG elements through `setAttributeNS` (`patchAttr` in `@vue/runtime-dom`), which is the ordinary path for SVG sprite icons, so those attributes always arrive namespaced.

Mounting `<svg><use xlink:href="#icon" href="#fallback" /></svg>` on `main` gives:

```js
wrapper.find('use').html() // '<use xlink:href="#icon" href="#fallback"></use>'
wrapper.find('use').attributes() // { href: '#fallback' }, xlink:href is gone
wrapper.find('use').attributes('xlink:href') // undefined
```

Both attributes collide on the key `href` and the last one wins, so `attributes()` and `html()` disagree about what the element carries.

The same element shows the inconsistency from the other side: `xmlns:xlink` is written with plain `setAttribute` and therefore has no namespace, so its prefix survives, while `xlink:href` loses its own.

## Fix

Index by `Attr.name`, the qualified name, which is exactly what `html()` renders.

HTML attribute names are lowercased by the parser, so `name` and `localName` are the same string for every attribute without a prefix, and unprefixed attributes are untouched. On an element that carries only `xlink:href`, `attributes('href')` no longer resolves to it: it is now reported under `xlink:href`.

## Validation

The regression test added to `tests/attributes.spec.ts` fails on `main` with `AssertionError: expected { href: '#fallback' } to deeply equal { 'xlink:href': '#icon', ... }` and passes with the fix.

- `vitest run --maxWorkers=3`: 54 files, 505 passed, 1 todo, 0 failed
- `rollup -c rollup.config.ts --bundleConfigAsCjs` then `vitest run --mode test-build tests/attributes.spec.ts`: 6 passed, confirming the built bundle behaves the same
- `pnpm run lint`: exit 0, and `oxlint src/baseWrapper.ts tests/attributes.spec.ts` reports nothing
- `pnpm run format-check`: all 239 files correctly formatted
- `pnpm run tsd`: exit 0
- `pnpm run vue-tsc`: exit 0


simonyang08:
Reviewed the change and verified the DOM semantics it relies on (jsdom, same as the test env): for `<use xlink:href="#icon" href="#fallback" viewBox="0 0 2 2">`, the two keying strategies produce

```js
// main (attribute.localName) — xlink value silently lost to the collision
{ href: '#fallback', viewBox: '0 0 2 2' }
// PR (attribute.name) — both preserved
{ 'xlink:href': '#icon', href: '#fallback', viewBox: '0 0 2 2' }
```

Two things worth noting in favor:

- For every non-prefixed attribute — including case-adjusted SVG ones like `viewBox` — `Attr.name === Attr.localName`, so this is a strictly additive fix; nothing that works today changes.
- `Vue` sets `xlink:*` through `setAttributeNS` (as the PR body says), so the reproduction is realistic, and `wrapper.attributes('xlink:href')` becomes usable.

No other concerns: the regression test locks both the prefixed key and the coexistence case. One process note — the Node `build` checks haven't run on this PR yet (only the Netlify ones), which looks like the first-contribution workflow-approval gate rather than anything about the change; worth a maintainer click.

From my side this is good to merge once CI has run.

cexbrayat (approved):
Thanks!

PR #2948, opened 2026-09-18
feat(types): infer emitted event names and payload types

### LINKED ISSUE #2598: Feature request: Type inference support for wrapper emitted method

Is there a reason why the signature of the emitted method in the wrapper class is limited to primitive types and does not infer possible event names and types of their arguments using it's generic class type argument.

Indeed, in typescript projects, it is much more convenient to have autocompletion for possible events that are declared by a component.

Infer possible event names (and types of their arguments) if the component declares the appropriate type annotations.

As a temporary alternative, I created a wrapper over the mount function, which redefines the signature of the emitted function in the returned object using typecast (while its runtime implementation remains the same). I calculate the types based on the signature of the $emit method using the generic Parameters helper.

**Additional context**

\-

niukanen1:
wrapper.emitted() now infers the declared event names and the payload tuple of each event from the component's emits option, via the onXxx handler props Vue generates for them

Components with typed emits get autocompletion for event names and typed payload tuples; array-form emits infer the names; components without declared emits keep the previous string/unknown[] behavior, and the explicit type parameter still overrides the inference

Closes #2598


cexbrayat (changes requested):
Thanks for the PR! That looks good to me, even the TypeScript dark magic used is hard to follow for mere humans... 

Can you fix the CI build so we can merge?

niukanen1:
The build was failing on the new oxfmt 0.68 formatting rules — reformatted the two touched files, CI should be green now

cexbrayat (approved):

